PR #1100, opened 2024-03-18
[ERA-9328] i18n -- Adding Portuguese

JoshuaVulcan:
### What does this PR do?
- Adds the Portuguese translations provided by volunteers

### Relevant link(s)
* Tracking tickets: [ERA-9328](https://allenai.atlassian.net/browse/ERA-9328)
* Env still building

[ERA-9328]: https://allenai.atlassian.net/browse/ERA-9328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


fonzy911:
Hi @JoshuaVulcan. It seems that the UI is not reflecting the Portuguese translations.

image

I do see the components enabled but nothing reflected:

image

public/locales/pt/components.json
```diff
@@ -0,0 +1,179 @@
+{
+  "additembutton": {
```
luixlive: The environment is not working as @fonzy911 mentioned cause the keys are not resolving. All of them were "lowercase-ed". For example this one should be `addItemButton`, not `additembutton` 👍

luixlive (changes requested):

JoshuaVulcan:
Sorry y'all, I made a silly mistake with my conversion script to turn the spreadsheet pages into our translation files. Should be fixed now! @luixlive @fonzy911

luixlive (approved):

fonzy911 (commented):
Hi @JoshuaVulcan, everything looks better now, thank you. It seems that the time slider part is also not translated in portuguese:

image

Also, it seems that the word "Events" and the actual events names are not being translated either:

Screen Shot 2024-04-03 at 14 26 37

tiffanynwong:
- Toast is not translated
Screen Shot 2024-04-03 at 11 39 47 AM

- "Event" button in event detail not translated (Portuguese, Spanish, and French)
Screen Shot 2024-04-03 at 11 42 14 AM
- 404 on alert tab
Screen Shot 2024-04-03 at 11 44 26 AM

- Timeslider control popup says "About 1 Month" (Portuguese & French)
Screen Shot 2024-04-03 at 11 46 40 AM
- Should it say Portuguese (Brazil)? Or remove (United States) from the English option for consistency? 
Screen Shot 2024-04-03 at 11 53 34 AM

luixlive:
> * Should it say Portuguese (Brazil)? Or remove (United States) from the English option for consistency?
>   Screen Shot 2024-04-03 at 11 53 34 AM

About this, I think it makes sense to keep it as it is, or we can remove the `(United States)` for now. The thing is that english is localized to `en-US`, while Spanish, French and Portuguese are just `es`, `fr` and `pt`. The reasoning behind not localizing those is simply that I'd think that we will want to eventually localize english to different countries given the amount of users we have, in contrast to the other ones. But we can always change any of these configurations.

luixlive:
> * Timeslider control popup says "About 1 Month" (Portuguese & French)
>   Screen Shot 2024-04-03 at 11 46 40 AM

I think this one should be taken as a separate bug ticket cause seems like it could be caused by something wrong going on with the automatic translation of the `calcFriendlyDurationString` method.

fonzy911:
Hi @luixlive , thank you very much for the fixes, just a couple more things if I may. It seems that, for some reason, the method that calls the translated tooltip name for the timeslider, is not being called upon appropriately, cause it looks like this, instead of showing the actual translated name:

image

Also, whenever we want to create either a new event or patrol, only half of the modal's fields appear to be translated as well, not sure if this goes out of the scope of this particular ticket:

Screen Shot 2024-04-08 at 13 16 34

PR #1116, opened 2024-04-30
ERA-9687: Geopermissions - Profile with geopermission assigned, accessed using PIN don't honor their permissions.

luixlive:
### What does this PR do?
The bug was that we always check if the location parameter should be appended to the events request from the active `user`. When we are under a profile, the active `user` is still the main user, so the permissions may not match, resulting in no location parameter being added to the request when it should. This PR does the geo-restriction permissions check from the `selectedUserProfile` instead when there is an active profile.

### Relevant link(s)
* [ERA-9687](https://allenai.atlassian.net/browse/ERA-9687)
* [Env](https://era-9687.dev.pamdas.org/login)

[ERA-9687]: https://allenai.atlassian.net/browse/ERA-9687?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ducks/events.js
+++ b/src/ducks/events.js
@@ -63,7 +63,10 @@ export const SOCKET_EVENT_DATA = 'SOCKET_EVENT_DATA';
 const UPDATE_EVENT_STORE = 'UPDATE_EVENT_STORE';
 
 const shouldAppendLocationToRequest = (state) => {
-  return userIsGeoPermissionRestricted(state?.data?.user) && !!state?.view?.userLocation?.coords;
+  const currentUser = state?.data?.selectedUserProfile?.username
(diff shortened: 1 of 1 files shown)


JoshuaVulcan (approved):

AlanCalvillo:
I know I didn't mention it in the bug itself but I noticed that even though the Events are visible in the Map, I was expecting that they would be visible in the Feed List as well. 
They do appear when I login directly into the profile so I guess it would be a similar fix.

image

JoshuaVulcan (approved):

AlanCalvillo (approved):

PR #1118, opened 2024-05-02
[ERA-9327] Swahili i18n

JoshuaVulcan:
### What does this PR do?
- Adds Swahili language support to the Earthranger web client

### How does it look
- N/A

### Relevant link(s)
* https://allenai.atlassian.net/browse/ERA-9327
* https://era-9327.dev.pamdas.org/


AlanCalvillo:
Any thoughts on why the first part of this doesn't get translated? :( (About 1 ...)
image

JoshuaVulcan:
@AlanCalvillo yes, our "human readable date string" utility function's dependencies don't appear to support i18n in the way we'd expect. It's an issue with the current codebase as well, i'd recommend filing it as a bug ticket but not blocking i18n work

AlanCalvillo (approved):
Bug to follow-up is linked to the Jira ticket.

PR #1121, opened 2024-05-16
[ERA-9719] feat:Nepali language support

JoshuaVulcan:
https://allenai.atlassian.net/browse/ERA-9719
https://era-9719.dev.pamdas.org/


luixlive:
What's the status of this PR?

.env.development
```diff
@@ -1,4 +1,4 @@
-REACT_APP_DAS_HOST=https://develop2.pamdas.org
+REACT_APP_DAS_HOST=https://develop.pamdas.org
```
luixlive: I guess it should be `root.develop` by default now?

src/PrintTitle/index.js
```diff
@@ -1,17 +1,18 @@
-import React, { memo } from 'react';
```
luixlive, line 1: Just a refactor?

src/ReportManager/ReportDetailView/index.js
```diff
@@ -441,8 +441,7 @@ const ReportDetailView = ({
 
     setReportForm({ ...reportForm, event_details: { ...reportForm.event_details, ...formData } });
 
-    reportTracker.track('Change Report Form Data');
```
luixlive, line 444: Is removing this line intended?

src/i18n.js
```diff
@@ -54,6 +54,7 @@ i18n
           es: '0.7',
```
luixlive: I'd say we are ready to bump all to v1 😄 That way all of them are in sync again and we can go from version 1 onwards.

public/locales/ne-NP/cookies-preferences.json
```diff
@@ -0,0 +1 @@
+{}
```
luixlive: Are we missing this one?

luixlive (commented):
Sadly I can't read Nepali, but the configuration seems well. I just added a few comments I'd like to discuss before approving.

src/ReportManager/ReportDetailView/index.js
JoshuaVulcan: Yeah, it fires on every keystroke in the form and is spamming our analytics in a not-helpful way.

public/locales/ne-NP/cookies-preferences.json
JoshuaVulcan: No -- I'm generating these from the translation spreadsheet with some code, so it created a cookie preferences file to match that tab in the sheet. My miss on not deleting it.

.env.development
JoshuaVulcan: fixed

src/i18n.js
JoshuaVulcan: done!

luixlive (approved):

PR #1123, opened 2024-05-22
ERA-9761: Add page_size to event feed requests

luixlive:


JoshuaVulcan (approved):

luixlive:
@JoshuaVulcan this PR was updated with further changes with the same intention that the original change: ease the feed request. Now we are not including the event updates when fetching the feed, instead now everytime we open an event page we reload that event so we make sure we have all we need to show the page.

src/ReportManager/index.js
```diff
@@ -109,24 +110,24 @@ const ReportManager = ({ onReportBeingAdded }) => {
   }, [isNewReport, location.pathname, location.search, location.state, navigate, newReportTemporalId, reportType]);
 
   useEffect(() => {
-    if (!isNewReport && !eventStore[reportId]) {
```
luixlive, line 112: We no longer check if the event is in the store, cause probably it is if it was loaded by the feed, but it won't have the updates property.

src/ReportManager/index.js
```diff
@@ -109,24 +110,24 @@ const ReportManager = ({ onReportBeingAdded }) => {
...
+    {shouldRenderReportDetailView ? <ReportDetailView
       formProps={navigationData?.formProps}
       isNewReport={isNewReport}
       key={reportId} // This resets component state when the id changes
       newReportTypeId={newReportTypeId}
       onAddReport={onAddReport}
       reportData={reportData}
       reportId={reportId}
-    />}
+    /> : <LoadingOverlay />}
```
luixlive, line 137: Added a loading spinner

src/SideBar/useReportsFeed/index.js
```diff
@@ -29,7 +29,7 @@ const useReportsFeed = () => {
     return isEqual(restEventFilter, INITIAL_FILTER_STATE);
   }, [eventFilter]);
   const eventParams = useRef(calcEventFilterForRequest(
-    { params: { exclude_contained: shouldExcludeContained }, format: 'object' },
+    { params: { exclude_contained: shouldExcludeContained, include_updates: false, page_size: 25 }, format: 'object' },
```
luixlive: We exclude the `updates` property from the feed request and make sure to only load 25 events

JoshuaVulcan: `include_related_events`, `include_notes`, `include_details`, and `include_files` are also available as parameters here. Should we slim this down even further?

src/ReportManager/index.js
JoshuaVulcan: I don't mind this, but it's a little blunt and takes away an (overall helpful) optimization.  Clicking any previously-viewed event will cause duplicate requests. Could we eval with a function to test if the event's data structure is complete and only fetch if not?

JoshuaVulcan (approved):

chrisj-er (approved):

AlanCalvillo:
After we create a brand new event. When we try to open it back, the load spinner in the event details gets clogged. It's interesting, though, bc no request in the Network is open and no console error is prompted.
image

PR #1124, opened 2024-05-27
ERA-9660: Close event details view when hitting the escape key

luixlive:
### What does this PR do?
Close details view when pressing esc key.

### Relevant link(s)
* [ERA-9660](https://allenai.atlassian.net/browse/ERA-9660)
* [Env](https://era-9660.dev.pamdas.org)

- [src/SideBar/index.js](https://github.com/PADAS/das-web-react/compare/ERA-9660?expand=1#diff-788e631783fac0966b699a5548727810acc1d4abab140b5c51f5af1fd07f6946) has the logic to listen to the Escape key and navigate to the feed from the event details pages.
- The rest of the files just stops the propagation of the Escape pressing event so the navigation doesn't happen when using is closing a menu.

Since we have lots of visual elements that close when pressing the escape key, like modals, popups and menus, we need to be very specific on how we listen to the event. It only works when the details view or its children are focused.

I couldn't find a way to stop the propagation of the Escape keyboard event when the location and geometry popovers are open, so in those cases the details view will be closed:
image

[ERA-9660]: https://allenai.atlassian.net/browse/ERA-9660?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/DatePicker/index.js
+++ b/src/DatePicker/index.js
@@ -256,6 +256,12 @@ const CustomDatePicker = ({
     onCalendarClose?.();
   }, [onCalendarClose]);
 
+  const onKeyDown = useCallback((event) => {
+    if (event.key === 'Escape' && isOpen) {
+      event.stopPropagation();
+    }
+  }, [isOpen]);
+
   const CustomHeader = useMemo(() => renderCustomHeader(rest?.maxDate, rest?.minDate, language, locale), [rest?.maxDate, rest?.minDate, language, locale]);
 
   return <DatePicker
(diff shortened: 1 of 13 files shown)


fonzy911 (approved):
Was able to check developed feature against acceptance criteria and everything works as expected:

https://github.com/PADAS/das-web-react/assets/60930654/f3e8129e-eacd-4a2a-bd5b-5ce2492f98c3

JoshuaVulcan (approved):
Clean and simple! This will be much appreciated.

PR #1129, opened 2024-06-03
ERA-9761: Add page_size to event feed requests

luixlive:


JoshuaVulcan (approved):

fonzy911 (approved):
Issue seems to be corrected. The spinner is not "hanging" anymore, it only takes like a flicker or so to load the events info now.

https://github.com/PADAS/das-web-react/assets/60930654/b68f90bb-5e02-4f5b-adaf-3ba29583c99b

AlanCalvillo (approved):

PR #1139, opened 2024-07-06
ERA-9773: Header and section settings updates

gaboDev:
Frontend for ERA-9773


JoshuaVulcan:
@gaboDev can you close this if it's not needed?

luixlive:
@JoshuaVulcan @gaboDev this ticket was finished last sprint I think. It's status is `Done`. Also, we don't support these types of deployments for the EFB (again 😞 ). So I think we are good closing it.

PR #1152, opened 2024-08-19
[ERA-10175] fix:simplifying schema field select widget code to allow touch events to open/close the select menu

JoshuaVulcan:
### What does this PR do?
- Removes legacy code which appears to be swallowing event propagation in a way that 

### Relevant link(s)
* https://allenai.atlassian.net/browse/ERA-10175
* env still building

https://allenai.atlassian.net/browse/ERCS-2577
This impacts a number of sites who prefer to use mobile web.


src/SchemaFields/index.js
```diff
@@ -258,10 +258,6 @@ export const SelectWidget = ({
 
   const [isMenuOpen, setMenuOpen] = useState(false);
 
-  const SelectContainer = (props) => <div ref={containerRef}>
```
luixlive, line 261: Do we still need the `containerRef` if this was deleted?

luixlive (approved):
Approved. I just left an observation.

src/SchemaFields/index.js
JoshuaVulcan: I'm not sure 🤔 honestly I'm trying to keep the surface area of changes as small as possible

fonzy911 (approved):
Feature appears to be working properly now:

https://github.com/user-attachments/assets/c37eda23-c77c-4030-a170-c17f115a5f71

PR #1170, opened 2024-10-03
ERA-10102: Top-level "clear all" filter control

gaboDev:
### What does this PR do?
- It adds global reset button for patrol filters
- It adds missing validation for the search text to the global reset button of events

### Relevant link(s)
* Tracking tickets: [ERA-10102](https://allenai.atlassian.net/browse/ERA-10102)
* [Hosted demo environments](https://era-10102.dev.pamdas.org)

- The new global reset button should reset all filter for patrols including search text, dates, tracked by, patrol type and patrol status
- Now the global reset button for events should react as well with search text

[ERA-10102]: https://allenai.atlassian.net/browse/ERA-10102?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/EventFilter/index.js
+++ b/src/EventFilter/index.js
@@ -224,7 +224,7 @@ const EventFilter = ({
         totalFeedCount={feedEvents.count}
       />
       {
-        (filterModified || isDateRangeModified || isSortModified) &&
+        (filterModified || isDateRangeModified || isSortModified || !!filterText) &&
         <Button type="button" variant='light' size='sm' onClick={resetAllFilters} data-testid='general-reset-btn'>
           <RefreshIcon title={t('resetButton')} />
           {t('resetButton')}
--- a/src/PatrolFilter/FiltersPopover/index.js
+++ b/src/PatrolFilter/FiltersPopover/index.js
@@ -127,7 +127,8 @@ const FiltersPopover = React.forwardRef(({
 
   const patrolLeaderFilterOptions = patrolLeaderSchema?.trackedbySchema?.properties?.leader?.enum_ext?.map(({ value }) => value)
     || [];
-  const selectedLeaders = !!selectedLeaderIds?.length ?
+
+  const selectedLeaders = !!selectedLeaderIds?.length && !isEmpty(patrolLeaderSchema) ?
     selectedLeaderIds.map(id => patrolLeaderFilterOptions.find(leader => leader.id === id))
     : [];
 
--- a/src/PatrolFilter/index.js
+++ b/src/PatrolFilter/index.js
@@ -173,8 +173,8 @@ const PatrolFilter = ({ className }) => {
       {
           (filtersModified || dateRangeModified || !!filterText) &&
           <Button type="button" variant='light' size='sm' onClick={resetAllFilters}>
-            <RefreshIcon title={t('resetButton')} />
-            {t('resetButton')}
+            <RefreshIcon title={t('globalResetFilterButton')} />
+            {t('globalResetFilterButton')}
           </Button>
       }
     </div>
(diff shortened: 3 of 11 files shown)


src/ReportedBySelect/index.js
```diff
@@ -168,7 +168,7 @@ const ReportedBySelect = ({
 
     if (isMulti) {
       if (value.length) {
-        return value.map((item) => item.hidden
+        return value.map((item) => item?.hidden
```
gaboDev: Some reason the lead array was having 1 item being undefined, not sure how this happened but it was blocking me, I'm still trying to reproduce the issue

Anyways, that is not related with this ticket, but when I get something more reproducible I'll come back here and check if this was indeed the fix for it

src/PatrolFilter/FiltersPopover/index.js
```diff
@@ -127,7 +127,8 @@ const FiltersPopover = React.forwardRef(({
 
   const patrolLeaderFilterOptions = patrolLeaderSchema?.trackedbySchema?.properties?.leader?.enum_ext?.map(({ value }) => value)
     || [];
-  const selectedLeaders = !!selectedLeaderIds?.length ?
+
+  const selectedLeaders = !!selectedLeaderIds?.length && !isEmpty(patrolLeaderSchema) ?
```
gaboDev, line 131: in the `FilterPopover` component of the patrol filters, the `selectedLeaders` prop which is passed as the value prop to `ReportedBySelect`sometimes was having this array value `[undefined]` that is because the `patrolLeaderSchema` is empty, causing the mapping of the array happens even if there is no leader information still available

src/ReportedBySelect/index.js
gaboDev: I reset this change, instead I added the [validation here](https://github.com/PADAS/das-web-react/pull/1170/files#r1787908920)

src/EventFilter/index.js
```diff
@@ -224,7 +224,7 @@ const EventFilter = ({
         totalFeedCount={feedEvents.count}
       />
       {
-        (filterModified || isDateRangeModified || isSortModified) &&
+        (filterModified || isDateRangeModified || isSortModified || !!filterText) &&
```
luixlive, line 227: Could we test the fact that the button shows up if there is a text?

src/PatrolFilter/index.js
```diff
@@ -48,13 +51,38 @@ const PatrolFilter = ({ className }) => {
     patrolFilterTracker.track('Change Search Text Filter');
   }, []);
 
-  const resetSearch = useCallback((e) => {
-    e.stopPropagation();
+  const resetSearch = useCallback((event = null) => {
```
luixlive: nit: Seems like the `= null` is unnecessary. Adding the `?.` protection is enough.

src/PatrolFilter/index.js
```diff
@@ -48,13 +51,38 @@ const PatrolFilter = ({ className }) => {
...
+    event?.stopPropagation();
     setFilterText('');
 
     patrolFilterTracker.track('Clear Search Text Filter');
   }, []);
 
+  const resetDateRange = useCallback(() => {
```
luixlive: Why do we define these methods `resetDateRange` and `resetFiltersPopover` if they are not reused and add memoization processing? We could simply write the code they have directly in `resetAllFilters` right?

luixlive (approved):
A couple comments, nothing blocking. Regarding i18n, are we sure these translations are correct? They seem trivial enough, but let's make sure to get right translations for the context of the button from ChatGPT if you didn't already.

fonzy911 (approved):

PR #1173, opened 2024-10-11
ERA-10399: Fixing unit tests

gaboDev:
### What does this PR do?
- Fixes broken unit test for `<PrintTitle />`

### Relevant link(s)
* Tracking tickets: [ERA-10399](https://allenai.atlassian.net/browse/ERA-10399)

[ERA-10399]: https://allenai.atlassian.net/browse/ERA-10399?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/PrintTitle/index.test.js
+++ b/src/PrintTitle/index.test.js
@@ -1,11 +1,12 @@
 import React from 'react';
-import { render, screen } from '@testing-library/react';
 import '@testing-library/jest-dom/extend-expect';
 import { Provider } from 'react-redux';
+import { render, screen } from '@testing-library/react';
 import configureStore from 'redux-mock-store';
-import PrintTitle from './PrintTitle';
 
-// Mock matchMedia function to simulate print media
+import PrintTitle from './';
+
+//Mock matchMedia function to simulate print media
 beforeAll(() => {
   window.matchMedia = jest.fn().mockImplementation(query => {
     return {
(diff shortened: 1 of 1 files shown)


AlanCalvillo:
just wondering if fixing the import wouldn't solve the issue instead of changing the coverage of the test? based on the logs from the [gh action](https://github.com/PADAS/das-web-react/actions/runs/11037506435/job/30658524859#step:8:205)that seemed to be the only issue 🤔

gaboDev:
> just wondering if fixing the import wouldn't solve the issue instead of changing the coverage of the test? based on the logs from the [gh action](https://github.com/PADAS/das-web-react/actions/runs/11037506435/job/30658524859#step:8:205)that seemed to be the only issue 🤔

The previous test case was not working, you are right the error in the build points to the import statement, but once fixed the test case was broken, so I decided to add just a simple working test case, also since the some coverage is missing I created a ticket to add it later

luixlive:
Do we understand what caused the test to fail?

gaboDev:
> Do we understand what caused the test to fail?

I do not have more context on this, this is [the original commit](https://github.com/PADAS/das-web-react/commit/564afe8fcf1a8f70e6cee2acee067a2a64176abe) as you can see the import state was already broken there and once fixed, the  existing test case was broken

luixlive (approved):
Approving to unblock the builds stopped by this issue. But @JoshuaVulcan should take a look at this once he is back since it seems related to a PR he sent a couple sprints ago.

AlanCalvillo (approved):

PR #1175, opened 2024-10-14
ERA-10108: Include sort order in saved event/patrol filters

gaboDev:
### What does this PR do?
- Adds sort filter to the redux store making it able to be persisted

### Relevant link(s)
* Tracking tickets: [ERA-10108](https://allenai.atlassian.net/browse/ERA-10108)
* [Hosted demo environments](https://era-10108.dev.pamdas.org)

[ERA-10108]: https://allenai.atlassian.net/browse/ERA-10108?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ColumnSort/index.js
+++ b/src/ColumnSort/index.js
@@ -11,7 +11,7 @@ import { ReactComponent as ArrowUp } from '../common/images/icons/arrow-up.svg';
 import { ReactComponent as CheckIcon } from '../common/images/icons/check.svg';
 import { ReactComponent as SortLines } from '../common/images/icons/sort-lines.svg';
 
-import { DEFAULT_EVENT_SORT, SORT_DIRECTION } from '../utils/event-filter';
+import { DEFAULT_EVENT_SORT, SORT_DIRECTION } from '../ducks/event-filter';
 
 import styles from './styles.module.scss';
 
@@ -21,7 +21,7 @@ const ColumnSort = ({ className, sortOptions, orderOptions, value, onChange }) =
   const [isSortUp, setSortDirection] = useState(value === SORT_DIRECTION.up);
 
   const [direction, selectedOption] = value;
-  const isSortModified = !DEFAULT_EVENT_SORT.includes(selectedOption);
+  const isSortModified = !DEFAULT_EVENT_SORT.find(({ value, key }) => value === selectedOption.value && key === selectedOption.key );
 
   const onClickSortOption = useCallback((option) => {
     if (option.value !== selectedOption.value){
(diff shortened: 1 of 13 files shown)


src/SideBar/useReportsFeed/index.js
```diff
@@ -19,8 +19,8 @@ const useReportsFeed = () => {
   const events = useSelector((state) => getFeedEvents(state));
   const userIsGeoPermRestricted = useSelector((state) => userIsGeoPermissionRestricted(state?.data?.user));
   const userLocationCoords = useSelector((state) => state?.view?.userLocation?.coords);
+  const feedSort = eventFilter.filter.sort ?? DEFAULT_EVENT_SORT;
```
gaboDev: I know this line should be replaced by setting the default value directly into `INITIAL_FILTER_STATE` but it was causing a bunch of errors due the `DEFAULT_EVENT_SORT` not being initialized when using the initial state, in order to solve that I should have to move around several declarations and impacting a lot of files where those consts were being imported I just wanted to avoid all those changes but not sure, If you want me I can go ahead and do that 🫡

JoshuaVulcan: It seems like you could just move some of the `const`s from the event filter utils file to somewhere initialized sooner and it wouldn't be that much hassle to me? Unless I'm missing something here.

luixlive: Don't we have a constants file?

src/ColumnSort/index.js
```diff
@@ -21,7 +21,7 @@ const ColumnSort = ({ className, sortOptions, orderOptions, value, onChange }) =
   const [isSortUp, setSortDirection] = useState(value === SORT_DIRECTION.up);
 
   const [direction, selectedOption] = value;
-  const isSortModified = !DEFAULT_EVENT_SORT.includes(selectedOption);
+  const isSortModified = !DEFAULT_EVENT_SORT.find(({ value, key }) => value === selectedOption.value && key === selectedOption.key );
```
luixlive, line 24: Seems like a good place to use `isEqual`

luixlive (commented):
The PR makes sense to me. The juicy part is in `useReportsFeed` where we change from a `useState` to the Redux `eventFilter`.

I feel like moving the constants from the utils to the ducks is just arbitrary and not really meaningful. I know we do it because of order the files get initialized, but this just feels hacky. Can we put those constants in a higher constant file that gets initialized as we need?

luixlive:
The PR is good! I'd just like to see if there is an alternative place to put those constants that makes more sense and I'll be ready to approve 😄

AlanCalvillo (approved):

JoshuaVulcan (approved):

PR #1183, opened 2024-10-30
ERA-10338: Audio ping for new InReach messages

luixlive:
### What does this PR do?
Add experimental feature to reproduce a sound everytime there is a new InReach message.

### How does it look
image

### Relevant link(s)
* [ERA-10338](https://allenai.atlassian.net/browse/ERA-10338)
* [Env](https://era-10338.dev.pamdas.org/)

The refactoring of `SoundNotificationsPlayer/index.js` to understand how the logic of each type of notification is organized, and then `SoundNotificationsPlayer/NewInReachMessageNotificationPlayer/index.js` which is the new type of notification.

The best way to emulate a message delivery is:
1- Create a new source provider that supports 2-way messaging
2- Create a new source (from the previous source provider) that also support 2-way messaging
3- Assign the source to a new subject
4- Go to https://era-10338.dev.pamdas.org/api/v1.0/docs/interactive/ and use the `POST api/v1.0/messages/` endpoint to send a message providing the following parameters (example):
Query
```
manufacturer_id: Ludwig Source (needs to match the source manufacturer id)
```

Body
```
{
  "message_type": "inbox",
  "text": "Hello!",
  "status": "received",
  "message_time": "2024-11-05T19:00:00.000Z",
  "additional": {}
}
```

[ERA-10338]: https://allenai.atlassian.net/browse/ERA-10338?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/SoundNotificationsPlayer/index.js
```diff
@@ -1,124 +1,59 @@
...
-    {subject.last_position?.geometry?.coordinates && <LocationJumpButton
-      bypassLocationValidation
-      className={styles.subjectLocationJumpButton}
-      onClick={onClickJumpToLocation}
-    />}
-  </div>;
-};
+export const SOUND_DEBOUNCE_TIME = 5000;
 
 const SoundNotificationsPlayer = () => {
```
luixlive, line 18: Hey @JoshuaVulcan I thought about how to simplify the pattern as you suggested. But using a hook didn't really feel right, since we already have `const [play] = useSound(ding);` which we can use anywhere. Also that would result on having the logic of our notifications dispersed in the code. Any file could call that hook without any kind of organization. Keeping a render-less component whose logic is to handle the notifications felt like the most organized way.

However, now the component just reads the user settings in the main component and abstracts the logic to debounce the sounds. But the internals of each type of notification is abstracted in different components so we keep the logics separated and avoid rendering hooks like `useSelector` or `useEffect` that are useless thanks to the conditional rendering of each `NotificationPlayer` component.

What do you think?

public/locales/ne-NP/menu-drawer.json
```diff
@@ -1,71 +1,72 @@
 {
+  "alertsModal": {
```
luixlive, line 2: This file was reformatted because it used indentation of 4 spaces. You can ignore the big diff.

public/locales/pt/menu-drawer.json
```diff
@@ -1,69 +1,70 @@
 {
+  "alertsModal": {
```
luixlive, line 2: Same here.

src/SideBar/SettingsPane/styles.module.scss
```diff
@@ -18,6 +18,7 @@
 
 .tab {
   padding: 1rem 1.5rem;
+  width: var(--sidebar-width);
```
luixlive, line 21: The settings panel was redimensioning its width in each tab. This keeps it static.

JoshuaVulcan: Any downsides here? overflow/trimming effects as a result of this?

JoshuaVulcan (approved):

src/SideBar/SettingsPane/styles.module.scss
luixlive: The issue we had is that it was increasing its width dynamically with its content. Long texts increased its width until it covered all the screen. I didn't see anything wrongly affected.

fonzy911 (approved):

PR #1185, opened 2024-11-01
ERA-10340: Ping and toast when radio state changes to red/emergency

luixlive:
### What does this PR do?
Play a ping and show a toast whenever a radio state changes to red (alarm).

### How does it look
image

### Relevant link(s)
* [ERA-10340](https://allenai.atlassian.net/browse/ERA-10340)
* [Env](https://era-10340.dev.pamdas.org/)

The transition: `src/NewEventNotifier/index.js` -> `src/SoundNotificationsPlayer/index.js`

The previous `NewEventNotifier` used to handle just the scenario for playing a ping whenever there was a new event, but the component was redefined to be a sound notifications player that also handles the subscription and pings for radio state changes.

The best way to emulate a radio changing its state is in https://era-10340.dev.pamdas.org/api/v1.0/docs/interactive/ 

Go to the `/api/v1.0/sensors/{sensor_type}/{provider_key}/status/` endpoint and trigger it with the following parameters:
Parameters:
```
sensor_type: generic
provider_key: ludwig
```

Request body:
```
{
  "message_key": "observation",
  "location": { "lat": 10, "lon": 10 },
  "recorded_at": "2024-11-01T10:00:00.000000Z",
  "manufacturer_id": "Ludwig Source",
  "subject_name": "Ludwig",
  "subject_subtype": "radio",
  "model_name": "Ludwig Source",
  "source_type": "gps-radio",
  "additional": {
    "event_action": "device_state_changed",
    "radio_state": "alarm",
    "radio_state_at": "2024-11-01T10:00:00.000000Z"
  }
}
```

Of course you can use a different source provider, source and subject. But that request would modify the radio state of the existing subject I created named Ludwig. Also, in order to re-trigger the notification, you will need to first send another request making the radio state green (`"radio_state": "online-gps"`).

Remember that in order to hear the ping and see the toast you need to first enable the feature:
image

**Note:** The translations where added to https://docs.google.com/spreadsheets/d/1ljvgLO2N8gXLk8350953OCi2J_WI0Mmp_mhlcaixp1M/edit?gid=1104334091#gid=1104334091

[ERA-10340]: https://allenai.atlassian.net/browse/ERA-10340?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/SoundNotificationsPlayer/index.js
```diff
@@ -0,0 +1,124 @@
+import React, { useEffect, useRef } from 'react';
```
JoshuaVulcan, line 1: This component seems a bit muddled to me. We now have:

1. Two main effect hooks, for playing sounds under different conditions, each with some bespoke qualifying logic but duplicating a lot of the same 
2. One extra toast component defined inline

It seems like the refactor could go a step further -- a `useNotificationSound` hook which receives a qualifying function to determine the conditions under which a notification plays, and a separate hook (or some other means, I don't have strong feelings on the **how**) which determines when to show the toast.  

Then you can pull the duplicate code such as
```
 setTimeout(() => {
        canPlayNewEventNotificationSound.current = true;
      }, SHOULD_PLAY_DEBOUNCE_MS);
```

and have more legible code handling these sounds, without expanding this individual component file if/whenever we want to have a new sound-triggering condition.

Does this make sense to you?

JoshuaVulcan (commented):
Looks mostly good! I think you could take the refactor a step further, rather than tweaking and moving the original effect-hook-driven code I wrote a few years back.

Separately, I just want to say it's awesome you added the i18n for new strings. Major 🌮s to you!

src/SoundNotificationsPlayer/index.js
luixlive: Yes, I like the idea. However I have a proposal for you. Doing those changes here implies me doing double work: the refactor here and then applying the changes in the other ticket I have for playing sounds for new messages. Are you ok if we keep this one with this pattern for now and just focus on feedback regarding the logic of when to play the sound? And I will add a cool pattern in the next PR, which btw, I'm already working on it @JoshuaVulcan

public/locales/sw/menu-drawer.json
```diff
@@ -4,7 +4,7 @@
     "iframeTitle": "Sanidi arifa zako za EarthRanger"
   },
   "betaToggles": {
-    "patrolUI": "UI Mpya wa Fomu ya Doria",
+    "playSoundForAlarmRadioState": "Cheza sauti na onyesha arifa wakati hali ya redio inabadilika kuwa nyekundu (kengele)",
```
gaboDev, line 7: Just out of curiosity, Where did you get all those new translations? using just a regular online translator?

I'm asking just to be aware of the process We are doing in this scenario

gaboDev (approved):

public/locales/sw/menu-drawer.json
luixlive: For english, spanish and french I have confidence. For the others, I used chatGPT, since I think it understands more the context of the text than a simple translator

fonzy911 (approved):

PR #1186, opened 2024-11-05
ERA-10485: Support EFB schemas in das-web-react form rendering -  Form Component

gaboDev:
### What does this PR do?
- It adds the SchemaForm component which handles the general logic to render EFB schemas 
- It adds the base pattern to manage EFB schema forms 

### Relevant link(s)
* Tracking tickets: [ERA-10485](https://allenai.atlassian.net/browse/ERA-10485)

- For this PR still there is nothing to QA to test, all these changes are the base for adding EFB fields support

[ERA-10485]: https://allenai.atlassian.net/browse/ERA-10485?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/ReportManager/DetailsSection/index.js
```diff
@@ -10,7 +10,7 @@ import { useTranslation } from 'react-i18next';
 import { ReactComponent as PencilWritingIcon } from '../../common/images/icons/pencil-writing.svg';
 
 import { calcGeometryTypeForReport } from '../../utils/events';
-import { EVENT_FORM_STATES, VALID_EVENT_GEOMETRY_TYPES } from '../../constants';
+import { EFB_SCHEMA_DRAFT, EVENT_FORM_STATES, VALID_EVENT_GEOMETRY_TYPES } from '../../constants';
```
JoshuaVulcan: `EFB` doesn't belong anywhere in the code here. We're matching a new spec for event type schemas, and the fact that these come from a form builder UI using terminology which is entirely internal to us isn't good naming convention.

src/ReportManager/ReportDetailView/index.js
```diff
@@ -727,6 +727,64 @@ const ReportDetailView = ({
...
   const isReadOnly = reportSchemas?.schema?.readonly;
 
+  const efb = {
+    'json': {
+      '$schema': 'https://json-schema.org/draft/2020-12/schema',
+      'additionalProperties': false,
+      'properties': {
+        'this_is_a_text': {
+          'default': '',
+          'deprecated': false,
```
JoshuaVulcan: `deprecated: false` doesn't necessarily belong in a schema, let alone a fixture for testing. We only want this to validate that `deprecated: true` fields are hidden from display and rendering.

src/ReportManager/DetailsSection/index.js
luixlive: I agree, but also I wonder, how should we differentiate those two flows then? They read different schemas, they use different components... We sure need to call them differently somehow.

src/ReportManager/DetailsSection/index.js
```diff
@@ -210,7 +211,7 @@ const DetailsSection = ({
         : null}
     </div>
 
-    {!!formSchema && <Form
+    {!!formSchema && !isEFBSchema && <Form
       className={`${styles.form} ${reportForm.is_collection ? styles.hidden : ''}`}
       disabled={formSchema?.readonly}
```
luixlive, line 221: Hmm seeing that the old version of the schema supports this makes me wonder... @JoshuaVulcan @tiffanynwong did we forget to add a checkbox in the event type page to make it `readonly`? Don't we want to support this read-only form any longer?

src/ReportManager/DetailsSection/index.js
```diff
@@ -234,6 +235,12 @@ const DetailsSection = ({
       <button ref={submitFormButtonRef} type="submit" />
     </Form>}
 
+    {!!formSchema && isEFBSchema &&
+      <EFBForm schema={formSchema}
```
luixlive: I want to know, what do you think of the props that we pass to that other Form component? Like `onFormSubmit`, `onFormError`, `formValidator`... I know we don't need to follow that old way to interact with the form, but we need somehow to make it work similar @gaboDev

src/ReportManager/DetailsSection/index.js
JoshuaVulcan: This will be coming from `v2` APIs, so the distinction point is from the very beginning. You can key off of the version of the schema, the API endpoint source, where the data comes from in the redux store, etc etc. Tons of good clear indicators of what "version" of the schema you're using, without getting stuck on a weird naming convention

src/ReportManager/EFBForm/FormSchemaContext/index.js
```diff
@@ -0,0 +1,29 @@
+import React, { createContext } from 'react';
+
+export const FormSchemaContext = createContext(null);
+
+const FormSchemaContextProvider = ({ schema, onFieldChange, formData, children }) => {
```
luixlive: I like this idea! A way to provide friendly methods and hide the internals of the schema.

Just one thing, you are forgetting to provide the `order` UI schema property that tells the ordering of the sections: https://allenai.atlassian.net/wiki/spaces/ER/pages/30116741125/Form+Builder+Generated+JSON+Schemas#Order

Also, we don't need to keep the `{ json: {...}, ui: {...} }` format for the fields anymore. Since the responsibility here is just to render the form, we can merge them into a single object that contains all the information needed to render the component. The tricky part is that every field will have different properties, so we will need to clean one by one. But the good thing is that we already have this logic covered in the EFB: https://github.com/PADAS/report-form-builder/blob/master/app/components/EventTypeForm/utils/makeFieldsFromSchema/index.js#L40

An example of how that would look for a text field is:
```
// Outside of the component
const buildTextFieldObject = (fieldName, jsonSchema, uiSchema, isRequired) => ({
  defaultInput: jsonSchema.default,
  description: jsonSchema.description,
  inputType: uiSchema.inputType,
  isActive: !jsonSchema.deprecated,
  isRequired,
  label: jsonSchema.title,
  placeholder: uiSchema.placeholder,
  value: fieldName,
});

...

  const getFieldProps = (fieldName) => {
    const jsonProps = schema.json.properties[fieldName];
    const isRequired = schema.json.required.includes(fieldName);
    const uiProps = schema.ui.fields[fieldName];

    switch (uiProps.type) {
      case 'TEXT':
        return buildTextFieldObject(fieldName, jsonProps, uiProps, isRequired);
      ...
    }
  };
```

This way we match the field objects just like in the EFB reducing the complexity between the two repos! You can see the full lists of props that each field needs to have here: https://github.com/PADAS/report-form-builder/blob/master/app/components/EventTypeForm/FormBuilder/canvasFieldDetailsFactory/index.js

luixlive: Let me know what you think. If you like it, but you think it will take you long, we don't need to do it now, we can add each field type function as we work on that field type. But you can at least leave that switch with its // TODO

src/ReportManager/EFBForm/fields/Header/index.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+
+const Header = ({ label, size, section, name }) => {
```
luixlive: I would recommend to keep the naming intact everywhere to make it as easy as possible to future devs to understand what's going one. Here I would call it `fieldName` since that is how we call it everywhere, instead of simply `name`

luixlive: By the way, shouldnt' this component just receive the `fieldName` as prop and then call `getHeaderProps(fieldName)` from the context method? I see you are already expecting to receive label, size, section...

src/ReportManager/EFBForm/fields/Section/index.js
```diff
@@ -0,0 +1,43 @@
...
+const SECTION_ITEM_TYPES = {
+  HEADER: 'header',
+  FIELD: 'field'
+};
+
+const FORM_FIELDS = {
+  'TEXT': Text
+};
+
+const Section = ({ id }) => {
```
luixlive: Again, here you changed the `fieldName` now to `id`. Let's keep it consistent.

src/ReportManager/EFBForm/FormSchemaContext/index.js
```diff
@@ -0,0 +1,29 @@
...
+
+  const getSectionProps = (sectionId) => schema.ui.sections[sectionId];
```
luixlive: Here you call it sectionId, but in the header you call it headerName and in fields fieldName haha. Can we call it sectionName here to match?

src/ReportManager/EFBForm/fields/Section/index.js
```diff
@@ -0,0 +1,43 @@
...
+  const { getSectionProps, getFieldProps, getHeaderProps } = useContext(FormSchemaContext);
+  const sectionProps = getSectionProps(id);
+
+  const renderColumnSectionItems = (column) => column.map(({ name, type }) => {
+    const isFieldType = type === SECTION_ITEM_TYPES.FIELD;
+    const itemProps = isFieldType
+      ? getFieldProps(name)
+      : getHeaderProps(name);
```
luixlive: Ok this answers one of my other questions, that's why the header was receiving its props already. But why not to keep the same pattern eveywhere? Just send the name to the component and let the component fetch its props? So we only have to: `<SectionItem fieldName={fieldName} key={name} />` and the SectionItem can read its own props.

src/ReportManager/EFBForm/index.js
```diff
@@ -0,0 +1,37 @@
+import React, { useEffect, useState } from 'react';
+
+import FormSchemaContextProvider from './FormSchemaContext';
+import Section from './fields/Section';
+
+const EFBForm = ({ schema, formData: initialFormData = {}, isDisabled, className, onChange, onError }) => {
+  const [formData, setFormData] = useState(initialFormData);
```
luixlive: What's the purpose of having the form data controlled here? WHy not to simply call `onChange` and let the parent keep the changes?

src/ReportManager/EFBForm/FormSchemaContext/index.js
```diff
@@ -0,0 +1,29 @@
...
+
+  const getHeaderProps = (headerName) => schema.ui.headers[headerName];
+
+  const getFieldProps = (fieldName) => {
```
luixlive: Another thing here. The term `props` is something specifically related to React. Why are we mixing our language with React language? Why not `getSectionDetails`, `getHeaderDetails`, etc... Which is how we call them in the EFB repo actually.

luixlive:
Nice! I think this is clearly taking form. I really liked the idea of the context. But I think you can take it one step further with a custom hook instead of keeping with the regular context pattern. I mean, something like:
```
const Header = ({ fieldName }) => {
  // This hook finds out if the field is a header, section or field and it fetches its details from the context provider, making it a
  // trivial dev experience for the ones using it!
  const headerDetails = useFieldDetails(fieldName);
};
```

I left a few other comments with questions and suggestions to keep our ubiquitous language matching the other repo. But overall, nice job 👍

src/ReportManager/DetailsSection/index.js
luixlive: Hmm I'm not sure if I get you or if we are not talking about the same thing. I know the data will come from a v2 endpoint but how do I know that when reading this file?

We already have a Form component, Gabo is adding a second one. My question is how can we call the new`<Form />` component in a way that it tells you that it's the new component for the new form. `FormV2`? Hahaha. And even if reading from a new redux store is a way to infer the version of the schema, I don't think that's a good way to expect devs to understand what's going on. I'd rather "scream" with file names, variable names, method names, etc... Making the code tell you what is going on.

src/ReportManager/EFBForm/fields/Header/index.js
gaboDev: Sure, I have moved all that, now all field props are being fetched from context only

src/ReportManager/EFBForm/index.js
gaboDev: I haven't get there yet, after addressing the feedback of this round I will return to take a look to the behavior of the form component 🫡

src/ReportManager/EFBForm/FormSchemaContext/index.js
gaboDev: Done 🫡

gaboDev:
Just a heads up, this PR has nothing that can be tested by our QA folks, I started working on the primitive fields support I put that branch behind this one, there will be actual testable features, I used this ticket to set the foundations 

@luixlive @JoshuaVulcan

gaboDev:
Before I got too deep on adding support for primitive fields, I know the idea is to make the form in das-web-react look like the one in EFB, the most easy way to do it is to install React MUI in this repo, but that will increase the loading time of the page for the users, also I've been searching for some lightweight CSS material implementation, I would like to hear your opinion on this matter, also I think that an option can be use CSS on our own and get something really good 👀 

@luixlive @JoshuaVulcan

gaboDev:
All the components inside `fields/` have partial implementation and are missing test coverage, that will be addressed in their respective PRs, just add them to make the patter/approach easier to follow 🫡

src/ReportManager/SchemaFormContext/utils.js
```diff
@@ -0,0 +1,8 @@
+
```
gaboDev: I just realized I was doing a circular reference with this utils methods, I'm refactoring the code to avoid this, also realized I was missing logic of active/inactive fields for rendering and form validation, on it 🫡

src/ReportManager/DetailsSection/index.js
```diff
@@ -234,6 +236,12 @@ const DetailsSection = ({
       <button ref={submitFormButtonRef} type="submit" />
     </Form>}
 
+    {!!formSchema && isEFBSchema &&
+      <SchemaFormContextProvider schema={formSchema} onFieldChange={onFormChange} formData={reportForm.event_details} onFormSubmit={onFormSubmit} >
```
luixlive: Why don't we have the implementation of the context provider inside the `SchemaForm` component? We can send all these props to the `SchemaForm` and inject them to the context value there:
```
// SchemaForm/index.js
const SchemaForm = ({ ... }) => {
  ...
  return <SchemaFormContextProvider value={{ ... }}>
    ...
  </SchemaFormContextProvider>;
};
```

This way the parent doesn't have to implement the context, which is cleaner.

src/ReportManager/DetailsSection/index.js
```diff
@@ -65,7 +67,7 @@ const DetailsSection = ({
 }, ref) => {
   const dispatch = useDispatch();
   const { t } = useTranslation('reports', { keyPrefix: 'reportManager.detailsSection' });
-
+  const isEFBSchema = true;
```
luixlive: I think you forgot to change this.

src/ReportManager/ReportDetailView/index.js
```diff
@@ -727,6 +727,64 @@ const ReportDetailView = ({
 
   const isReadOnly = reportSchemas?.schema?.readonly;
 
+  const efb = {
```
luixlive: We should remove this before merging. Or at least adding some kind of temporal feature flag to use only in development to inject this data.

src/ReportManager/SchemaFormContext/SchemaForm/fields/Section/index.js
```diff
@@ -0,0 +1,41 @@
...
+  HEADER: 'header',
+  FIELD: 'field'
+};
+
+const FORM_FIELDS = {
+  [FORM_FIELDS_TYPES.TEXT]: Text
+};
+
+const Section = ({ sectionName }) => {
+  const { getFieldUIType, isFieldActive } = useContext(SchemaFormContext);
```
luixlive: Why are we receiving these method through context? Both of them sound like utilities. Sending them through context has the potential to cause rerenders. Static util methods don't.

src/ReportManager/SchemaFormContext/SchemaForm/fields/Section/index.js
```diff
@@ -0,0 +1,41 @@
...
+  const sectionDetails = useFieldDetails(sectionName);
+
+  const renderColumnSectionItems = (column) => column.map(({ name, type }) => {
+    if ( type === SECTION_ITEM_TYPES.FIELD && isFieldActive(name) ){
+      const Field = FORM_FIELDS[getFieldUIType(name)];
```
luixlive: If we already have a getter, why then we need to search in another object `FORM_FIELDS`? Why don't the getter just returns the field at once?

src/ReportManager/SchemaFormContext/SchemaForm/fields/Text/index.js
```diff
@@ -0,0 +1,41 @@
...
+  [VALIDATION_ERROR_TYPES.REQUIRED]: (fieldValue, fieldName, schema) => {
+    return isSchemaFieldRequired(schema, fieldName) ? !fieldValue : false;
+  }
+};
+
+const Text = ({ fieldName }) => {
+  const { onFieldChange } = useContext(SchemaFormContext);
+  const textFieldDetails = useFieldDetails(fieldName);
+
+  const handleOnChange = (e) => onFieldChange(fieldName, e.currentTarget.value);
```
luixlive: nit: `e` is not a great variable name. I know it's common to call it like that but still reduces readability of the code.

src/ReportManager/SchemaFormContext/SchemaForm/fields/Text/index.js
```diff
@@ -0,0 +1,41 @@
...
+
+  return <div data-testid={`schema-form-text-field-${fieldName}`}>
+    <label htmlFor={fieldName}>{textFieldDetails.label}</label>
+    <input id={fieldName}
+           value={textFieldDetails.value}
```
luixlive: Another nit: this kind of indentation wastes lots of space. All those empty spaces at the beginning of the line pushes the code further to the right, which then the linter will want to break because it passes the max chars per line limit.

src/ReportManager/SchemaFormContext/SchemaForm/fields/fieldDetailsFactory/index.js
```diff
@@ -0,0 +1,12 @@
+
+export const textFieldDetailsFactory = ({ description, title, deprecated, default: defaultValue }, { inputType, placeholder, isRequired }, formValue, formError) => ({
```
luixlive: nit: destructuring here makes this line longer and doesn't provide information on what the arguments of the function are. A developer would have to infer which each of those args is. Why not to call them by their name `(jsonSchema, uiSchema, value, error)`? We can then read them from their objects.

src/ReportManager/SchemaFormContext/SchemaForm/fields/fieldDetailsFactory/index.test.js
```diff
@@ -0,0 +1,32 @@
+import { textFieldDetailsFactory } from './index';
+
+describe('ReportManager - SchemaForm - fieldDetailsFactory', () => {
+
+  test('builds details object of text field properly', () => {
+    const schemaDetails = {
```
luixlive: To keep our language consistent, this should be called `jsonSchema`

src/ReportManager/SchemaFormContext/SchemaForm/fields/fieldDetailsFactory/index.test.js
```diff
@@ -0,0 +1,32 @@
...
+      description: 'A hint of the field',
+      title: 'Important Text',
+      deprecated: true,
+      default: 'initial value'
+    };
+    const uiDetails = {
```
luixlive: And this one `uiSchema`. Remember that there are two things: the schema and the field details. Each one of those can be converted to the other. The "schema" is what we store in the DB and it respects the JSON schema protocol, with the UI complementing it with what we need that is not supported by the protocol. The "field details" is how we call the object we construct with the information on how to render the field. Those two terms are properly separated in the EFB. In there, the field details come from a factory and it is handled by the drag and drop interface, and the schema is generated by a hook `useSchema` from the details object. In here, we are mixing those two concepts in a few places, which will cause confusion. 
Summary:
- Schema: stored in the DB, it is built by the jsonSchema (which adders to the latest draft of JSON schema) and the uiSchema (which was defined by us complementing the json schema)
- Field details: a single object with all the information needed to render a field (type, placeholder, label, etc...)
One of those can be transformed to the other when it is convenient. The schema is useful for the backend to be stored and to run the schema validations, the field details is convenient for the frontend to render the field.

src/ReportManager/SchemaFormContext/SchemaForm/index.js
```diff
@@ -0,0 +1,70 @@
...
+import styles from './styles.module.scss';
+import { FORM_FIELDS_TYPES } from '../../constants';
+import { TextFieldValidators } from './fields/Text';
+import { getSchemaFieldUIType, isFieldActive } from './SchemaFormContext/utils';
+
+const FormFieldValidators = {
+  [FORM_FIELDS_TYPES.TEXT]: TextFieldValidators
+};
+
+const SchemaForm = ({ schema, formData = {}, onFormSubmit, className, onChange }, ref) => {
```
luixlive: This forwardedRef doesn't follow the classic pattern of being a ref injected to the whole component. It is being injected just to the button, and there's no way the parent can know this explicitly. An alternative to make what that ref is explicit is to use a renderProp:
```
const Parent = () => {
  const submissionButtonRef = useRef();

  return <SchemaForm renderSubmissionButton={() => <button ref={submissionButtonRef} ... />} />;
};

const SchemaForm = ({ renderSubmissionButton }) => {
  ...
  return <form onSubmit={handleOnSubmit} className={className}>
    {
      schema?.ui?.order.map((sectionName) => <Section sectionName={sectionName} key={sectionName} />)
    }
    {renderSubmissionButton()}
  </form>;
};
```

Now you have the ref in the parent, where you can hide the button and trigger the submission of the form programmatically!

src/ReportManager/SchemaFormContext/SchemaForm/index.js
```diff
@@ -0,0 +1,70 @@
...
+
+  const [formErrors, setFormErrors] = useState({});
+
+  const handleOnSubmit = (e) => {
+    e.preventDefault();
+
+    const formErr = { ...formErrors };
+
+    for (const [fieldName, fieldValue] of Object.entries(formData)) {
```
luixlive: nit: not really an issue... But why are we using old `for`s instead of iterator methods?

src/ReportManager/SchemaFormContext/SchemaForm/index.js
```diff
@@ -0,0 +1,70 @@
...
+        const fieldUIType = getSchemaFieldUIType(schema, fieldName);
+        const validators = FormFieldValidators[fieldUIType];
+
+        for (const [errorType, validatorFn] of Object.entries(validators)) {
+          const hasError = validatorFn(fieldValue, fieldName, schema);
+          if (hasError){
+            formErr[fieldName] = errorType;
+            break; // breaking the loop for the current field to avoid handling multiple errors per field on screen
+          } else {
+            delete formErr[fieldName];
```
luixlive: What was the purpose of creating a `const formErr = { ...formErrors };` if anyway we are going to add manually the errors that are still there and delete the ones that are not? Feels like we could start this formErrors object from scratch.

src/ReportManager/SchemaFormContext/SchemaForm/index.test.js
```diff
@@ -0,0 +1,155 @@
+import React from 'react';
+import userEvent from '@testing-library/user-event';
+
+import { render, screen, within } from '../../test-utils';
+
+import SchemaForm from './';
+
+describe('ReportManager - SchemaForm', () => {
+
+  const efb = {
```
luixlive: This doesn't feel like a proper name for this variable. Why not `schema` since that is what it is?

src/ReportManager/SchemaFormContext/SchemaForm/index.test.js
```diff
@@ -0,0 +1,155 @@
...
+    const onFormSubmit = jest.fn();
+
+    renderSchemaForm({
+      ...initialProps,
+      onFormSubmit,
+    });
+
+    const inputField = screen.getByTestId('schema-form-text-field-input-this_is_a_text');
+
+    await userEvent.type(inputField, '{enter}');
```
luixlive: Do we want to do a form submission on enter?

src/ReportManager/SchemaFormContext/SchemaForm/useFieldDetails/index.js
```diff
@@ -0,0 +1,15 @@
+import { useContext } from 'react';
+
+import { SchemaFormContext } from '../SchemaFormContext';
+
+const useFieldDetails = (fieldName) => {
+  const { getSectionDetails, getFieldDetails, getHeaderDetails, isSection, isField } = useContext(SchemaFormContext);
```
luixlive: Again, some of these functions should be utilities, not context passed methods. Context should not be used as a way to pass random stuff, but specifically state related variables that are expected to cause re-renders.

src/ReportManager/SchemaFormContext/SchemaForm/useFieldDetails/index.test.js
```diff
@@ -0,0 +1,131 @@
+import React from 'react';
+import { renderHook } from '@testing-library/react-hooks';
+
+import useFieldDetails from './';
+import SchemaFormContextProvider from '../SchemaFormContext';
+
+describe('ReportManager - SchemaForm - useFieldDetails', () => {
+
+  const efb = {
```
luixlive: Again `efb`

src/ReportManager/SchemaFormContext/index.js
```diff
@@ -0,0 +1,46 @@
+import React, { createContext } from 'react';
```
luixlive: Continuing with another of my comments, I think it would be better to call this folder `SchemaForm` and export that component while keeping the context as an internal implementation. It's weird that we are calling it context and saving everything that is not context below it, and then importing the `SchemaForm` component from here and rendering it. That is the most important component, and everything else should be an implementation detail of it. Not the context it uses.

src/ReportManager/SchemaFormContext/index.test.js
```diff
@@ -0,0 +1,226 @@
+import React, { useContext } from 'react';
+import { render, screen } from '../../../test-utils';
+
+import SchemaFormContextProvider, { SchemaFormContext } from './index';
+import userEvent from '@testing-library/user-event';
+import { FORM_FIELDS_TYPES } from '../../../constants';
+
+describe('ReportManager - SchemaForm - SchemaFormContext', () => {
+
+  const efb = {
```
luixlive: x3

src/ReportManager/SchemaFormContext/utils.js
```diff
@@ -0,0 +1,8 @@
...
+export const isSchemaFieldRequired = (schema, fieldName) => schema.json.required.includes(fieldName);
+
```
luixlive: The utils methods should not have any external dependencies otherwise than external modules or constants. That way there wouldn't be any circular dependencies. And everyone should import these methods from here instead of expect them to come from context 👍

src/constants/index.js
```diff
@@ -350,3 +350,8 @@ export const EVENT_SORT_OPTIONS = [
 export const SORT_DIRECTION = { up: 'up', down: 'down' };
 
 export const DEFAULT_EVENT_SORT = [SORT_DIRECTION.down, EVENT_SORT_OPTIONS[0]];
+
+export const FORM_FIELDS_TYPES = {
```
luixlive: Is this going to be used anywhere int he app? If not, this is probably a too broad place to put it. We could have a constants file specifically under the `SchemaForm` component.

luixlive (changes requested):
Lots of comments, but the pattern is there. I just think we still need to tweak some wording, files and code style to keep our code and language consistent across the EFB and here. Also, this component will be the main point of rendering for the new schemas, the better written and optimized we leave it, the easier and smoother the development and execution of the new schemas will be. Well done @gaboDev !

src/ReportManager/SchemaFormContext/index.js
gaboDev: initially was like that, but I was mixing responsibilities, let me organize it as it was

src/ReportManager/SchemaFormContext/SchemaForm/index.test.js
gaboDev: Currently is behaving like that in the old schemas for an event in the web app

src/ReportManager/SchemaFormContext/index.js
luixlive: If we are able to hide the context under the form component (which we should), it will be way clearer to simply call `SchemaForm` and keep everything under it as implementation details. Btw, `SchemaForm` is the definitive name we decided to differentiate old from new versions? Just asking, kind of like it.

src/ReportManager/SchemaFormContext/SchemaForm/fields/Text/index.js
gaboDev: Done

src/ReportManager/SchemaFormContext/SchemaForm/fields/fieldDetailsFactory/index.js
gaboDev: Done

src/ReportManager/SchemaFormContext/SchemaForm/index.js
gaboDev: Got it, I did not put much attention when I was trying to homologate the new form behavior with the old one, but yeah, this approach makes way more sense, thanks 🫡

src/ReportManager/SchemaFormContext/SchemaForm/fields/Section/index.js
gaboDev: I have extract all those methods to a utils file, I'll let you know when the changes are pushed

src/ReportManager/ReportDetailView/index.js
gaboDev: Sure, also I was thinking about adding a FF for better management, also, since We are going to use mock data to tests these form I was thinking on adding a select control, that will allow QA and us to select a mock object to render a form based on the selection, by doing this We can use several configurations of schemas within a feature env. 

Let me finish with all major things regarding this ticket, once that done I will elaborate a little bit more on this idea

src/ReportManager/SchemaFormContext/SchemaForm/index.js
gaboDev: Just for personal preference, I have no problem changing to iterator methods

gaboDev: Actually, I remember I wanted to break the iteration when a error was found, but that happens only in the second level for loop, so not sure I the code would look good mixing both techniques

src/ReportManager/SchemaFormContext/SchemaForm/useFieldDetails/index.js
gaboDev: Already changed it 🫡

src/ReportManager/SchemaFormContext/index.js
gaboDev: That is right, that was my intention to hide implementation behind `<SchemaForm />` the file you were sawing was a WIP for something I tried, but at the end I returned to the initial approach, sorry about that I forgot to mentioned it

gaboDev: And yes, SchemaForm is the name I'm proposing haha

gaboDev:
@luixlive I have addressed most of the feedback, I left some things out I would like first to have a green light before getting into these

1. Schema mocks naming and condition: you made some comments about this, my idea is to add a FF 
2. Failing unit test, I would like to work on fix them once We hit common ground
3. Implementing a mock selector inside the event form to facilitate QA explorations

src/ReportManager/DetailsSection/index.js
luixlive: Are you going to clean these once the PR is approved @gaboDev ?

src/ReportManager/SchemaForm/SchemaFormContext/index.js
```diff
@@ -0,0 +1,50 @@
...
+        ...uiDetails,
+        isRequired
+      }, formValue, formError);
+    }
+
+    default: return {};
+    }
+  };
+
+  const getSchema = () => ({ ...schema });
```
luixlive: Is there a reason to spread the schema instead of just returning it? Remember that spreading just updates the pointers of the top level properties, but the inner properties below `ui` and `json` will still be passed by reference. If you want a cloned object you may want to use a helper like `_.clone`. But is it necessary?

src/ReportManager/SchemaForm/SchemaFormContext/index.test.js
```diff
@@ -0,0 +1,226 @@
+import React, { useContext } from 'react';
+import { render, screen } from '../../../test-utils';
+
+import SchemaFormContextProvider, { SchemaFormContext } from './index';
+import userEvent from '@testing-library/user-event';
+import { FORM_FIELDS_TYPES } from '../constants';
+
+describe('ReportManager - SchemaForm - SchemaFormContext', () => {
+
+  const efb = {
```
luixlive: Lets change the names of these to schema. `efb` is not appropriate since that is not what it is.

src/ReportManager/SchemaForm/SchemaFormContext/index.test.js
```diff
@@ -0,0 +1,226 @@
...
+    };
+
+    renderSchemaFormContext(undefined, Component, { onClick });
+
+    await userEvent.click(screen.getByRole('button'));
+
+    expect( onClick ).toHaveBeenCalledWith(true);
+  });
+
+});
```
luixlive, line 114: We are missing a test for `onFieldChange`

src/ReportManager/SchemaForm/SchemaFormContext/useFieldDetails/index.test.js
```diff
@@ -0,0 +1,131 @@
+import React from 'react';
+import { renderHook } from '@testing-library/react-hooks';
+
+import useFieldDetails from './';
+import SchemaFormContextProvider from '../SchemaFormContext';
+
+describe('ReportManager - SchemaForm - useFieldDetails', () => {
+
+  const efb = {
```
luixlive: x2

src/ReportManager/SchemaForm/index.js
```diff
@@ -0,0 +1,61 @@
...
+        const fieldUIType = getFieldUIType(fieldName);
+        const validators = FormFieldValidators[fieldUIType];
+
+        for (const [errorType, validatorFn] of Object.entries(validators)) {
+          const hasError = validatorFn(fieldValue, fieldName, schema);
+          if (hasError){
+            formErr[fieldName] = errorType;
+            break; // breaking the loop for the current field to avoid handling multiple errors per field on screen
+          } else {
+            delete formErr[fieldName];
```
luixlive: If `const formErr = {};` is starting now from scratch, why we still `delete` stuff? This shouldn't be necessary right?

src/ReportManager/SchemaForm/index.test.js
```diff
@@ -0,0 +1,155 @@
+import React from 'react';
+import userEvent from '@testing-library/user-event';
+
+import { render, screen, within } from '../../test-utils';
+
+import SchemaForm from './';
+
+describe('ReportManager - SchemaForm', () => {
+
+  const efb = {
```
luixlive: x3

src/ReportManager/SchemaForm/utils/index.js
```diff
@@ -0,0 +1,22 @@
...
+
+export const isSection = (fieldName, schema) => !!schema.ui.sections[fieldName];
+
+export const isField = (fieldName, schema) => !!schema.ui.fields[fieldName];
+
+export const getFieldUIType = (fieldName, schema) => schema.ui.fields[fieldName].type;
+
+export const isFieldActive = (fieldName, schema) => !schema.json.properties[fieldName].deprecated;
+
+export const getFormFieldComponent = (fieldName, schema) => FORM_FIELDS_COMPONENTS[getFieldUIType(fieldName, schema)];
```
luixlive, line 23: All these are missing tests 👍

luixlive (commented):
Nothing blocking. My comments this time are mostly suggestions. I was going to approve it already but given I see a few tests missing I will wait for that before giving it my ✅

src/ReportManager/SchemaForm/SchemaFormContext/index.js
gaboDev: There are places where the schema is required, my intention was to expose the schema object without allowing mutation to the original object which is the source of truth, if you agree on this I'll update this method accordingly

src/ReportManager/SchemaForm/SchemaFormContext/index.test.js
gaboDev: I left all feedback of test out just to wait for your approval on the second iteration, I'm working on fixing tests rn

src/ReportManager/SchemaForm/utils/index.js
gaboDev: That is is right, some of them have already tests but in another file, I'll proceed to work on those

src/ReportManager/SchemaForm/SchemaFormContext/index.js
luixlive: I supposed it was to protect it from mutation. But again, that will only work for the top level properties of the object "ui" and "json". When you spread, JS creates copies of the first level of the properties, but not for the second level onwards:
```
// Original schema
const schema = {
  json: { /* Deeper properties */ },
  ui: { /* Deeper properties */ },
};

// This method of clonation is "shallow", it copies the content of the first level properties "json" and "ui", but
// the deeper object properties will still be copied by reference (same pointers, meaning that if you modify them
// in "shallowSchemaClone" they will also be modified in the original "schema"!!
const shallowSchemaClone = { ...schema };

// This won't modify the original "schema.json"
shallowSchemaClone.json = { /* New object */ }

// This WILL modify the original "schema.json.properties" because it is copied by reference
shallowSchemaClone.json.properties = { /* New object */ };

// Real clonation at all levels, you can modify any property and it won't be modified in the original "schema"
const deepSchemaClone = _.clone(schema);

// This won't modify the original "schema.json.properties" because it is copied by value
shallowSchemaClone.json.properties = { /* New object */ };
```

luixlive: However, deep cloning can be expensive. If we are going to call this method several times it can have impact on performance because the schemas can go infinitely deep (thanks collections). So, do we really need to clone it? I also like defensive code, but maybe here the best is just to make sure nobody modifies the schema ever (we shouldn't, there's no reason to do so).

src/ReportManager/DetailsSection/index.js
```diff
@@ -210,7 +215,7 @@ const DetailsSection = ({
         : null}
     </div>
 
-    {!!formSchema && <Form
+    {!!formSchema && !isEFBSchema && <Form
```
JoshuaVulcan: `EFB` doesn't belong in the code anywhere here. You should be considering which version the schema is and nothing else

src/ReportManager/SchemaForm/SchemaFormContext/index.js
```diff
@@ -0,0 +1,52 @@
...
+import { isFieldRequired } from '../utils';
+import {cloneDeep} from "lodash-es";
+
+export const SchemaFormContext = createContext(null);
+
+
+const SchemaFormContextProvider = ({ schema, onFormChange, formData, formErrors, children }) => {
+
+  const onFieldChange = (field, value) => {
+    onFormChange({
```
JoshuaVulcan, line 14: This doesn't look like it'll work for collections/nested field data

src/ReportManager/SchemaForm/SchemaFormContext/index.js
```diff
@@ -0,0 +1,52 @@
...
+      formData: {
+        [field]: value
+      }
+    });
+  };
+
+  const getFieldDetails = (fieldName) => {
+    const fieldSchema = schema.json.properties[fieldName];
```
JoshuaVulcan, line 22: This also doesn't look like it'll work for collections/nested field data

gaboDev: It wont, I'm just setting the grounds, I think any particular implementation/modification should be attended in its own ticket, for this case in the Collection ticket

src/ReportManager/SchemaForm/SchemaFormContext/index.js
gaboDev: Same as [here](https://github.com/PADAS/das-web-react/pull/1186#discussion_r1844336035) I did not wanted to have a wider scope for this ticket, once the foundations and responsibilities of the pattern are set can be extended in the particular ticket

src/ReportManager/DetailsSection/index.js
gaboDev: Done

src/ReportManager/DetailsSection/index.js
luixlive: What do you think of this @JoshuaVulcan ?

JoshuaVulcan: Thanks for the bump. We do indeed want to be able to create readonly event types, so we can model out incoming integration events.

src/ReportManager/SchemaForm/SchemaFormContext/index.js
luixlive: I feel like this is a start, but eventually we can add a util function just like we do in the EFB that translates the entire schema into a flat fields object that makes all the rendering details of each field easy to access, instead of having to call `getFieldDetails` each time, which will require some processing for nested fields and that probably will be called on each render. We can work on that later.

JoshuaVulcan (commented):
I like the approach overall, but a lot of the functions appear to take a single-dimension approach which won't map well to our deeply-nested data structures. @gaboDev I'm happy to provide some direct guidance on refactoring to support complex object structures if you'd like 🙂

luixlive (approved):
Good work Gabo. My mind keeps bringing possible improvements but we still have many incoming PRs for iterating this. This is a great start and sets good foundations to start rendering the new form 👍

PR #1192, opened 2024-11-19
ERA-10348: Support EFB schemas in ER > Text Field

gaboDev:
### What does this PR do?
- Adds support for Text types  
- Adds feature flag logic
- Adds Schema selector under settings pane

### How does it look
Form with a schema with short and text type fields
image

Schema selector
image

### Relevant link(s)
* Tracking tickets: [ERA-10348](https://allenai.atlassian.net/browse/ERA-10348)
* [Hosted demo environments](https://era-10348.dev.pamdas.org)

[ERA-10348]: https://allenai.atlassian.net/browse/ERA-10348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/constants/index.js
```diff
@@ -332,11 +332,13 @@ export const SUPPORTED_LANGUAGES = {
 export const FEATURE_FLAG_LABELS = {
   I18N_ENABLED: 'I18N_ENABLED',
   LEGACY_RT_ENABLED: 'LEGACY_RT_ENABLED',
+  EFB_FORM_SCHEMA_SUPPORT_ENABLED: 'EFB_FORM_SCHEMA_SUPPORT_ENABLED',
```
luixlive, line 335: We should have a modification in the file `.env` adding the ff:
```
# Feature flags
REACT_APP_LEGACY_RT_ENABLED=false
REACT_APP_I18N_ENABLED=true
EFB_FORM_SCHEMA_SUPPORT_ENABLED=false
```

Which reminds me... @JoshuaVulcan we should create a ticket to get rid of the `I18N_ENABLED` flag. Should be easy.

src/ReportManager/DetailsSection/index.js
```diff
@@ -67,7 +72,7 @@ const DetailsSection = ({
 }, ref) => {
   const dispatch = useDispatch();
   const { t } = useTranslation('reports', { keyPrefix: 'reportManager.detailsSection' });
-  const isNewFormSchema = false;
+  const isNewFormSchemaSupportEnabled = DEVELOPMENT_FEATURE_FLAGS[FEATURE_FLAG_LABELS.EFB_FORM_SCHEMA_SUPPORT_ENABLED];
```
luixlive: This should be a constant at the top of the file. We don't need to define a property in the scope of the component really 😄 
```
const EFB_FORM_SCHEMA_SUPPORT_ENABLED = DEVELOPMENT_FEATURE_FLAGS[FEATURE_FLAG_LABELS.EFB_FORM_SCHEMA_SUPPORT_ENABLED];
```

src/ReportManager/DetailsSection/index.js
```diff
@@ -216,7 +221,7 @@ const DetailsSection = ({
         : null}
     </div>
 
-    {!!formSchema && !isNewFormSchema && <Form
+    {!!formSchema && !isNewFormSchemaSupportEnabled && <Form
```
luixlive: Hmm this is weird. Remember that we need both implementations to work at the same time. What I read here is that depending on the feature flag we use the old or the new Form component. But what we want is that if the feature flag is off, we only use the old version. If the feature flag is on, we use both and we depend of the schema draft to decide which Form component to use.

src/ReportManager/DetailsSection/index.js
```diff
@@ -249,7 +254,7 @@ const DetailsSection = ({
           renderSubmitButton={renderSchemaFormSubmitButton} />
     }
 
-    {!formSchema && !reportForm.is_collection && loadingSchema && <ResizeSpinLoader
+    {!formSchema && !reportForm.is_collection && !isNewFormSchemaSupportEnabled && loadingSchema && <ResizeSpinLoader
```
luixlive: I guess even if we use the new version of the schema we will want a loader while we load it right?

src/ReportManager/SchemaForm/SchemaSelector/index.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import schemas from './mockedSchemas';
+
+import styles from './styles.module.scss';
+
+const SchemaSelector = ({ onChange }) => {
```
luixlive: Let's add a `TODO` somewhere remembering to delete this file eventually 👍

src/ReportManager/SchemaForm/index.js
```diff
@@ -44,10 +45,15 @@ const SchemaForm = ({ schema, onFormChange, formData, onFormSubmit, renderSubmit
...
+  const handleOnSchemaSelectorChange = (selectedSchema) => {
+    setSchema(selectedSchema);
+  };
+
   return <SchemaFormContextProvider schema={schema} formData={formData} onFormChange={onFormChange} formErrors={formErrors} >
+    <SchemaSelector onChange={handleOnSchemaSelectorChange} />
     <form onSubmit={handleOnSubmit} className={className}>
       {
-        schema.ui.order.map((sectionName) => (
+        schema && schema.ui.order.map((sectionName) => (
```
luixlive: nit: a more modern approach is to use conditional chaining

luixlive (commented):
I like the approach. Provided a few suggestions. However, the opinion that matters the most here is from QA since they will be the ones taking advantage of this. Let's provide them with a way to test this and see what they have to say.
cc: @AlanCalvillo @fonzy911

luixlive:
I just noticed I didn't before 😅 The field to select the schema is exactly over the view where QA should be testing the correct rendering of the form:
image

I don't think that's a good idea, we want them to be able to see how the full event view will show to the user. We should put that field somewhere else, maybe in the settings tab?

src/constants/index.js
gaboDev: You mean setting it to false by default? Didn't get it

src/ReportManager/DetailsSection/index.js
gaboDev: Got it 🫡

src/ReportManager/DetailsSection/index.js
gaboDev: I have a question regarding check if a schema comes from EFB or not, I remember that using a condition based on the draft property caused some discussions about if that was the best way to decide, also a suggestion was check which endpoint was used to fetch the data? not sure, Can you point me on the right direction? I'm a little bit lost here   

@luixlive @JoshuaVulcan

src/ReportManager/DetailsSection/index.js
gaboDev: I mean, isn't rendering huge and complex schemas a heavy computing task for the browser?? If yes, We should handle this somehow in the SchemaForm right? At least the current library of json-schema forms does it 

@JoshuaVulcan @luixlive

gaboDev:
> I just noticed I didn't before 😅 The field to select the schema is exactly over the view where QA should be testing the correct rendering of the form

> I don't think that's a good idea, we want them to be able to see how the full event view will show to the user. We should put that field somewhere else, maybe in the settings tab?

Sure I agree, but not sure what you refer by settings tab 😅

src/constants/index.js
luixlive: I mean adding it to the file, just like the other flags.

luixlive:
Before reviewing, I see some conflicting files 👀

.env
```diff
@@ -13,3 +13,4 @@ REACT_APP_DEFAULT_PATROL_FILTER_FROM_DAYS=1
 # Feature flags
 REACT_APP_LEGACY_RT_ENABLED=false
 REACT_APP_I18N_ENABLED=true
+REACT_APP_EFB_FORM_SCHEMA_SUPPORT_ENABLED=true
```
luixlive, line 16: This should be `false` by default. We don't want this to show up in prod envs. Developers must turn it on locally when working on this epic.

src/ReportManager/DetailsSection/SchemaForm/fields/Text/index.js
```diff
@@ -0,0 +1,81 @@
...
+const TEXT_INPUT_TYPE_TO_INPUT = {
+  [INPUT_TYPE.SHORT]: ShortTextInput,
+  [INPUT_TYPE.LONG]: LongTextInput,
+};
+
+const VALIDATION_ERROR_TYPES = {
+  REQUIRED: 'required'
+};
+
+export const TextFieldValidators = {
```
luixlive: Remember we won't be using these validators, we should remove this and any related code.

src/ReportManager/DetailsSection/SchemaForm/fields/Text/index.js
```diff
@@ -0,0 +1,81 @@
...
+  }
+};
+
+const Text = ({ fieldName }) => {
+  const textFieldDetails = useFieldDetails(fieldName);
+  const label = textFieldDetails.isRequired ? `${textFieldDetails.label} *` : textFieldDetails.label;
+  const TextInput = TEXT_INPUT_TYPE_TO_INPUT[textFieldDetails.inputType];
+  const { onFieldChange } = useContext(SchemaFormContext);
+
+  const stylesTheme = {
```
luixlive: This variable doesn't have any dependency on the content of the component. We should declare it as a constant outside of the component so it's not attached to its lifecycle and its not redefined in every render 👍

src/ducks/schema-selector.js
```diff
@@ -0,0 +1,24 @@
...
+  const { payload, type } = action;
+  if (type === SET_MOCKED_JSON_SCHEMA) {
+    return {
+      schema: payload
+    };
+  }
+  return state;
+};
+
+export default globallyResettableReducer(schemaSelectorReducer, INITIAL_SCHEMA_SELECTOR_STATE);
```
luixlive, line 26: I think it would be a good idea to persist this reducer with `redux-persist` so we don't have to choose the schema everytime we reload the page.

src/reducers/index.js
```diff
@@ -124,6 +125,7 @@ const rootReducer = combineReducers({
     userLocation: userLocationReducer,
     userLocationAccessGranted: userLocationAccessGrantedReducer,
     showReportHeatmap: reportHeatmapStateReducer,
+    schemaSelector: schemaSelectorReducer,
```
luixlive: A couple TODOs here and in the duck file would be good to remember to delete the reducer after we are done.

luixlive (changes requested):
This works great. I added just a couple suggestions and I see a couple differences between your components and the designs that I think should be very easy to fix. 
Figma:
image

Env:
image

1- There should be some space between the input and the description text.
2- Font size of the description should be 14px (in rems).
3- The long text input border should be a lighter gray.
4- Also in the long text input I think we should add some padding to the content in all directions (the text seems too close to the borders).

This should be ready once we fix these things ✅

luixlive:
@tiffanynwong you may want to test this!

.env
gaboDev: I left this as true to facilitate QA process, before this PR get merge I will update it

src/ReportManager/DetailsSection/SchemaForm/fields/Text/index.js
gaboDev: Sure

gaboDev:
Forgot to add a clarification for QA folks: This PR is only for rendering feature, form validation and submission are not part of this ticket 👀

luixlive (approved):
Looks great! Approved. Still I think it would be great to have thumbs up from @tiffanynwong if that doesn't block us from delivering in time.

A recommendation I have is to change the fixture you added for one that has more fields with more combinations (required / not required, default input, placeholder, long / short field, etc...). You don't need to worry about activated / hidden fields, that is handled by the method I'll add in the next PR `makeFieldsFromSchema` to remove the hidden fields there!

No required for this PR, but I don't think the persistence of the schema selector is working properly 🤔

fonzy911 (approved):

PR #1202, opened 2024-12-05
ERA-10519: [Events API Optimizations] Page Size change and Parallelization of calls

gaboDev:
### What does this PR do?
- It adds logic to handle parallel calls for pagination for map events fetch
- It adds modification to fetch more events on feed

### Relevant link(s)
* Tracking tickets: [ERA-10519](https://allenai.atlassian.net/browse/ERA-10519)
* [Hosted demo environments](https://era-10519.dev.pamdas.org)

[ERA-10519]: https://allenai.atlassian.net/browse/ERA-10519?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


luixlive:
The deployment failed. An Eslint issue it seems?

.env.development
```diff
@@ -1,4 +1,4 @@
-REACT_APP_DAS_HOST=https://root.dev.pamdas.org
+REACT_APP_DAS_HOST=https://ERA-10519.dev.pamdas.org
```
luixlive: Let's remove this change from the PR

src/SideBar/useReportsFeed/index.js
```diff
@@ -35,7 +35,7 @@ const useReportsFeed = () => {
       include_files: false,
       include_notes: false,
       include_updates: false,
-      page_size: 25,
+      page_size: 50,
```
luixlive: Maybe a good time to move this to a constant at the top of the file?

src/ducks/events.js
```diff
@@ -441,17 +442,15 @@ export const fetchMapEvents = (map, parameters) => async (dispatch, getState) =>
   });
 
   let resultsToDate = [];
-  const onEachRequest = onePageOfResults => {
+
+  const onEachRequest = ({ results: onePageOfResults }) => {
```
luixlive: We could follow a uniform language if we change the name of this method to `onPageFetch` 👍

src/ducks/events.test.js
```diff
@@ -16,7 +16,15 @@ describe('fetchMapEvents', () => {
   });
 
   test('appending a bbox parameter from the map object', async () => {
```
luixlive, line 18: Another good place to add an improvement. A long time ago we decided to stop mocking or spying requests. Instead now we intercept them and mock a server response with MSW so the client code runs intact and our test quality improves. We could add MSW handlers here and stop spying axios 😄

src/utils/parallelPaginatedRequest.test.js
```diff
@@ -0,0 +1,61 @@
...
+  test('handles all promises properly', async () => {
+    const itemsPerPage = 150;
+
+    const results = Array.from({
+      length: itemsPerPage
+    },
+    (_, i) => i
+    );
+
+    axios.get = jest.fn(() => {
```
luixlive: Same than before, but here I'd say is more important since these tests are good. Let's not mock axios, let's use MSW to intercept requests 👍

src/utils/parallelPaginatedRequest.test.js
```diff
@@ -0,0 +1,61 @@
+import axios from 'axios';
+
+import parallelPaginatedRequest from './parallelPaginatedRequest';
+import { EVENTS_API_URL } from '../ducks/events';
+
+describe('parallelPaginatedRequest', () => {
```
luixlive: I'd say we can add more testing scenarios here. Different items per page, different max retries, different concurrency, make sure that the callback is called with the right data, etc... I just see two very generic tests that cover most of the method, but we should dig in more scenarios.

src/utils/parallelPaginatedRequest.js
```diff
@@ -0,0 +1,73 @@
...
+      const response = await axios.get(`${apiUrl}&page=${page}&page_size=${pageSize}`, requestConfig);
+
+      if (response.status !== 200) {
+        throw new Error(`HTTP error ${response.status}`);
+      }
+
+      data = response.data.data;
+      break;
+    } catch (error) {
+      console.log(`Error fetching page ${page} (attempt ${attempt}):`, error.message);
```
luixlive: Maybe console.error?

src/utils/parallelPaginatedRequest.js
```diff
@@ -0,0 +1,73 @@
...
+    }
+  }
+
+  if (attempt >= maxRetries) { // just logging in error, this place can be used to trigger a callback on maxRetries exhausted per page
```
luixlive: Why ">=" and not "==="? Seems like this code makes it impossible to be greater than maxRetries right? Maybe a better way to rewrite this condition is `if (data === null)`

src/utils/parallelPaginatedRequest.js
```diff
@@ -0,0 +1,73 @@
...
+
+  if (!firstPageData) {
+    throw new Error('Failed to fetch the first page. Aborting.');
+  }
+
+  onPageFetch?.(firstPageData);
+
+  const totalItems = firstPageData.count;
+  let allItems = [...firstPageData.results];
+  const pagesToFetch = Array.from({ length: Math.ceil(totalItems / itemsPerPage) - 1 }, (_, i) => i + 2); // Exclude page 1
```
luixlive: This is hard to read and understand. Could we refactor this a little so its easier to understand what are we doing here and what `pagesToFetch` ends up being?

src/utils/parallelPaginatedRequest.js
```diff
@@ -0,0 +1,73 @@
...
+    }
+  };
+
+  for (const page of pagesToFetch) {
+    while (activeFetches.size >= concurrencyLimit) {
+      await Promise.race([...activeFetches]);
+      console.log('Wait for any fetch to complete.');
+    }
+
+    const fetchPromise = processPage(page)
```
luixlive, line 62: We are mixing async / await and .then .catch patterns. Could we stay with async / await everywhere?

src/utils/parallelPaginatedRequest.js
```diff
@@ -0,0 +1,73 @@
...
+      .then(() => console.log(`Page ${page} completed.`))
```
luixlive: I think this will add lots of noise in console. We should keep the console.logs to only the essentials (which is probably nothing since we have all this info in the network tab)

src/utils/parallelPaginatedRequest.js
```diff
@@ -0,0 +1,73 @@
...
+    const data = await fetchPage(apiUrl, requestConfig, page, itemsPerPage, maxRetries);
+    if (!!data) {
+      onPageFetch?.(data);
+      allItems = allItems.concat(data.results);
+    }
+  };
+
+  for (const page of pagesToFetch) {
+    while (activeFetches.size >= concurrencyLimit) {
+      await Promise.race([...activeFetches]);
```
luixlive: Hmm I don't understand why we are using Promise.race here 🤔 The nature of this feature is not a race at all. We not only care about the response of the fastest request. We want to call 5 promises in parallel and everytime one responds just add another one, like a pool. It seems like everytime the race finishes, we will iterate in the while  calling the Promise.race again with the 4 requests that didn't finish, causing doing the same requests several times (?) Could you explain a little the reasoning behind the race here please.

luixlive (changes requested):
Well done Gabo. This is an interesting pattern that we can then reuse in other places in the code to optimize our requests. I have many questions and suggestions since the code is kind of hard to understand in certain places. Maybe adding comments or refactorizing the code to making more readable would help.

src/utils/parallelPaginatedRequest.js
gaboDev: That's a typo, I thought I was fixed it, it should be an if statement

Regarding the Promise.race, it finishes when one of the promises of the set is processed, meaning it succeed or failed, so that result is taken and Promise.race if finished also, as far as I understand it only checks promises statuses so there is no request duplication

src/SideBar/useReportsFeed/index.js
gaboDev: Sure, good catch

src/utils/parallelPaginatedRequest.js
gaboDev: Sure, I split it into 2 const just to clarified each responsability

src/utils/parallelPaginatedRequest.js
gaboDev: I got your point, but in this case I'm working with the reference of the promise not with the promise result. 

Meaning that I want to set the `finally` callback for later when the promise is resolved/rejected in order to remove it
from the Set, and also need to add that promise to the pool, both of those steps need to be accomplished without waiting the promise to be solved (not blocking the code execution), the blocking process is only required when `Promise.race` enters in action.

src/utils/parallelPaginatedRequest.js
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  if (data === null) { // just logging in error, this place can be used to trigger a callback on maxRetries exhausted per page
+    console.error(`Failed to fetch page ${page} after ${maxRetries} attempts.`);
+  }
+
+  return data;
+}
+
+async function parallelPaginatedRequest(apiUrl, requestConfig = {}, { itemsPerPage = 150, maxRetries = 3, concurrencyLimit = 5, onPageFetch = null } = {}) {
```
luixlive, line 30: I'm ok having defaults I guess, but I see you added the defaults of the specific scenario you are working on (fetching map events). But this method is designed to be reused so maybe these are not the best generic defaults. I guess we shouldn't really have a default `itemsPerPage` since that will vary every time. The rest seems ok for me

luixlive (approved):
The Promise.race is still confusing for me, I can't really link how a race is related to what we are doing here (I understand how it works though, just wonder if there is a more straightforward way to implement it). But now with the env I can see it working and it does what it is expected.

However, I get the impression that when using this new fetching system the environment gets kind of slow. But I'm not sure if I'm doing a false relationship here and maybe my computer is slow because of something else. I'd leave further testing on that to the QA team @fonzy911 @AlanCalvillo

gaboDev:
> The Promise.race is still confusing for me, I can't really link how a race is related to what we are doing here (I understand how it works though, just wonder if there is a more straightforward way to implement it). But now with the env I can see it working and it does what it is expected.
> 
> However, I get the impression that when using this new fetching system the environment gets kind of slow. But I'm not sure if I'm doing a false relationship here and maybe my computer is slow because of something else. I'd leave further testing on that to the QA team @fonzy911 @AlanCalvillo

We can discuss the implementation in more detail, may be the pool can be handled only by a counter or something way more simpler ? not sure haha 

Also, I'm not sure if I'm getting wrongly the behavior/usage of [Promise.race](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/race) also it would be great to have more discussion on it


Regarding the slowness of the site, its quite a point, lets see QA results 👀

JoshuaVulcan (changes requested):
Nicely implemented 🙂  I did notice what feels like a UX problem; it seems like the map data may be getting cleared too early, before the newly-fetched data is retrieved. It's noticeable when zooming in towards a particular area, like this. Watch all the events disappear:

map-data-disappearing-parallelization

gaboDev:
> Nicely implemented 🙂 I did notice what feels like a UX problem; it seems like the map data may be getting cleared too early, before the newly-fetched data is retrieved. It's noticeable when zooming in towards a particular area, like this. Watch all the events disappear:




This change has been addressed, I sent the commit this morning but forgot to mentioned it here 🫡

JoshuaVulcan (approved):

fonzy911 (commented):
Hi @gaboDev, this is looking great so far. I'm able to see the adjusted items number for both feed and map, however, just wanted to run something by you real quick @netomo. 

When it comes to the map items, each batch of 150 within each page, is taking longer as they progress being fetched:

Screen Shot 2024-12-12 at 10 33 46

On root.dev, even though the map items count is lower (25) it is only taking this time to be fetched:

Screen Shot 2024-12-12 at 9 22 43

not sure if these are the numbers that we were expecting or if we should dive in a tad more.

AlanCalvillo:
> Hi @gaboDev, this is looking great so far. I'm able to see the adjusted items number for both feed and map, however, just wanted to run something by you real quick @netomo.
> 
> When it comes to the map items, each batch of 150 within each page, is taking longer as they progress being fetched:
> 
> Screen Shot 2024-12-12 at 10 33 46
> 
> On root.dev, even though the map items count is lower (25) it is only taking this time to be fetched:
> 
> Screen Shot 2024-12-12 at 9 22 43
> 
> not sure if these are the numbers that we were expecting or if we should dive in a tad more.

@fonzy911 this isn't an apples-to-apples comparison. root.dev is fetching way fewer events than feature-env, can you ensure you are testing similar amounts of data?

fonzy911:
> > Hi @gaboDev, this is looking great so far. I'm able to see the adjusted items number for both feed and map, however, just wanted to run something by you real quick @netomo.
> > When it comes to the map items, each batch of 150 within each page, is taking longer as they progress being fetched:
> > Screen Shot 2024-12-12 at 10 33 46
> > On root.dev, even though the map items count is lower (25) it is only taking this time to be fetched:
> > Screen Shot 2024-12-12 at 9 22 43
> > not sure if these are the numbers that we were expecting or if we should dive in a tad more.
> 
> @fonzy911 this isn't an apples-to-apples comparison. root.dev is fetching way fewer events than feature-env, can you ensure you are testing similar amounts of data?

Hi @AlanCalvillo, yes, thank you very much for that input. You are right, I have made the comparison now using my local instance, pointing to this backend: https://era-10519.dev.pamdas.org, and I see the following:

scoping this to map calls, I see that 6 calls of 25 map items take this amount of time to render:

image

Which is around 3.30seconds per 150 items

Now, with Gabo's fix, I see that 1 call of 150 map events, it takes around 2.24seconds to be completed

image

And the backend params are showing appropriately:

image

src/utils/parallelPaginatedRequest.js
```diff
@@ -0,0 +1,75 @@
+import axios from 'axios';
+
+async function fetchPage(apiUrl, requestConfig, page, pageSize, maxRetries) {
```
JoshuaVulcan, line 3: nit: we're pretty consistent throughout the repo with the `const fetchPage = async (...whatever)` convention. Why go `async function` instead?

JoshuaVulcan (approved):

fonzy911 (approved):

src/utils/parallelPaginatedRequest.js
```diff
@@ -0,0 +1,75 @@
...
+
+      data = response.data.data;
+      break;
+    } catch (error) {
+      console.error(`Error fetching page ${page} (attempt ${attempt}):`, error.message);
+    }
+  }
+
+  if (data === null) { // just logging in error, this place can be used to trigger a callback on maxRetries exhausted per page
+    console.error(`Failed to fetch page ${page} after ${maxRetries} attempts.`);
```
JoshuaVulcan: This is logging on fetch cancellations, which are intentional and happen quite frequently. As a result, the console is getting spammed in Chrome. Can we tighten that up to only log errors if they're not cancellations?

AlanCalvillo:
> > > Hi @gaboDev, this is looking great so far. I'm able to see the adjusted items number for both feed and map, however, just wanted to run something by you real quick @netomo.
> > > When it comes to the map items, each batch of 150 within each page, is taking longer as they progress being fetched:
> > > Screen Shot 2024-12-12 at 10 33 46
> > > On root.dev, even though the map items count is lower (25) it is only taking this time to be fetched:
> > > Screen Shot 2024-12-12 at 9 22 43
> > > not sure if these are the numbers that we were expecting or if we should dive in a tad more.
> > 
> > 
> > @fonzy911 this isn't an apples-to-apples comparison. root.dev is fetching way fewer events than feature-env, can you ensure you are testing similar amounts of data?
> 
> Hi @AlanCalvillo, yes, thank you very much for that input. You are right, I have made the comparison now using my local instance, pointing to this backend: https://era-10519.dev.pamdas.org, and I see the following:
> 
> scoping this to map calls, I see that 6 calls of 25 map items take this amount of time to render:
> 
> image
> 
> Which is around 3.30seconds per 150 items
> 
> Now, with Gabo's fix, I see that 1 call of 150 map events, it takes around 2.24seconds to be completed
> 
> image
> 
> And the backend params are showing appropriately:
> 
> image

@fonzy911 Did you gather metrics around how long the client takes to load X number of events, not only on a per-call basis but on the overall sum?
E.g 15 secs to render 3000 events vs 30 secs to render the same 3000 events. I don't think comparing individual calls is enough since they might vary. So an overall time will give us a glance at how would be the real UX. (See https://allenai.atlassian.net/wiki/spaces/ER/pages/30470471682/Results for inspo)
As discussed earlier, we don't only want to validate AC in a vacuum but actually make sure that the foundational objective is being met, which for this is story is "Optimize the loading of ER events to improve the performance of the web app. ")

fonzy911:
> https://allenai.atlassian.net/wiki/spaces/ER/pages/30470471682/Results

Hi @AlanCalvillo, yes, I was able to draft some numbers at a larger scale, this is what I got:

image

Without Parallelization, every call (for 25 items) takes in average 570ms to be completed

With Parallelization, it takes an average of 3.55 seconds for every 150 items call, making 591ms per every 25 items called.

AlanCalvillo:
> > https://allenai.atlassian.net/wiki/spaces/ER/pages/30470471682/Results
> 
> Hi @AlanCalvillo, yes, I was able to draft some numbers at a larger scale, this is what I got:
> 
> image
> Without Parallelization, every call (for 25 items) takes in average 570ms to be completed
> 
> With Parallelization, it takes an average of 3.55 seconds for every 150 items call, making 591ms per every 25 items called.

So if time w parallelization > w/o parallelization, to me, kinda suggests that the objective is not being met. why is it approved?

fonzy911:
> > > https://allenai.atlassian.net/wiki/spaces/ER/pages/30470471682/Results
> > 
> > 
> > Hi @AlanCalvillo, yes, I was able to draft some numbers at a larger scale, this is what I got:
> > image
> > Without Parallelization, every call (for 25 items) takes in average 570ms to be completed
> > With Parallelization, it takes an average of 3.55 seconds for every 150 items call, making 591ms per every 25 items called.
> 
> So if time w parallelization > w/o parallelization, to me, kinda suggests that the objective is not being met. why is it approved?

Previously, when running on an individual basis, the Parallelization process seemed to be faster than the original feature, that's where I originally approved it, but yes, it seems that performance diminishes as calls number rises. I have removed my approval due to new findings.

JoshuaVulcan (changes requested):
Requesting changes as per the recent QA dialogue

JoshuaVulcan (approved):

gaboDev:
> > > > https://allenai.atlassian.net/wiki/spaces/ER/pages/30470471682/Results
> > > 
> > > 
> > > Hi @AlanCalvillo, yes, I was able to draft some numbers at a larger scale, this is what I got:
> > > image
> > > Without Parallelization, every call (for 25 items) takes in average 570ms to be completed
> > > With Parallelization, it takes an average of 3.55 seconds for every 150 items call, making 591ms per every 25 items called.
> > 
> > 
> > So if time w parallelization > w/o parallelization, to me, kinda suggests that the objective is not being met. why is it approved?
> 
> Previously, when running on an individual basis, the Parallelization process seemed to be faster than the original feature, that's where I originally approved it, but yes, it seems that performance diminishes as calls number rises. I have removed my approval due to new findings.



After some meetings and [discussions](https://nimblegravity.slack.com/archives/C07D1LJSPBQ/p1736203650813199?thread_ts=1733783868.732719&cid=C07D1LJSPBQ) We agreed on changing the original QA strategy since it was not covering the whole process, it was changed to something more suitable in regards of the technical approach of this ticket

cc
@fonzy911

fonzy911 (approved):
Hi @gaboDev Gabo,

Thank you very much for implementing the new feature and addressing my previous comments so thoughtfully. I really appreciate the effort you’ve put into this.

To provide additional context, I conducted some tests to evaluate the performance of the new parallelization code compared to the previous implementation. Here's a quick summary of my approach:

I loaded a local instance of the app without the parallelization code and measured the time it took to render 1000, 2000, and 3000 items.
I repeated the same process using an instance with the new parallelization code applied.
I then compared the rendering times for both setups.
Initially, the results showed that the instance with the parallelization code took slightly more time to load all the items.

After discussing this with the development team, they clarified that the primary objective of this story was to implement the parallelization feature. They emphasized that summing individual API response times (e.g., A=2s, B=3s, C=4s) is no longer an accurate measure due to concurrent execution. Instead, testing should focus on improvements in overall UX, error handling, and HTTP call pool behavior.

When I reviewed this live using a waterfall view in the network tab, I observed that the map items were indeed rendered while calls were being executed concurrently within the same period, which supports the devs' explanation.

Screen Shot 2025-01-09 at 9 43 11

Based on these findings, I’m approving this PR. Testing it in a higher environment will give us a clearer picture of how it behaves under more extensive conditions.

Great work, and thanks again for your effort on this!

cc: @amicavi , @JoshuaVulcan

PR #1208, opened 2024-12-12
ERA-10520: Support EFB schemas in ER > Numeric field

gaboDev:
### What does this PR do?
- It adds EFB support for numeric fields

### How does it look
- image

### Relevant link(s)
* Tracking tickets: [ERA-10520](https://allenai.atlassian.net/browse/ERA-10520)
* [Hosted demo environments](https://era-10520.dev.pamdas.org)

- I added an mocked schema with several configurations for numeric fields

[ERA-10520]: https://allenai.atlassian.net/browse/ERA-10520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gaboDev:
Regarding the disadvantages of the native numeric input type, I have replaced it with a simple text with some tweaks which will improve the UX and the accessibility for the input, anyways, event if this text input is behaving as a numeric one it doesn't render the control of the counter (the two arrows for augmenting/reducing the input value) 

I'm more the happy to add that missing piece, just wanted to ask about your opinion if that would be a worthy effort 🫡 

cc
@luixlive @JoshuaVulcan

gaboDev:
> Regarding the disadvantages of the native numeric input type, I have replaced it with a simple text with some tweaks which will improve the UX and the accessibility for the input, anyways, event if this text input is behaving as a numeric one it doesn't render the control of the counter (the two arrows for augmenting/reducing the input value)
> 
> I'm more the happy to add that missing piece, just wanted to ask about your opinion if that would be a worthy effort 🫡
> 
> cc @luixlive @JoshuaVulcan


As discussed with the team, I will be creating a re-usable custom component for input numeric, this will give us the opportunity to use it not only in the EFB support but in the entire application when required

src/NumericInput/index.js
```diff
@@ -0,0 +1,111 @@
...
+
+  const handleOnKeyDown = (event) => {
+    const acceptedKeys = ['Backspace', 'ArrowRight', 'ArrowLeft', 'Shift', '.'];
+    const hasPointAlready = event.key === '.' && value.split('.').length > 1;
+    const notValidKey = !isNumber(event.key) && !acceptedKeys.includes(event.key);
+    const shouldBlockPropagation = notValidKey || hasPointAlready;
+
+    if (notValidKey){
+      if (event.key === 'ArrowUp'){
+        onUpArrowClick();
```
JoshuaVulcan: This needs a little love. These **are** valid keys, as shown by you doing something with input received by their presses. The naming is a little misleading, and the organization should be a little tidied up, especially because you then call a `click` event despite triggering from a key press.

src/NumericInput/styles.module.scss
```diff
@@ -0,0 +1,44 @@
...
+  width: 100%;
+
+  > input {
+    flex-grow: 2;
+    border-right: none;
+    margin-right: -2px;
+  }
+
+  .controls {
+    width: 25px;
```
JoshuaVulcan: Width, height, pretty much anything except border-width, we use `rem` values

JoshuaVulcan (changes requested):

luixlive (commented):
image
3 things before I start the code review:
1- It allows me to type number outside of the min/max range. I'd expect the new numeric component to not allow that.
2- The error state doesn't work. The second field is required but if I try to submit I don't see any error.
3- Those buttons are a little ugly 😅  It would only take a bit of CSS to make them look a little bit more friendly. Also they lack proper accessibility aria labels.
You can take as an example the Material field:
image
I don't think it's that nice either, but the buttons look better.

public/locales/en-US/reports.json
```diff
@@ -136,7 +136,8 @@
       "reportedByLabel": "Reported By",
       "schemaForm": {
         "errors": {
-          "required": "This is a required field."
+          "required": "This is a required field.",
+          "outOfRange": "Field value is out of range."
```
luixlive: I think it's a better user experience to actually provide the valid range as part of the error. You can send them as variables to the i18n `t` function. Like `Value out of range: must be below {{max}} and above {{min}}`. Of course there will be times where you only have one of the two constraints, but with some logic we can build a nice error string providing more info.

src/NumericInput/index.js
```diff
@@ -0,0 +1,111 @@
+import React, { useEffect, useState } from 'react';
+
+import { ReactComponent as ArrowUpSimpleIcon } from '../common/images/icons/arrow-up-simple.svg';
+import { ReactComponent as ArrowDownSimpleIcon } from '../common/images/icons/arrow-down-simple.svg';
+
+import styles from './styles.module.scss';
+
+
+const isNumber = key => !isNaN( parseInt(key) );
```
luixlive: This doesn't really control very much the string the user enters. It's a simple-functional approach, but we could do further validations. For example, if the user enter "1e5" this method would return true. Native numeric inputs support those kind of inputs, but do we want to support them? Here's an example of a more customized control of the input:
```
// utils.js
const isValidNumber = (number) => /^-?[0-9]+\.*[0-9]*$/.test(number);

// NumericInput.js
const handleOnChange = (event) => {
  // If the user enters a "." we autofill an initial "0"
  if (event.target.value === '.') {
    onChange('0.');
  }

  if (isValidNumber(event.target.value)) {
    // The input is valid, and this last step makes sure to remove things like trailing zeroes.
    const newValue = parseFloat(event.target.value).toString();
    onChange(newValue);
  }
};
```

This is just a example, I'm not sure if this code runs. And chat GPT helped me with the regex. But the point is that we could do further processing of the input to add railwards and help the user fill the input as it is expected.

src/NumericInput/index.js
```diff
@@ -0,0 +1,111 @@
+import React, { useEffect, useState } from 'react';
+
+import { ReactComponent as ArrowUpSimpleIcon } from '../common/images/icons/arrow-up-simple.svg';
+import { ReactComponent as ArrowDownSimpleIcon } from '../common/images/icons/arrow-down-simple.svg';
+
+import styles from './styles.module.scss';
+
+
```
luixlive: The following methods could be abstracted into a utils file to clean this one a little 👍

src/NumericInput/index.js
```diff
@@ -0,0 +1,111 @@
...
+const isNumber = key => !isNaN( parseInt(key) );
+
+const parseStringValueToNumber = (value) => {
+  return isFloat(value) ? parseFloat(value) : parseInt(value);
```
luixlive: Why not always use `parseFloat` since it also parses ints?
```
parseFloat('3'); // Output: 3
```

src/NumericInput/index.js
```diff
@@ -0,0 +1,111 @@
...
+
+  return newestValue.toFixed(precision);
+};
+
+
+const NumericInput = ({ id, value: formSchemaValue = null, setValue: setFormSchemaValue, min = null, max = null, ...otherProps }) => {
+
+  const [value, setValue] = useState(formSchemaValue === null ? '' : formSchemaValue.toString() );
+
+  const onUpArrowClick = () => setValue(augmentValue(value, min, max));
```
luixlive: Is there any value on defining these functions? We are changing a one-liner for another one-liner that is redefining itself in every render.

src/NumericInput/index.js
```diff
@@ -0,0 +1,111 @@
...
+
+  const onDownArrowClick = () => setValue(reduceValue(value, min));
+
+  const handleOnKeyDown = (event) => {
```
luixlive, line 61: We already discussed that this method should be refactored. Seems unnecessary to do so much processing on every key down, specially if we can control the onChange with a validation regex. Also remember that we should stop propagation since for many reasons we may be listening to key events in other places in the code. Here's an example of the suggestion I made of using a switch statement for controlling keys: https://github.com/PADAS/das-web-react/pull/1210/files#diff-58903a3355912365ad544cb70f4373124f2733968c0a0724dfc715fa0efe06edR107

src/NumericInput/index.js
```diff
@@ -0,0 +1,111 @@
...
+  };
+
+  useEffect(() => {
+    if (isNumber(value)){
+      setFormSchemaValue( parseStringValueToNumber(value) );
+    }
+  }, [setFormSchemaValue, value]);
+
+
+  return <div className={styles.numericInput}>
```
luixlive: We are missing accessibility tags here, specially for the buttons. Also, this numeric component that has the intention to be used for many purposes in our app should support more props. Like: required, disabled, readOnly, forwarding a ref... I see you spread the `otherProps` but that is not enough. If I want to disable the field also the buttons should be disabled. If it is readonly, I would expect the buttons not to render at all. Some of these functionalities may be out of the scope of the EFB schema, but implementing them takes a couple minutes and it makes these component way more complete.

src/NumericInput/index.test.js
```diff
@@ -0,0 +1,120 @@
...
+    const setValue = jest.fn();
+    renderNumericInput({ ...initialProps, value: '', setValue });
+
+    const numberInput = screen.getByRole('textbox');
+    await userEvent.type( numberInput, '1.5');
+
+    await expect(setValue).toHaveBeenCalledWith(1.5);
+  });
+
+  test('augment number using keyboard up arrow', () => {
```
luixlive: We are using a weird wording here: augment and decrement. It think the right word would be: increment and decrement.

src/NumericInput/index.test.js
```diff
@@ -0,0 +1,120 @@
...
+    renderNumericInput({ ...initialProps, value: '9', setValue });
+
+    const [upArrowButton] = screen.getAllByRole('button');
+
+    await userEvent.click(upArrowButton);
+
+    expect(setValue).toHaveBeenCalledWith(10);
+  });
+
+  test('decrease number using up button', async () => {
```
luixlive: This test name suggest that you can decrease the number by cliking the up button?

src/ReportManager/DetailsSection/SchemaForm/fields/Numeric/index.js
```diff
@@ -0,0 +1,52 @@
...
+
+      shouldAutofillDefaultInputRef.current = false;
+    }
+  }, [details.defaultInput, id, onFieldChange, value]);
+
+  return <div data-testid={`schema-form-numeric-field-${id}`} className={styles.numeric}>
+    <label className={`${styles.label} ${hasError ? styles.error : ''}`} htmlFor={id}>{label}</label>
+
+    <NumericInput
+        max={!details.maxInput ? null : details.maxInput}
```
luixlive: Joshua already pointed out that here we have a logic error if the schema includes a `maxInput = 0`, same for min.

src/ReportManager/DetailsSection/SchemaForm/fields/Numeric/index.js
```diff
@@ -0,0 +1,52 @@
...
+        min={!details.minInput ? null : details.minInput }
+        value={value === '' ? null : value}
```
luixlive: nit: why can't we send a empty string as the value? The null works too but an empty string shouldn't harm, and it's the most intuitive way to express an empty text input (even if we are masking it to appear like an only-numeric input)

src/ReportManager/DetailsSection/SchemaForm/fields/Numeric/index.js
```diff
@@ -0,0 +1,52 @@
...
+        id={id}
+        aria-describedby={hasDescription ? `${id}-description`: undefined}
+        aria-errormessage={hasError ? `${id}-description` : undefined}
+        aria-invalid={hasError}
+        aria-required={details.isRequired}
+        className={styles.numInput}
+        placeholder={details.placeholder}
+        setValue={(number) => {
```
luixlive: A standard way to call this method is `onChange`. Set value is not wrong or anything, but we could just keep the classic convention 👍

src/ReportManager/DetailsSection/SchemaForm/fields/Numeric/index.test.js
```diff
@@ -0,0 +1,149 @@
...
+    expect(onFieldChange).not.toHaveBeenCalled();
+  });
+
+  test('does not change the input value automatically if the autofill default input flag is off', () => {
+    renderNumericField();
+
+    expect(onFieldChange).not.toHaveBeenCalled();
+  });
+
+  test('shows a non required text field', () => {
```
luixlive: I think here you forgot to rename some of the copied tests jeje. In a few tests it says "text field" but it should be "numeric field"

src/ReportManager/DetailsSection/SchemaForm/utils/useSchemaValidations/index.js
```diff
@@ -16,6 +16,11 @@ const useSchemaValidations = (schema) => {
           return { ...accumulator, [error.params.missingProperty]: t('required') };
         }
 
+        if (error.keyword === 'minimum' || error.keyword === 'maximum' ) {
+          const [, fieldName] = error?.instancePath?.split('/');
```
luixlive: Is this going to work with collections? Because the instancePath will eventually mean nested paths. I don't remember very well why, but in my PR for date times I ended up extracting the name like this:
```
const fieldId = error.instancePath.split('/').pop();
```

luixlive (changes requested):
Lots of small suggestions 😄 It shouldn't take long and the improvements could be huge. Nice work so far! I'll keep this PR as my priority so let me know when I can do a second pass.

public/locales/en-US/reports.json
gaboDev: I have mixed feelings about this, this validation message will never be shown since currently the new custom numeric input is not allowing the user to type any number out of range 

That brings this question: the validation of the value is being done in real time, so if this validation found an error, should I show the error message when the value changed? instead of on submitting?

All fields are being validated until submit event, but this numeric one is doing those validations in real time (since it is a shared component)

So, should I avoid doing real time validation only for EFB forms and let ajv validation do the work allowing the user the type out-of-range numbers?

src/NumericInput/index.js
gaboDev: I have a question here, what is the expected behavior for the numeric input shared component when receiving the required prop as true?

public/locales/en-US/reports.json
gaboDev: Kind of the same applies for increasing/decrementing the value by using arrow keys, the validation is being done in real time

I think it would be good to keep both, validations in RT and on submit event

@luixlive @JoshuaVulcan

src/NumericInput/utils/index.js
```diff
@@ -0,0 +1,76 @@
+export const DECIMAL_POINT_SYMBOL = '.';
```
gaboDev, line 3: I'm already working on the coverage for this set of functions , it would be great to have a review while I finish this just to make some async progress

gaboDev: Also, most of this methods are taking for granted that `value` is a treated string which is coming from the onChange validations and filters, that is why some of them are not doing some extra checks

public/locales/en-US/reports.json
luixlive: I guess I don't understand why only one field should behave differently from the rest. Actually no controlled field should do its validations by itself. This is responsibility of the implementator. So, why is our shared numeric component doing its own validations? I understand that the component receives max and min and that it adds mechanisms to not allow invalid values, but that is not the same than running the validations 🤔

luixlive: Anyway, it feels like a different conversation to my original comment, that is providing more info to the user regarding the valid range of values.

src/NumericInput/index.js
luixlive: I would say just pass it down to the input, which you are already doing. Of course in the form builder implementation we don't want to use that flag cause it would trigger native validations. We can just use aria-required and add the * to the label.

src/ReportManager/DetailsSection/SchemaForm/fields/Numeric/index.test.js
luixlive: These tests still have incorrect tests @gaboDev

src/NumericInput/utils/index.js
luixlive: That's ok, it's responsibility of the implementator to pass a well formed value. We don't have to be that defensive since these components are for our own use.

public/locales/en-US/components.json
```diff
@@ -110,6 +110,11 @@
       "radioIconTitle": "Radio icon for {{subjectName}}"
     }
   },
+  "numericInput": {
+    "numericInputLabel": "Numeric field",
+    "incrementValueNumericInputButton": "Increment numeric value button",
```
luixlive: nit: it would be good to specify that this is a label `incrementValueNumericInputButtonLabel`

public/locales/es/reports.json
```diff
@@ -125,6 +125,7 @@
       "reportedByLabel": "Reportado por",
       "schemaForm": {
         "errors": {
+          "outOfRange": "El valor del campo está fuera de rango.",
```
luixlive: nit: this applies to all translations modified here, it would be nice to keep the translations sorted alphabetically for easy scanning.

src/ReportManager/DetailsSection/SchemaForm/fields/Numeric/index.test.js
gaboDev: Thanks, totally forgot about this

src/NumericInput/index.js
```diff
@@ -0,0 +1,147 @@
...
+    return value;
+  };
+
+  const handleOnChange = ({ currentTarget: { value } }) => {
+    const newValue = eraseNonValidChars(value);
+    const filteredValue = sanitizeExtraDecimalSymbols(newValue);
+
+    // ToDo: validate max and min
+    if ( min && parseFloat(filteredValue) < min ) {
+      // ToDo: define behavior
```
luixlive: I see missing logic here?

Remember that the validations are responsibility of the implementator. All you should do regarding max and min here is add the mechanisms to not allow values outside of the valid range.

src/NumericInput/index.js
```diff
@@ -0,0 +1,147 @@
...
+  readOnly = false,
+  min = null,
+  max = null,
+  inputClassName = '',
+  ...otherProps
+},
+ref) => {
+
+  const { t } = useTranslation('components', { keyPrefix: 'numericInput' });
+  const [value, setValue] = useState( `${formSchemaValue}` );
```
luixlive: This is supposed to be a controlled component, why are we handling the value in here? 🤔

src/NumericInput/index.js
```diff
@@ -0,0 +1,147 @@
...
+    if ( max && parseFloat(filteredValue) > max ) {
+      // ToDo: define behavior
+      console.log(filteredValue, ' is greater than max: ', max);
+      return;
+    }
+
+    setValue(filteredValue);
+  };
+
+  useEffect(() => {
```
luixlive: Why do we have a `useEffect` that triggers an onChange everytime the value changes? This will cause redundant calls to the `onChange` function!
Think of this:
1- The user types something
2- We trigger the onChange
3- The implementator receives the call and updates the value
4- Since the value changed we trigger this useEffect
5- We call onChange in here again

src/NumericInput/index.test.js
```diff
@@ -0,0 +1,137 @@
+import React from 'react';
+import userEvent from '@testing-library/user-event';
+import { fireEvent } from '@testing-library/react';
+
+import { render, screen } from '../test-utils';
+
+import NumericInput from './';
+
+describe('NumericInput', () => {
```
luixlive, line 9: I know you are finishing with the testing, but just a reminder to add tests for the new states like: readOnly, disabled, required, etc... You can find copy-able examples in my pickers 👍

src/NumericInput/index.js
```diff
@@ -0,0 +1,147 @@
...
+      onChange(
+        isNumber(value)
+          ? parseStringValueToNumber(value)
+          : null
+      );
+    }
+  }, [value]);
+
+
+  return <div className={styles.numericInput}>
```
luixlive: I think a good semantic practice is to add the role=group to this div to clarify that it is not only working as a division between elements of a webpage, but more like a single entity that groups several elements (a numeric input).

src/ReportManager/DetailsSection/SchemaForm/fields/Numeric/index.js
```diff
@@ -0,0 +1,54 @@
...
+    <label className={`${styles.label} ${hasError ? styles.error : ''}`} htmlFor={id}>
+      {label}
+    </label>
+
+    <NumericInput
+        max={details.maxInput ?? null}
+        min={details.minInput ?? null}
+        value={value}
+        id={id}
+        aria-describedby={hasDescription ? `${id}-description`: undefined}
```
luixlive: I see that the `otherProps` are being spreaded to the input and not to the parent div, but since the div really represents the numeric component grouping all its parts, I think some of these aria props should be in the div. It's a debatable, but that's what I've seen in some implementations. I guess you could just move the `...otherProps` from the input to the div since that is the root of our numeric component. But keeping functionality props like disabled, readonly, etc... In the input since those won't work in the div.

luixlive (commented):
Some nits. This seems pretty good. I was about to approve it but I saw something that really called my attention. Why are we handling the value of the numeric input as an inner state variable? We want it to be controlled. And I see you are doing some funky stuff with the onChange and the parent value to make it behave like a controlled component, but it's not really. Why can't we just receive the value, sanitize it before injecting it, and call the onChange so the parent updates it after doing the right processing of the input?

Two more things:
image
1- Small UX suggestion: It would be good to remove trailing dots `onBlur`. I did something similar in the time and date picker so you could take a look to the way I added logic to the blurring.

2- This is actually a bug: If you add a number with decimals, the increment and decrement buttons (or even the up and down arrow keys) stop working!

Final note: remember to add the translations to the i18n spreadsheet 😄

src/NumericInput/index.js
```diff
@@ -0,0 +1,135 @@
...
+  id,
+  inputAriaProps= {},
+  inputClassName = '',
+  max = '',
+  min = '',
+  onChange,
+  placeholder = '',
+  required = false,
+  readOnly = false,
+  value: formSchemaValue = '',
```
luixlive: This is a reusable global component, it's not going to be used exclusively by the form schema. This variable shouldn't have the name `formSchemaValue`. Actually, I'd recommend to leave it as `value` because that is the common naming pattern for the injected prop and instead rename the parsed one like `const stringifiedValue = ...` or something like that.

src/NumericInput/index.js
```diff
@@ -0,0 +1,135 @@
...
+  ...otherProps
+},
+ref) => {
+
+  const [decimalSymbol, setDecimalSymbol] = useState(null);
+  const [isNegative, setIsNegative] = useState(false);
```
luixlive: I understand that we need to keep track of what symbol does the user use for display purposes because the incoming value will always have a ".". But why do we need to keep track of if the value is negative or positive? That should come with the value everytime right?

src/NumericInput/index.js
```diff
@@ -0,0 +1,135 @@
...
+           onKeyDown={handleOnKeyDown}
+           onChange={handleOnChange}
+           onBlur={handleOnBlur}
+           value={value}
+           disabled={disabled}
+           readOnly={readOnly}
+           placeholder={placeholder}
+           required={required}
+           aria-label={t('numericInputLabel')}
+           {...inputAriaProps}
```
luixlive, line 131: Maybe I was a bit confusing in my comment regarding these aria props in my last revision, but what I meant was actually to keep them in the wrapper (the div). Since it has the role=group and its intention is to behave as the whole input including the actual <input> and the buttons, we can just inject the aria properties there. At least that's the pattern I had seen in other places with custom components like these.
Basically this would mean to get rid of this `inputAriaProps` and just inject them directly to the <NumericInput aria-...="" /> so they get spread in the div.

luixlive (approved):
Nice job! These components will be very useful in the future, and they are an improvement to our UI/UX. I don't have anything blocking and the patterns look great. I just left 3 last minor recommendations 👍

src/NumericInput/index.js
gaboDev: I need it to allow the user to type negative number directly and do not interfere with the parsing logic

src/NumericInput/index.js
gaboDev: I got mixed feelings, We do have aria props injected directly into other inputs, there are some tests and styles attached to this, I did not wanted to change that pattern

luixlive: That's ok. This should work well with assistive technologies. Let's just keep iterating and improving our custom implementations so they have full support for a11y.

AlanCalvillo:
the field behaves oddly when typing 0 in decimal values. for e.g: i can't type 1.09, but i can enter 109. if I paste the value 1.09 in the field being copied from elsewhere, it can be entered w/o issues.

src/SideBar/SettingsPane/SchemaSelector/mockedSchemas.js
```diff
@@ -1,3 +1,72 @@
+const schemaWithNumerics = {
+  label: 'Schema with numerics',
+  schema: {
+    'json': {
```
AlanCalvillo, line 4: can you add a field to this schema where a numeric field is inactive?

AlanCalvillo:
few questions for the team after discussing implementation with @gaboDev :

1. the current implementation throws a warning/error when the input is outside the given range in the schema until the user hits save. 
image

With this new implementation, users are not allowed to type numbers outside of the min/max range, which prevents them from reaching that warning message. I feel that we should be consistent between the two implementations, allowing users to type values and throwing the message upon saving changes.

2. scroll-down is active on previous implementation, for consistency i had the feeling that we should allow that here too, however, @gaboDev pointed out that there is a popular concern on that behavior, bc it might add undesired changes on input if they mistakenly scroll. i agreed w/ him, but just wanted to make sure everyone is ok w/ that.

3.- what should be the behavior when there are float/integer equivalences? if a user types 2.0, should we turn it into 2? or should we save the value as it was entered? my suggestion is that we keep the value as it was entered, and not turn it into its integer counterpart. 

cc: @luixlive @JoshuaVulcan @tiffanynwong

gaboDev:
> the field behaves oddly when typing 0 in decimal values. for e.g: i can't type 1.09, but i can enter 109. if I paste the value 1.09 in the field being copied from elsewhere, it can be entered w/o issues.

I have added some logic to handle this scenario, it was already there just needed some tweaks

@AlanCalvillo

gaboDev:
> few questions for the team after discussing implementation with @gaboDev :
> 
> 1. the current implementation throws a warning/error when the input is outside the given range in the schema until the user hits save.
> 
> With this new implementation, users are not allowed to type numbers outside of the min/max range, which prevents them from reaching that warning message. I feel that we should be consistent between the two implementations, allowing users to type values and throwing the message upon saving changes.
> 
> 2. scroll-down is active on previous implementation, for consistency i had the feeling that we should allow that here too, however, @gaboDev pointed out that there is a popular concern on that behavior, bc it might add undesired changes on input if they mistakenly scroll. i agreed w/ him, but just wanted to make sure everyone is ok w/ that.
> 
> 3.- what should be the behavior when there are float/integer equivalences? if a user types 2.0, should we turn it into 2? or should we save the value as it was entered? my suggestion is that we keep the value as it was entered, and not turn it into its integer counterpart.
> 
> cc: @luixlive @JoshuaVulcan @tiffanynwong



I've added a prop to the custom component that controls numeric range validation, so for the Numeric EFB implementation I have deactivated this feature, meaning a user can type numbers out of the range (also applies for the arrow buttons) letting the validation happens only at submit (also added a more meaningful validation error)

Regarding the point 3, I have added a piece of logic to keep the 2.0 or 5.0 or any X.0 number shown with the digits, anyway the data is being stored as plain integer number, this due the parsing mechanism of JS.

Meaning the user is allowed to use/type `X.0` numbers but once the event is saved the number will be stored as a plain integer.

@AlanCalvillo

luixlive:
Just providing my share:

> 1. the current implementation throws a warning/error when the input is outside the given range in the schema until the user hits save.

Yes, and with the new forms we don't want those native validations anymore. They are inconsistent in different browsers and disruptive with the flow of a HTML form. We are now handling our own error messages on submission and our fields are more restrictive so the users don't reach an error state (we try to enforce valid values while the user fills the inputs).

> 2. scroll-down is active on previous implementation [...]

Yes, and just as you mentioned, it's a concern around web technologies. Most libraries disable that functionality since it's pretty common that users scrolling a form modify the value in the input without noticing.

I feel like we shouldn't really try to stay with the old implementation if what we are doing enhancements to the fields. Adding our custom error messages, following best practices or adding mechanisms to help the user navigate or fill the form faster may make the experience a bit different and will be new to users, but should be better in the short term @AlanCalvillo .

AlanCalvillo:
> Yes, and with the new forms we don't want those native validations anymore. They are inconsistent in different browsers and disruptive with the flow of a HTML form. We are now handling our own error messages on submission and our fields are more restrictive so the users don't reach an error state (we try to enforce valid values while the user fills the inputs).

That context about inconsistency makes sense. I'm just doubtful about the possible confusion it might create for users who consume a form(but didn't create it themselves) and when trying to input a value outside the range, then the UI will not allow them to do so, and won't have any visual feedback. Hints/Description should add clarity, though.
I guess we can go ahead with the implementation of this PR, and try to gather feedback from users on how much confusion it creates if any.

tiffanynwong:
Love the inline errors 😍 

The way the input range is working now is great. @AlanCalvillo I'm seeing now that you can type in any numeric input and the field gets validated on save or if the user hits enter while the field is focused.

Some other feedback

1. Typing invalid inputs: I get that since it's a numeric field text or certain symbols aren't acceptable but we should have some feedback to the user about what inputs are accepted in the field. That way the user isn't angrily typing in text and not seeing a response from the ui. Could we do something like this, or is this out of scope?

- User types invalid inputs (letters, symbols)
- the field changes to error state
- the error message can say: Invalid input: Please enter numbers only.

2. Out of range error message: Just an update to the text here to be more accurate. "This value should be greater than or equal to 10"
Screen Shot 2025-01-22 at 10 05 42 AM
And "This value should be less than or equal to 20"
Screen Shot 2025-01-22 at 10 40 08 AM

3. > Meaning the user is allowed to use/type X.0 numbers but once the event is saved the number will be stored as a plain integer.

I agree w/ Alan here, we should keep whatever value the user types in. Since don't have a way for the user to specify formatting of the number inputed in the form builder we shouldn't assume that the input should be a whole number. It's very possible that the user will have to type in a value like 20.45 to say how much something cost for example.

luixlive:
@tiffanynwong 

> - User types invalid inputs (letters, symbols)
> - the field changes to error state
> - the error message can say: Invalid input: Please enter numbers only.

I'd say this is getting a bit out of scope 😅 We haven't done anything similar to other fields. All validations are done on submission. Also, I'm not sure I've seen a form anywhere else with that kind of validations. Even native HTML numeric input simply won't allow you to enter invalid characters: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number That's pretty standard. I feel like admins have some responsibility too to write good labels, descriptions and hints to help their users to understand the forms.

> It's very possible that the user will have to type in a value like 20.45 to say how much something cost for example.

I think Gabo mentioned that the only values that are stored as integers are if they have only zeroes after the dot, like `20.0 -> 20`. I don't think that is too bad,  but anyway it may make sense to keep the format users want for their values.

gaboDev:
> I agree w/ Alan here, we should keep whatever value the user types in. Since don't have a way for the user to specify formatting of the number inputed in the form builder we shouldn't assume that the input should be a whole number. It's very possible that the user will have to type in a value like 20.45 to say how much something cost for example.


As Luis mentioned, currently only the values which has only a zero as decimal digit `3.0`, `145.0` or `3.5 ` are stored as plain integer the reason is javascript data typing, but if the user enters a specific decimal number like 3.45 it will be stored as it is, a regular decimal number


and thanks for the feedback of the validation messages I will be addressing it 
@tiffanynwong

AlanCalvillo:
> Love the inline errors 😍
> 
> The way the input range is working now is great. @AlanCalvillo I'm seeing now that you can type in any numeric input and the field gets validated on save or if the user hits enter while the field is focused.

Yep, it got addressed after the comment 😄

AlanCalvillo:
> > the field behaves oddly when typing 0 in decimal values. for e.g: i can't type 1.09, but i can enter 109. if I paste the value 1.09 in the field being copied from elsewhere, it can be entered w/o issues.
> 
> I have added some logic to handle this scenario, it was already there just needed some tweaks
> 
> @AlanCalvillo

I think it's almost there. I'm just having issues when handling multiple zeroes:
- only one zero can be added. When I try to type in 15.00, the second zero deletes the first for some reason.
- also if I want to type, 15.908, the zero after the 9 can't be typed in. it can be pasted in, though.

https://github.com/user-attachments/assets/623f9071-1e10-46d9-b238-a7cab0e76f11

gaboDev:
> > > the field behaves oddly when typing 0 in decimal values. for e.g: i can't type 1.09, but i can enter 109. if I paste the value 1.09 in the field being copied from elsewhere, it can be entered w/o issues.
> > 
> > 
> > I have added some logic to handle this scenario, it was already there just needed some tweaks
> > @AlanCalvillo
> 
> I think it's almost there. I'm just having issues when handling multiple zeroes:
> 
> * only one zero can be added. When I try to type in 15.00, the second zero deletes the first for some reason.
> * also if I want to type, 15.908, the zero after the 9 can't be typed in. it can be pasted in, though.
> 
>  Screen.Recording.2025-01-22.at.17.06.07.mov

Good catch @AlanCalvillo 
I just sent a commit with the fixes 🫡

AlanCalvillo (approved):
let's friggin' gooo  💥 🚀 🚀 🚀

JoshuaVulcan (approved):

PR #1210, opened 2024-12-24
ERA-10486: Support EFB schemas in ER > Date and Time

luixlive:
# What does this PR do?
- Redoes our `DatePicker` and `TimePicker` components:
  - Use a custom input instead of a `<input type="time/date" />` for improved UI/UX and component controlling.
  Now we handle the input value as a string directly as the user write it so we have higher flexibility. The formats in which the input expects the value are `yyyy-MM-dd` and `HH:mm`, being `--` and `:` the expected value for empty inputs (there are constants for this).
  - Add proper dynamic accessibility labels.
  - Add internationalization: time 24 vs 12 + period formats and label wording (the only exception is that dates always stay in the same format `yyyy/MM/dd` as it was requested a few months ago).
  - Fix / add keyboard navigation.
  -  Improved validations and input restrictions.
  - Add missing functionality like supporting `required`, `readonly`, `disabled`, `onBlur`, `onFocus`, `min`, `max`.
- Adds the new `DateTimePicker` component:
  - It's a wrapper that groups together our `DatePicker` and `TimePicker` components, properly handling them as a single input and including all the features they have.
  - It also expects a string value with the format `yyyy-MM-ddTHH:mm` so the implementator has free handling of localization.
- Updates all the places where we used the previous versions of these components.
- Adds the `DateTime` schema form field using the new date and time pickers.
 
## How does it look
image

### Error state:
image

### TimePicker:
image

### DatePicker:
image

## Relevant link(s)
* [ERA-10486](https://allenai.atlassian.net/browse/ERA-10486)
* [Env](https://era-10486.dev.pamdas.org)

## Where / how to start reviewing (optional)
My suggestion is to ignore the diff of the `DatePicker`, `TimePicker` and `DateTimePicker` components since its a complete redoing and the differences in code just adds noise. I would recommend to pull the branch and check the three component implementations locally, they're pretty straightforward. There are comments explaining the parts that could be a bit confusing. 
Then you can go and take a look at the new `src/ReportManger/DetailsSection/SchemaForm/fields/DateTime` component and at the places where we had to update the implementations of the new components.

## Any background context you want to provide(if applicable)
Our old date / time components were barely functional. They didn't support lots of common input attributes, they didn't have the proper labeling and keyboard navigation, and the UI felt a bit improvised. They were corrected here so we have nice inputs in our new forms, but also the places where the old versions were used were updated.

[ERA-10486]: https://allenai.atlassian.net/browse/ERA-10486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -12,6 +12,7 @@
     "@turf/turf": "^7.1.0",
     "ajv": "^8.17.1",
     "ajv-draft-04": "^1.0.0",
+    "ajv-formats": "^3.0.1",
```
luixlive, line 15: This package is an extension for AJV to support validating string formats like `date`, `time` and `date-time`.

src/App.scss
```diff
@@ -265,10 +264,6 @@ aside {
   }
 }
 
-.react-datepicker {
```
luixlive, line 268: We no longer need to add app-level styling to the datepicker since we encapsulate the usage of that library into a single component that applies the styles.

src/DatePicker/CalendarPopper/MonthPicker/index.js
```diff
@@ -0,0 +1,220 @@
+import React, { forwardRef, useRef, useState } from 'react';
```
luixlive, line 1: The month picker is an inner component of our DatePicker to easily switch the month from the header of the calendar:
image

src/DatePicker/CalendarPopper/index.js
```diff
@@ -0,0 +1,154 @@
+import React, { forwardRef, useMemo, useRef } from 'react';
```
luixlive, line 1: The calendar popper is the menu that opens when you click the button at the end of the DatePicker to visually choose a date:
image

Here we use `react-datepicker` but without the input since we have our custom input. We just want the calendar and a button to open it.

src/DatePicker/CalendarPopper/index.js
```diff
@@ -0,0 +1,154 @@
...
+  maxDate={maxDate}
+  minDate={minDate}
+  nextMonthButtonDisabled={nextMonthButtonDisabled}
+  nextYearButtonDisabled={nextYearButtonDisabled}
+  onKeyDown={onKeyDown}
+  prevMonthButtonDisabled={prevMonthButtonDisabled}
+  prevYearButtonDisabled={prevYearButtonDisabled}
+/>;
+
+const CalendarPopper = ({
```
luixlive, line 83: There are comments explaining this, but I want to make emphasis because it's very important. I decided to handle the values of our custom components as plain strings. Just as if our inputs were text boxes. This is the most realistic and flexible way to handle the input of the user. They type or choose dates that end in the format `2020-01-01` so that's the way we control the input value.

However `react-datepicker` expects the value as a Date object. This means that we need to do transformations when interacting with the library components.

src/DatePicker/CalendarPopper/styles.module.scss
```diff
@@ -0,0 +1,104 @@
+@use '../../common/styles/inputs';
```
luixlive, line 1: Here are most of the style customization we do to the `react-datepicker`.

src/DatePicker/index.js
```diff
@@ -1,302 +1,395 @@
-import React, { createRef, forwardRef, memo, useCallback, useEffect, useMemo, useRef, useState } from 'react';
```
luixlive, line 1: I recommend to review these components directly in your editor instead of seeing the diff in Github. The diff is meaningless, they were basically re-done from scratch.

They main points I'd recommend to put attention to:
- The pattern: having several text inputs forcing numeric values (year, month and day) and sharing a single value.
- The UX: autofocusing inputs, keyboard navigation, prefilling values (like if the user writes a 3, there are no months that have two digits and start with a 3, so we prefill the first digit with a 0 and focus the day input now).
- The accessibility: aria tags, labels, HTML hierarchy, etc...

src/DatePicker/utils/index.js
```diff
@@ -0,0 +1,62 @@
+export const EMPTY_DATE_VALUE = '--';
```
luixlive, line 1: Utility functions abstracted so we don't repeat code and improve the readability with nice function naming.

src/DateRangeSelector/index.js
```diff
@@ -1,18 +1,22 @@
-import React, { memo, useCallback, useRef, useState } from 'react';
```
luixlive, line 1: The changes in this file are basically removing the old date picker  and use the new one. All the date range selectors of the app are now updated. This is an example with the filters:
image

The UI/UX improves considerably.

src/DateTimePicker/index.js
```diff
@@ -0,0 +1,109 @@
+import React, { forwardRef, memo, useImperativeHandle, useRef } from 'react';
```
luixlive, line 1: New component that groups the date and time pickers as a single input receiving a value in format `yyyy-MM-ddTHH:mm`.
image

Just as with the DatePicker, I'd recommend to check this in your local editor.

src/PatrolDetailView/PlanSection/index.js
```diff
@@ -1,6 +1,6 @@
 import React, { memo, useCallback, useEffect, useMemo, useState } from 'react';
```
luixlive, line 1: This file was also using the old DatePicker so I change the pattern to support the new one.

src/PatrolDetailView/index.js
```diff
@@ -185,7 +185,7 @@ const PatrolDetailView = () => {
 
     const endDate = displayEndTimeForPatrol(patrolForm);
     const startDate = displayStartTimeForPatrol(patrolForm);
-    if (endDate && startDate > endDate) {
+    if (!startDate || (endDate && startDate > endDate)) {
```
luixlive, line 188: Since the new date picker is more flexible for the user, it allows to have incomplete dates:
image

So we need new rail-wards to make sure the user don't try to submit a patrol without a valid date.

src/ReportManager/DetailsSection/SchemaForm/fields/DateTime/index.js
```diff
@@ -0,0 +1,128 @@
+import React, { memo, useEffect, useState } from 'react';
```
luixlive, line 1: This is the new schema form date-time component. It follows the same pattern than the Text component but using the new pickers.

src/ReportManager/DetailsSection/SchemaForm/fields/Text/index.js
```diff
@@ -8,19 +8,19 @@ const ShortTextInput = (props) => <input type="text" {...props} />;
 
 const LongTextInput = (props) => <textarea {...props} />;
 
-const FIELD_INPUTS = {
```
luixlive, line 11: Minor updates to keep the pattern consistent.

src/ReportManager/DetailsSection/SchemaForm/utils/useSchemaValidations/index.js
```diff
@@ -12,12 +14,25 @@ const useSchemaValidations = (schema) => {
   const runValidations = useCallback((formData) => {
     if (!validate(formData)) {
       const fieldErrors = validate.errors.reduce((accumulator, error) => {
+        if (error.keyword === 'format') {
```
luixlive, line 17: New AJV errors related to date-time fields.

src/ReportManager/DetailsSection/index.js
```diff
@@ -1,7 +1,7 @@
-import React, { forwardRef, memo, useCallback, useEffect, useMemo, useState } from 'react';
```
luixlive, line 1: Another place where we were using the old date picker and needed a refresh on the new pattern.

src/ReportManager/ReportDetailView/index.js
```diff
@@ -409,23 +409,11 @@ const ReportDetailView = ({
     reportTracker.track('Change Report Location');
   }, [reportForm, reportTracker]);
 
-  const onReportDateChange = useCallback((date) => {
```
luixlive, line 412: With the new components' pattern it doesn't make sense to treat date and time with separate handlers. We merge them into a single one.

src/SchemaFields/index.js
```diff
@@ -347,48 +347,28 @@ export const DateTimeWidget = ({
...
-  const handleDateChange = useCallback((newDate) => onChange(newDate ? newDate.toISOString() : newDate), [onChange]);
-
-  const handleTimeChange = useCallback((newTime) => {
-    const newTimeParts = newTime.split(':');
-    const updatedDateTime = date ? new Date(date) : new Date();
-    updatedDateTime.setHours(newTimeParts[0], newTimeParts[1], '00');
-
-    onChange(updatedDateTime.toISOString());
-  }, [date, onChange]);
+  let timeZoneCorrectedData = formData;
```
luixlive, line 350: Even the old schemas will benefit of the new components! We just need to handle the timezone correction ourselves now.

src/TimePicker/OptionsPopover/index.js
```diff
@@ -1,9 +1,9 @@
-import React, { forwardRef, useEffect, useMemo, useRef } from 'react';
```
luixlive, line 1: The last new component, the TimePicker. Same as with the DatePicker, I'd recommend to review and test the keyboard navigability, accessibility, pattern, UI...
image

luixlive:
### Important note for QA
@AlanCalvillo @fonzy911 this PR updates our old date and time pickers. This change affects every place where we use them like: report forms, report date, patrol start date and end date, date filtering, timeslider date configuration... It'd be nice to make sure that all them work properly.

Regarding report and patrol dates, it would be nice to check that they behave properly when opened in different timezones, since our app adjusts the time to the user's location.

src/DatePicker/index.js
```diff
@@ -1,302 +1,372 @@
...
+        onMonthChange(monthPlusOne);
+      }
+      break;
+
+    default:
+      break;
     }
+  };
+
+  const onDayChange = (newDay) => {
```
gaboDev, line 204: This method can be easily a one-liner

src/TimePicker/index.js
```diff
@@ -1,106 +1,482 @@
...
+      onTransformTo24HourAndChange(hour, minute, newPeriod);
+      break;
+
+    case 'ArrowRight':
+      event.preventDefault();
+
+      event.target.select();
+      break;
+
+    case internationalizedTimePeriods[AM_PERIOD][0].toLocaleLowerCase():
```
gaboDev: Can We declare this two vars at the beginning of the function an re-use them in all 4 cases?

src/DateTimePicker/index.js
```diff
@@ -0,0 +1,90 @@
...
+
+  useImperativeHandle(ref, () => innerRef.current);
+
+  // Value is expected to come as a string in format yyyy-MM-ddTHH:mm so we break it down to its parts.
+  const [dateValue, timeValue] = value.split('T');
+  const [hourValue = '', minuteValue = ''] = timeValue.split(':');
+
+  const [maxDate, maxTimeWithOffset] = max.split('T');
+  // We only apply a max time if there is a max date and the current value date matches it.
+  let maxTime;
```
gaboDev: Is there a reason to do not use useMemo or maybe an external function for maxTime and minTime calculation? instead of having all calculations floating there in the component, feels kind of messy

gaboDev (approved):
LGTM 🥳 

I just left some minor comments

This is fine work! 🎉

gaboDev:
@luixlive There are some conflicts with this branch 👀

fonzy911:
Hi @luixlive, everything looks great so far. 

Hey, just have a quick question. Whenever I pull the Data Range filter within events, there's a "now" button displayed, but, it seems that it's not functional:

image

Not sure if this is out of the scope of this story but, just wanted to run it by you, to be on the same page.

src/DateRangeSelector/index.js
JoshuaVulcan: The old layout had the controls side-by-side, which IMHO is more intuitive when setting boundaries on either side of a range:

Screenshot 2025-01-13 at 2 16 11 PM

Can we restore that layout?

gaboDev:
@luixlive Did you update the i18n spreadsheet with your changes? 👀

src/DateRangeSelector/index.js
luixlive: Well, the new UI that breaks the date and time in two fields has a wider width. In order to fit the new fields we need to expand the dialog considerably, making it look weird. If you'd rather have it that way I can do it, but honestly, I don't feel like this is "less" intuitive. Actually I think it is more consistent with our patrol view start and end dates. And we still have the labels "From" and "To". Let me know your thoughts, doing the change would be easy.

src/DateTimePicker/index.js
luixlive: useMemo feels unnecessary for such simple calculations, but I agree with you. The render method feels cluttered with all that code there. I'll create some methods outside of the component to handle that 😄

luixlive:
> Hi @luixlive, everything looks great so far.
> 
> Hey, just have a quick question. Whenever I pull the Data Range filter within events, there's a "now" button displayed, but, it seems that it's not functional [...]

Hey @fonzy911 ! That is not a button, it's just a legend to let the user know that when the `To` date-time field is empty, it basically means that the filters will apply until now. I didn't add it, I just had to move it there with the new UI. We used to have it inside the field box as you can see here:
image

If it's confusing, I could change the text to something more complete like: `Leaving the filter end date empty will apply filtering until now.` or something like that.

src/DatePicker/index.js
luixlive: You are right. I did the change, however i thought about it for a second, and we have several functions like this one: `onYearChange`, `onMonthChange` or even in the other pickers, like: `onHourChange` and `onMinuteChange`. And in all of them we follow the same pattern of defining a constant with the sanitized new input and then applying the change with a string template. I feel like we get more code beauty from keeping all those functions identical (consistent) than from reducing this one to a one-liner. What do you think?

fonzy911 (approved):

PR #1218, opened 2025-01-22
ERA-10346: Support EFB schemas in ER > Collections (One Level Deep)

luixlive:
### What does this PR do?
Add support for collection fields in the new form schema renderer.

### How does it look
Desktop:
image

Mobile:
image

### Relevant link(s)
* [ERA-10346](https://allenai.atlassian.net/browse/ERA-10346)
* [Env](https://era-10346.dev.pamdas.org/)

There are three important places to review carefully:
- `SchemaForm`: the main form component that is parent of all fields. Specifically, how the `formData` and `fieldErrors` objects, as well as the `renderField` method, are handled and passed down to sections, from there to fields and collections, and from collections to other inner fields and nested collections. 
- `Collection`: the new fields. They are broken in several fields, like: `Item`, `FormModal` and `FormPreview`. Collections are array of items, and each item has its own form. You can edit the form in the `FormModal` or see the current values using the expandable `FormPreview`.
- `useSchemaValidations`: the hook that runs AJV validations and builds an errors object. It needed a refactor to support nesting errors (fields within collections) and to provide a path so the user gets the visual cues to understand where are the errors located in the form.

There are a couple things I'd like to mention:
1- In the designs, the nested collections are navigable through the breadcrumbs and use the same view that just re-renders with the current collection.  This view is part of the form and when expanding it moves the fields around. This implementation was extremely complicated to achieve. Instead, we open modals with the collection forms. When you navigate into deeply nested collections you still have the breadcrumbs for orientation, but you can't really navigate using them since we render modals over modals.

2- I found a logical issue we didn't consider for collection constraints: the `minItems` validation is not applied if the collection is `undefined`. We were expecting that if a collection had, for example, `"minItems": 1`, AJV would return the min items error if the collection was not defined (empty). However, that is not the case. If `minItems` is set to 2 and the user only adds a 1 item, the error will be shown, but the validation doesn't run if the collection is undefined. For this to work, we need to do a combination of making the field required and having a minimum amount of items.

3- Figma designs expect collection items to be re-sortable by dragging and dropping them. That will be left for a separate ticket since that implementation can take some time and would complicate the review of this PR, which has already enough logic and visual elements involved.

[ERA-10346]: https://allenai.atlassian.net/browse/ERA-10346?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/DatePicker/CalendarPopper/index.js
```diff
@@ -142,6 +142,7 @@ const CalendarPopper = ({
     open={isOpen}
     popperClassName={styles.popper}
     popperPlacement="bottom"
+    popperProps={{ strategy: 'fixed' }}
```
luixlive, line 146: Needed to show the calendar popper correctly in modals (collections).

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/Item/FormModal/index.js
```diff
@@ -0,0 +1,93 @@
+import React from 'react';
+import Breadcrumb from 'react-bootstrap/Breadcrumb';
+import Modal from 'react-bootstrap/Modal';
+import { useTranslation } from 'react-i18next';
+
+import { ReactComponent as TrashCanIcon } from '../../../../../../../common/images/icons/trash-can.svg';
+
+import styles from './styles.module.scss';
+
+const FormModal = ({
```
luixlive, line 10: This is the modal that opens when a user wants to edit the content of a collection item form:
image

It doesn't do any calculations, all the data it needs is received. This way we don't need recursive functions to calculate the values or errors of its children, they are just passed down through the hierarchy of sections -> collections -> fields.

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/Item/FormPreview/index.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+
+import { getHumanizedValue } from '../utils';
+
+import styles from './styles.module.scss';
+
+const FormPreview = ({ errors, fieldIds, fields, formData }) => {
```
luixlive, line 8: This component is the form preview of each collection item. It's rendered as a collapsible by the `Item` component and it doesn't do any calculations either. Just as the `FormModal` the information is received and it just shows up humanized values and it renders an error state in the items that have errors: 
image

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/Item/index.js
```diff
@@ -0,0 +1,162 @@
...
+const getTitle = (formData, identifier, name, index, fields, language) => {
+  const defaultTitle = `${name} - ${index + 1}`;
+
+  if (!identifier || !formData[identifier]) {
+    return defaultTitle;
+  }
+  return getHumanizedValue(fields[identifier], formData[identifier], defaultTitle, language);
+};
+
+const Item = ({
```
luixlive, line 26: This component is responsible of rendering each collection item. It wraps the `FormPreview` in a collapsible component and controls the data and opening of the `FormModal`:
image

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/index.js
```diff
@@ -0,0 +1,167 @@
...
+
+import Item from './Item';
+
+import styles from './styles.module.scss';
+
+// Collections have an array of objects as their value in the form data object. Each of the objects is a collection
+// item and it contains the values of the fields rendered by a collection item. They can be nested within sections and
+// within other collections, so we propagate values, errors and breadcrumbs and their changes to the parent and the
+// children.
+const Collection = ({
```
luixlive, line 19: This is the main collection component. It's the parent of the array of `Item`s. It renders the label, collapsible items list, and the button. It has its own error state since collections may have constraints for the min and max amount of items:
image

Here we receive the values, errors and onChange callbacks from sections or other collections and pass them down to the children, this is the core of the collection nesting functionality.

src/ReportManager/DetailsSection/SchemaForm/fields/DateTime/index.js
```diff
@@ -120,7 +120,7 @@ const DateTime = ({ autofillDefaultInput: _autofillDefaultInput, details, error,
       className={`${styles.description} ${hasError ? styles.error : ''}`}
       id={`${id}-description`}
     >
-      {error || details.description}
+      {error?.message || details.description}
```
luixlive, line 123: The error data structure changed a little to support nesting collections so we needed to do some small changes in existing components.

src/ReportManager/DetailsSection/SchemaForm/fields/Section/index.js
```diff
@@ -2,27 +2,46 @@ import React, { memo } from 'react';
...
 import styles from './styles.module.scss';
 
-const Section = ({ details, id, renderField }) => <div
-    className={styles.section}
-    data-testid={`schema-form-section-${id}`}
-  >
-  {details.label && <h3 className={styles.header}>{details.label}</h3>}
+// Sections are just visual elements that are not present in the form data structure. Thus, fields contained by
+// sections are in the root objects of form data and field errors.
+const Section = ({ details, fieldErrors, formData, id, onFieldChange, onFieldErrorsChange, renderField }) => {
```
luixlive, line 7: The new pattern of cascading the field values, errors and change callbacks to support nested collections resulted in a small refactor in sections. Now they need to manually pass callbacks down to children and then report the changes to values and errors up to the `FormSchema`. This shouldn't result in any visual or logical difference, it's just a change in the pattern.

src/ReportManager/DetailsSection/SchemaForm/index.js
```diff
@@ -26,23 +26,17 @@ const SchemaForm = ({
...
+  // the form data, so we can keep the onSectionFieldChange dependency array empty.
   const shouldSendFormDataChangeRef = useRef(false);
 
   const [fieldErrors, setFieldErrors] = useState({});
   const [formData, setFormData] = useState(initialFormData);
 
   const fields = useMemo(() => makeFieldsFromSchema(schema), [schema]);
 
-  // TODO: Collections will require recusivity here.
-  const fieldValues = useMemo(() => Object.entries(formData).reduce((accumulator, [fieldId, fieldValue]) => ({
```
luixlive, line 39: We were planning to use this structure for a recursive approach, but instead we are cascading down the values so we don't need it anymore.

src/ReportManager/DetailsSection/SchemaForm/index.js
```diff
@@ -54,31 +48,47 @@ const SchemaForm = ({
...
-  const renderField = (fieldId) => {
-    const { type } = fields[fieldId];
-
-    const Field = FIELDS[type];
-
-    if (type === FORM_ELEMENT_TYPES.HEADER) {
-      return <Field details={fields[fieldId].details} id={fieldId} key={fieldId} />;
+  // This method is designed to render fields inside sections and collections. In order to support recursion we let the
+  // parents handle the propagation of values, change callbacks, errors, breadcrumbs (only for collections), etc...
+  const renderField = (id, value, onChange, error, breadcrumbs = []) => {
```
luixlive, line 64: The `renderField` method stays almost identical, but now it considers the propagation of data up and down in the hierarchy. So, in order to render children, we now need to pass down values, change callback, errors, and even the breadcrumbs (so they are built step by step while going down in the hierarchy of fields).

src/ReportManager/DetailsSection/SchemaForm/utils/useSchemaValidations/index.js
```diff
@@ -6,37 +6,91 @@ import { useTranslation } from 'react-i18next';
...
-          const fieldId = error.params.missingProperty;
-          return { ...accumulator, [fieldId]: t('required') };
+      // If the validation returned errors we iterate them.
+      return validate.errors.reduce((accumulator, error) => {
+        // First we calculate the error path, the field id and the message. The error path tells us if the erroneus
+        // field is nested in a collection and in which of its items.
+        let errorPath;
+        let fieldId;
+        let message;
+        switch (error.keyword) {
```
luixlive, line 47: We changed to a `switch` pattern since it made more sense than a bunch of `if`s and now we inject the error messages to an `errors` object that is built recursively and follows the collections / items paths so we can infer in which field of the hierarchy is the error:
```
// Example of an errors object where the date field in the first item of the court dates
// collection in the fourth item of the suspects collection was left empty but it is required
{
  suspects: {
    '3': {
      court_dates: {
        '0': {
          date: 'This is a required field'
        }
      }
    }
  }
}
```

JoshuaVulcan:
Initial usability feedback:

- There's a lot of extra space to the right of the caret indicating accordion state for each listed collection item. Makes the title feel cramped
- Without a trash can for collection items in the list, you have to individually drill down to each item's modal to delete it. This makes adding items too easy and removing items too difficult. Can we have a trash can next to the pencil icon to avoid dead ends like that?

tiffanynwong:
Looking so good!! 

1. **Flow for creating an item.** In the implementation the user can create an unlimited number of items before they have to edit the item. I think if we update the flow to force the user to fill out the item form first we can reduce some errors. So the flow can be:
- Click "+informant"
- the informant modal opens
- user adds inputs
- clicks done
- fields are validated and errors are shown if inputs are not valid
Screen Shot 2025-01-22 at 11 09 22 AM

2. The header on the modal should always be the item name. So in this example the header should be "Suspect" and typing in the name field shouldn't update the header. 
Screen Shot 2025-01-22 at 11 17 33 AM
That way the user will know what kind of item the are looking at. Only the breadcrumbs will update. 
Screen Shot 2025-01-22 at 11 19 36 AM

3. Is it possible to make the app look less like we're opening modals over modals 😅 ? Maybe if the modals that open after the first don't add a background, that could help. Maybe removing the slide in animation when opening a modal will help too. Is there a way to make the top modal open to the same size as the modal below it?
Screen Shot 2025-01-22 at 11 30 10 AM

4. Collapsed collection: Now that we've added the expand/ collapse button on the items, I think we should move that button to the right side of the collection label to match. [Here's an updated design.](https://www.figma.com/design/9oGOY9VuBYaHvOW3hjc88q/User-friendly-Admin?node-id=12871-34579&t=Ll8r1hpXh5Yj86sB-1) 
Screen Shot 2025-01-22 at 11 55 48 AM
Here's the example of the collection collapsed. 
Screen Shot 2025-01-22 at 12 35 40 PM


5. Is it possible to move the error message above the button? Like this?
Screen Shot 2025-01-22 at 11 56 42 AM

6. Items with two lines in the implementation aren't looking super great. I think reducing the line height will help. In this mock the text size is 16 and line height is 20. Screen Shot 2025-01-22 at 11 59 42 AM
Screen Shot 2025-01-22 at 12 13 55 PM

luixlive:
@JoshuaVulcan 
> Initial usability feedback:

Done ✅ 

@tiffanynwong 
> Flow for creating an item.

Now the modal opens immediately after adding an item, however I'm not running the validations when it is closed. We use AJV for validation schema so it would run all the validations of the form, not just the part of the modal. I feel like for this scope we can just run validations on submission but if we think it is worthy to run validations more often (even lively, with every change) we can discuss it separately.

> The header on the modal should always be the item name.

Done ✅ 

> Is it possible to make the app look less like we're opening modals over modals

Done ✅  Now it does't even look like it's a modal over another (but it still is). Just remember that navigation in the breadcrumb doesn't work because of modals. Also, I think it's better this way so it's clear for the user if the changes are done or cancelled.

> I think we should move that button to the right side of the collection label

Done ✅ 

> Is it possible to move the error message above the button?

Well, it is, but that means putting the error into the collapsed section. So if the user closes the collection they wouldn't be able to read the error. In my opinion it makes sense the description at the bottom of the field, like we do for the rest, and the add button is part of the field. If you still want the error to go above the button and you're ok with it being hidden with the collapsed elements, let me know.

> Items with two lines in the implementation aren't looking super great.

Agreed. However, adding fixes for multiline element causes other design pains and this could grow infinitely long. So instead I added ellipsis and if you hover the text you will get a native browser tooltip with the whole title 😄  Let me know if this doesn't work for any reason.

tiffanynwong:
> Flow for creating an item.
> 
> Now the modal opens immediately after adding an item, however I'm not running the validations when it is closed. We use AJV for validation schema so it would run all the validations of the form, not just the part of the modal. I feel like for this scope we can just run validations on submission but if we think it is worthy to run validations more often (even lively, with every change) we can discuss it separately.

Ok, got it. It's ok as it is since it's pretty easy to find the nested errors. 

> Is it possible to move the error message above the button?
> 
> Well, it is, but that means putting the error into the collapsed section. So if the user closes the collection they wouldn't be able to read the error. In my opinion it makes sense the description at the bottom of the field, like we do for the rest, and the add button is part of the field. If you still want the error to go above the button and you're ok with it being hidden with the collapsed elements, let me know.

Oh got it, yeah let's keep it where it is. 

> Items with two lines in the implementation aren't looking super great.
> 
> Agreed. However, adding fixes for multiline element causes other design pains and this could grow infinitely long. So instead I added ellipsis and if you hover the text you will get a native browser tooltip with the whole title 😄 Let me know if this doesn't work for any reason.

Is it easy to show all the text only when the item is expanded? If not, it's ok, the truncation works. 

Just a few more comments and otherwise it's looking really good!

1. Since the breadcrumbs don't work, can we change the color of the active item so it looks less like a link. Thinking a darker grey like the one used for the header "Witness."
Screen Shot 2025-01-23 at 4 19 55 PM

2. Using the same formatting to show total and which number item it is confusing. What if for total we do "Suspects (2)." And for the item, that doesn't have an identifier its "Suspect 1" and "Suspect 2"?
Screen Shot 2025-01-23 at 4 22 09 PM

luixlive:
@tiffanynwong 
> Is it easy to show all the text only when the item is expanded?
> [...]
> can we change the color of the active item so it looks less like a link.
> [...]
> What if for total we do "Suspects (2)." And for the item, that doesn't have an identifier its "Suspect 1" and "Suspect 2"?

All done ✅

tiffanynwong (approved):
Nice work! 🚀

AlanCalvillo:
Looking good so far, just a couple requests:

1. Can you add an inactive collection to the mocked schema, just to make sure that it's being handed correctly?

2.- When adding a new field to the collection, but canceling the process, the field persists. Is that expected?

https://github.com/user-attachments/assets/377c42fd-5e3a-4f7b-8d41-36d2ec752b7d

3.- Adding an error message for each field feels a bit clumsy 😬 
image

4.- Not sure if this is more part of a buggish for the text field, but on a mandatory field, a simple whitespace is accepted. 
image
image

JoshuaVulcan (approved):

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/index.js
```diff
@@ -0,0 +1,174 @@
...
+  value = [],
+}) => {
+  const { t } = useTranslation('reports', { keyPrefix: 'reportManager.detailsSection.schemaForm.fields.collection' });
+
+  const temporalIdentifierOfLastAddedItemRef = useRef(null);
+
+  const [isOpen, setIsOpen] = useState(true);
+  // React requires rendered arrays to have a unique key, but there's nothing we can use for the collection items so we
+  // handle an array of temporal identifiers with uuids for each item.
+  const [temporalIdentifiers, setTemporalIdentifiers] = useState(value.map(() => uuid()));
```
gaboDev, line 36: Can you elaborate a little bit more about having the tmpsIds on a local state?

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/Item/index.js
```diff
@@ -0,0 +1,170 @@
...
+  rightColumn,
+}) => {
+  const { i18n, t } = useTranslation('reports', {
+    keyPrefix: 'reportManager.detailsSection.schemaForm.fields.collection.item',
+  });
+
+  // We use these variables to store the initial errors and form data so we can restore those values if the user does
+  // changes and then clicks the cancel button.
+  const errorsBeforeEditingRef = useRef(null);
+  const formDataBeforeEditingRef = useRef(null);
```
gaboDev, line 49: So, having refs its your way of avoiding local state and re-renders coming from nested items, right?

tiffanynwong:
Ok, so feedback from Cassidy and Shan was about how the navigation between the nested collections is a little odd. 

1. Because the bottom buttons (delete, cancel, done) stay in the same position they look like a control that would save everything or cancel everything. Is it possible to have the "done" button say "Add [Item Name]"? 

2. When you press "cancel" the item still gets added to the collection. In this example pressing cancel should show you an empty suspect collection. 
Jan-24-2025 11-12-26

gaboDev (approved):
Nothing blocking this, just added a couple of questions

Great work! 🫡 🥳

luixlive:
@AlanCalvillo 
> Can you add an inactive collection to the mocked schema, just to make sure that it's being handed correctly?

✅ Done: You can see here I included a deprecated (hidden, inactive...) collection that doesn't show up https://github.com/PADAS/das-web-react/pull/1218/files#diff-9ed6d5efca4423ea2858ff006f6ddbc230c8dc0a1a88c17d89d0f2768f1c6cd0R637

> When adding a new field to the collection, but canceling the process, the field persists. Is that expected?

This and @tiffanynwong 's latest comment are about the same thing, the flow of the modal. We agreed to leave that out of this ticket since it wasn't defined before. We will have follow up tickets to improve that.

> 3.- Adding an error message for each field feels a bit clumsy 

Totally a bug, ✅ Done

> 4.- Not sure if this is more part of a buggish for the text field, but on a mandatory field, a simple whitespace is accepted.

Well, I wouldn't say it's a bug. Programmatically it makes sense, but as a human maybe it doesn't haha. It's not related to collections, this would be a change request for text fields, we can discuss this separately 😄

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/index.js
luixlive: Yes! This is just to satisfy the needing of providing `key`s to React components rendered as an array. When rendering the collection items we don't have a single thing that works as an identifier of them! The only thing that could be an id is the identifier value, which is the field's value we use to render the title, but they may be repeated, so it's not really unique, which causes rendering issues. So I had to create a temporal id which are uuids that will go away once the component unmounts.
Btw, I'm just starting with the drag and drop ticket for collection items, and this will go away since we will have now unique draggable identifiers that work better than these temporal ids.

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/Item/index.js
luixlive: Exactly, I could have handled these with state vars but that would get into the lifecycle of the component (causing unwanted renders). All we want is a copy of the initial data so if the user presses cancel we can un-do the changes. A ref is enough, and cleaner (no renders). 👍

src/SideBar/SettingsPane/SchemaSelector/mockedSchemas.js
```diff
@@ -586,155 +432,548 @@ const schemaWithDateTimes = {
...
+        },
+        estimated_number_of_people_involved_with_the_crime: {
+          default: '',
+          deprecated: false,
+          description: 'Number of people involved',
+          title: 'Estimated number of people involved with the crime',
+          type: 'string',
+        },
+        inactive_collection: {
+          deprecated: false,
```
AlanCalvillo: shouldn't this be set to `true` to make it really inactive?
image

luixlive: Duh... Yes haha. I forgot to deprecate it. And that's why it's showing up:
image

My fault, I'll fix it in a minute.

luixlive: Fixed 👍 @AlanCalvillo

AlanCalvillo (approved):

PR #1224, opened 2025-01-27
ERA-10488: Support EFB schemas in ER> Choice List with Dropdown input type selected

gaboDev:
### What does this PR do?
- Adding support for EFB choice list dropdown field

### How does it look
image

### Relevant link(s)
* Tracking tickets: [ERA-10488](https://allenai.atlassian.net/browse/ERA-10488)
* [Hosted demo environments](https://era-10488.dev.pamdas.org)

[ERA-10488]: https://allenai.atlassian.net/browse/ERA-10488?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


luixlive (changes requested):
A couple things I noticed before starting the code review:
1- Removing the line before the caret should be easy and it will match better the designs from Figma and the preview of the EFB:
image

2- The first field in your mock is required but if I try to submit without filling it I get no error state.

3- It would be great to add other fields in your mock, like numeric and text so we can compare the fields and make sure that the height of the inputs, colors, borders, etc... Match properly.

public/locales/en-US/components.json
```diff
@@ -19,6 +19,9 @@
     "partiallyCheckedTitle": "Partially visible",
     "uncheckedTitle": "Hidden"
   },
+  "choiceList": {
+    "noData": "No options to display"
```
luixlive: This key is not very descriptive. What is this text? A label, header, title, description... Actually. I just read the prop where we are passing it and I think the prop name would be a better key so the relation is explicit:
```
"choiceList": {
  "select": {
    "noOptionsMessage": "..."
  }
}
```

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/index.js
```diff
@@ -0,0 +1,97 @@
+import React from 'react';
+import { components as SelectComponent } from 'react-select';
+import { useTranslation } from 'react-i18next';
+
+import Select from '../../../../../Select';
```
luixlive: Two things:
1- Given that we are thoroughly configuring the `Select` and its components (like the options) why not to use the Select directly from `react-select` here? Our custom component doesn't seem to fit very well the new designs and configurations for the new forms.
2- Just as the Option component, we should customize the chips for the multiselect so they match better our designs. Should be simple 😄

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/index.js
```diff
@@ -0,0 +1,97 @@
...
+import { CHOICE_LIST_ELEMENT_INPUT_TYPES } from '../../constants';
+
+import styles from './styles.module.scss';
+
+const Option = ({ data, ...restProps }) => {
+  return <div>
```
luixlive: nit: No need for this `return`, and I see some weird indentation below.

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/index.js
```diff
@@ -0,0 +1,97 @@
...
+};
+
+
+const Dropdown = ({ details, onChange, value, id, hasError, disabled, ...otherProps }) => {
+
+  const { t } = useTranslation('components', { keyPrefix: 'choiceList' });
+  const getOptionLabel = (option) => option.title;
+  const getOptionValue = (option) => option.const;
+
+  return <div {...otherProps}>
```
luixlive: Why are we wrapping the select with a `div`? And more importantly, why are we spreading the rest of the props here? All our `aria-` labels should be spread directly to the input (select), and the docs specify that the component supports them: https://react-select.com/props

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/styles.module.scss
```diff
@@ -0,0 +1,55 @@
...
+.description {
+  color: colors.$alt-light-gray;
+  font-size: 0.875rem;
+  font-weight: 400;
+  margin: 0.125rem 0 0;
+}
+
+/* react-select overrides */
+.multiValue {
+  background: colors.$semi-light-gray !important;
```
luixlive, line 32: I see many `!important`s here. If we need the inevitably, that's ok, but I wonder if there's a better way to avoid some of them by using better selectors.

src/ReportManager/DetailsSection/SchemaForm/index.js
```diff
@@ -53,7 +55,7 @@ const SchemaForm = ({
     event.preventDefault();
 
     const fieldErrors = runValidations(formData);
-    if (fieldErrors) {
+    if (Object.values(fieldErrors).length > 0) {
```
luixlive: We need to undo this change! The logic behind `fieldErrors` is to return a `null` if there weren't errors. If `fieldErrors` is truthy (an object) it's because we hit at least one error. This makes me thing that we may be handling incorrectly a choice list error in the `runValidations` hook and we should fix that there, not patch this here. This can cause errors in other schemas that were already working properly.

src/ReportManager/DetailsSection/SchemaForm/utils/makeFieldsFromSchema/index.js
```diff
@@ -38,7 +38,10 @@ const addFieldToFieldsObjectRecursively = (
       uiSchema.fields[fieldId].allowableFileTypes;
   } else if (fields[fieldId].type === FORM_ELEMENT_TYPES.CHOICE_LIST) {
     fields[fieldId].details.inputType = uiSchema.fields[fieldId].inputType;
-    fields[fieldId].details.choices = uiSchema.fields[fieldId].choices;
+    fields[fieldId].details.choices = {
+      ...uiSchema.fields[fieldId].choices,
+      options: jsonSubschema.properties[fieldId].items.anyOf
```
luixlive: This is incorrect. And I think this originates because the mocked schemas may be incorrect too. Choice lists can be single choice or multiple choice, and depending on that the schema changes slightly: https://allenai.atlassian.net/wiki/spaces/ER/pages/30116741125/Form+Builder+Generated+JSON+Schemas#Choice-Lists-schema-format

Single choice:
image

If you see the schema, the value of a choice list property is of type string and it can be any of the values returned by a $ref. This means that form data should have something like: `{ singleChoiceListField: 'id-of-item-selected' }`.

Multi choice:
image

In here, the value of a choice list property is of type array and its items can be any of the values returned by the $ref. The form data should look like: `{ multiChoiceListField: ['id-of-item-selected-1', 'id-of-item-selected-2'] }`.

However, in this line you are implying that the schema will always have `items.anyOf` and that is not true! That is tru only for multi choice, but not for single choice. The logic needed here is more complex! I don't think we require the `...uiSchema.fields[fieldId].choices` anymore. That is necessary in the EFB to build the $ref, but in here you can get rid of it and only build the actual options array but the way you build it will depend on the schema.

src/SideBar/SettingsPane/SchemaSelector/mockedSchemas.js
```diff
@@ -734,7 +596,235 @@ const headerAndSections = {
...
+      'properties': {
+        'this_is_a_required_multi-choice_list': {
+          'deprecated': false,
+          'description': 'This is a really great description',
+          'title': 'This is a required multi-choice list',
+          'type': 'array',
+          'uniqueItems': true,
+          'items': {
+            'type': 'string',
+            'anyOf': [
```
luixlive: This doesn't seem like schemas will once DAS renders our refs. Right now we are building them this way:
```
"properties" {
  "choice_list": {
    "deprecated": boolean (field's active status),
    "description": string (field's helper text),
    "items": {
      "anyOf": [
        {
          "$ref": string (URL of the subschema)
        },
        ...
      ],
      "type": "string"
    },
    "title": string (field's label),
    "type": "array",
    "uniqueItems": true
  }
},
```
And the API (Ernesto's work) will resolve the $ref like this:
image

So the rendered schema should look like:
```
"properties" {
  "choice_list": {
    "deprecated": boolean (field's active status),
    "description": string (field's helper text),
    "items": {
      "anyOf": [
        {
          "$id": "http://tenant.pamdas.org/api/v1.0/featurecategories.json",
          "$schema": "https://json-schema.org/draft/2020-12/schema",
          "title": "FeatureCategories",
          "description": "A list of all feature categories available to the client",
          "type": "string",
          "oneOf": [
            {
              "const": "uuid1",
              "title": "Feature category 1"
            },
            {
              "const": "uuid2",
              "title": "Feature category 2"
            },
            ...
          ]
        },
        ...
      ],
      "type": "string"
    },
    "title": string (field's label),
    "type": "array",
    "uniqueItems": true
  }
},
```

luixlive (changes requested):
I provided a couple small suggestions regarding UI and code style, but also some important notes on things that are wrong regarding rendered refs, building the options and aria tags. The good thing of `react-select` is that it provides a very stable and complete select component, but we should configure it correctly and take a good look at the docs to make sure we enable accessibility and navigation correctly. 
Rendering choice lists as checkboxes and radio buttons will be extremely easier once we have this sorted, so I think we should take some time to have this well implemented and tested.

src/SideBar/SettingsPane/SchemaSelector/mockedSchemas.js
```diff
@@ -734,7 +596,235 @@ const headerAndSections = {
...
+                'title': 'Subject X'
+              }
+            ]
+          }
+        },
+        'this_is_a_non_required_single-choice_list': {
+          'deprecated': false,
+          'description': 'This is a description',
+          'title': 'This is a non required single-choice list',
+          items: {
```
luixlive: I forgot to send this comment in my review, but this is as important as the last one. Single choice schemas don't look like this! They are of type string, not array, so the property items doesn't exist in them. It makes sense if you thing the rationale behind. Multi choices are arrays of strings (they need to support multiple values) single choices are strings (because they only have a single value). You can take a look here: https://allenai.atlassian.net/wiki/spaces/ER/pages/30116741125/Form+Builder+Generated+JSON+Schemas#Choice-Lists-schema-format

And this circles back to my comment about how to build the options.

src/SideBar/SettingsPane/SchemaSelector/mockedSchemas.js
gaboDev: mmm this is something I talked with Ernesto as far as I understood the rendered part will be only an array with items of this form: 
```
{
   "const": "xxxxx-xxxxx-xxxx",
   "title": "Some label"
}
```

anyways, just tagged him to get a clarification 

cc 
@netomo

luixlive: Well, that wouldn't be properly replacing of the $ref. The screenshot I shared above comes from the actual PR where Ernesto added support for feature, subject and user schemas: https://github.com/PADAS/das/pull/3322

Refs should be replaced by the entire schema. But yeah let's hear Ernesto.

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/styles.module.scss
gaboDev: Yes, I can add online styles, but felt a little it messy

tiffanynwong:
Looking good, really like the multi select chips!

1. Agree w/ Luis that it would help a lot to compare the dropdown field to others that have already been made. 
Screen Shot 2025-01-27 at 9 40 21 AM Screen Shot 2025-01-27 at 9 41 02 AM
Please check that the font weight and text color on the label matches. Also looks like the focus state color is different too.

2. Bonus points, if it's easy to do can we add a check mark next to the choice that is selected in the single select list. Here's an example from the form builder. The multi-choice list works as is, since it looks like the choice is removed.
Screen Shot 2025-01-27 at 9 48 19 AM

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/index.js
JoshuaVulcan: Specifically, no need for the return **as long as you remove the curly braces from the function** 😅

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/index.js
```diff
@@ -0,0 +1,97 @@
...
+          data-testid={`schemaForm-field-choiceList-${id}`}
+          id={id}
+          onChange={(newValue) => {
+              onFieldChange(id, newValue);
+          }}
+          value={value}
+          details={details} />
+    </label>
+
+    {(hasDescription || hasError) && <p
```
JoshuaVulcan, line 109: Why not support cases where there's both a description **and** an error?

JoshuaVulcan (commented):
Test coverage looks solid -- code needs a little love but looks good overall! There's not a lot of technical feedback to give which @luixlive didn't already cover. Nice work

src/SideBar/SettingsPane/SchemaSelector/mockedSchemas.js
netomo: Interesting, @luixlive I was thinking the same way as @gaboDev, but if having the schema completely nested works then I like the way you're seeing it, working on the methods that will render the references, I'll test both ways of rendering references

AlanCalvillo: If you guys can get together to discuss this matter, can you please include me? i have a couple of questions on the scope of the mocks themselves.

luixlive: Sorry Alan, we did have a meeting to review this before you sent this comment. But I'm happy to help you with your questions.

luixlive:
@gaboDev let's merge develop changes (fix conflicts) and make sure to add some dropdowns inside collections. We need to make sure single and multi selects, errors, and every important aspect of the dropdown implementation, still work when nested in collections.

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/index.js
```diff
@@ -0,0 +1,113 @@
...
+import { ReactComponent as CheckIcon } from '../../../../../common/images/icons/check-light.svg';
+
+import { CHOICE_LIST_ELEMENT_INPUT_TYPES } from '../../constants';
+
+import styles from './styles.module.scss';
+
+const Option = ({ data, isSelected, isMulti, ...restProps }) => <div>
+  <SelectComponent.Option data={data} isMulti={isMulti} {...restProps}>
+    {
+      isSelected && !isMulti && <CheckIcon className={styles.checkMark} />
```
luixlive, line 15: Why `!isMulti`? I think we don't need that since multiselects don't even show the options right? And if they did, it would be nice to see the checkmark too.

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/index.js
```diff
@@ -0,0 +1,113 @@
...
+</div>;
+
+
+const Dropdown = ({ details, onChange, value, id, hasError, disabled, ...otherProps }) => {
+
+  const { t } = useTranslation('components', { keyPrefix: 'choiceList' });
+
+  const selectedValue = details.multiple && Array.isArray(value) ? value : details.options.find((item) => item.const === value);
+
+  const getOptionLabel = (option) => {
```
luixlive: nit: these two could be one-liners and even be written inline in the prop to clean a bit the component body.

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/index.js
```diff
@@ -0,0 +1,113 @@
...
+    return option.title ?? details.options.find((item) => item.const === option).title;
+  };
+
+  const getOptionValue = (option) => {
+    return option.const ?? option;
+  };
+
+  const handleOnChange = (newValue) => {
+    const isValueArray = Array.isArray(newValue);
+    if ( ( isValueArray && newValue.length === 0 ) || !newValue){
```
luixlive: I think it would be more idiomatic and the real logic would be more explicit if instead if checking if the value is an array, to check the `isMulti` flag in the details of the field. It's the same thing, but the logic would be easier to understand that the flow changes a little depending on if its multi or single select, instead of if the value is an array or a string. That applies in other parts of the code as well like the `selectedValue` constant where again I see that you check if its an array or not.

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/index.js
```diff
@@ -0,0 +1,113 @@
...
+      return onChange(undefined);
+    }
+    const returnedValue = isValueArray ? newValue.map((item) => item.const ?? item) : newValue.const;
+    return onChange( returnedValue );
+  };
+
+  return <ReactSelect
+      isClearable={true}
```
luixlive: nit: a small code styling suggestion, it's a common pattern in React to not add `={true}` and just set the prop like: `<ReactSelect isClearable isSearchable ... />`.

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/index.js
```diff
@@ -0,0 +1,113 @@
...
+      isMulti={details.multiple}
+      isSearchable={true}
+      onChange={handleOnChange}
+      options={details.options}
+      getOptionLabel={getOptionLabel}
+      getOptionValue={getOptionValue}
+      placeholder={details.hint}
+        components={{ Option }}
+        classNames={{
+          // control: () => ,
```
luixlive: I see some rare indentation in here, as well as a comment that should be removed.

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/index.js
```diff
@@ -0,0 +1,113 @@
...
+      {label}
+
+      <Input
+          aria-describedby={hasDescription ? `${id}-description`: ''}
+          aria-errormessage={hasError ? `${id}-description` : ''}
+          aria-invalid={hasError}
+          hasError={hasError}
+          aria-required={details.isRequired}
+          id={id}
+          onChange={(newValue) => {
```
luixlive: nit: possible on liner 🤠

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/Item/utils/index.js
```diff
@@ -18,9 +22,12 @@ export const getHumanizedValue = (field, value, defaultHumanizedValue, language,
...
   case FORM_ELEMENT_TYPES.CHOICE_LIST:
     if (field.details.multiple) {
-      return value.join(', ');
+      const humanizedValues = value.map((val) => {
+        return getChoiceListOptionLabel(val, field);
+      });
+      return humanizedValues.join(', ');
     }
-    return value;
+    return getChoiceListOptionLabel(value, field);
```
luixlive, line 30: We are replacing a one liner by a method that required 4 lines for its definition. We could just write the `.find` operation here.

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/Item/utils/index.test.js
```diff
@@ -3,6 +3,7 @@ import { format, parseISO } from 'date-fns';
 import { DATE_TIME_ELEMENT_INPUT_TYPES, FORM_ELEMENT_TYPES } from '../../../../constants';
 
 import { getHumanizedValue } from './';
+import { choicesOptions } from '../../../../../../../SideBar/SettingsPane/SchemaSelector/mockedSchemas';
```
luixlive: The `mockedSchemas` file will be deleted eventually. The intention is not to keep those. I think we need separate fixtures for these texts. If they are small we could even write them in here.

src/ReportManager/DetailsSection/SchemaForm/utils/makeFieldsFromSchema/index.js
```diff
@@ -13,6 +13,10 @@ const addHeaderToFieldsObject = (headerId, fields, uiSchema) => {
   };
 };
 
+const mapOptionsFromAnyOf = (ref) => {
+  return (ref.type === 'array' ? ref.items.anyOf : ref.anyOf).map(item => item.oneOf).flat();
```
luixlive, line 17: Seems good, but let's add some tests to make sure this function does what it should when we have multiple objects in the `anyOf`. Also, I'd suggest to rename the method to something like: `flattenChoiceListOptionsFromSchema`. Maybe a bit long that one haha, but at least is more explicit. We could think of something else but you get the point.

src/ReportManager/DetailsSection/SchemaForm/utils/makeFieldsFromSchema/index.test.js
```diff
@@ -8,6 +8,7 @@ import {
   ROOT_CANVAS_ID,
   TEXT_ELEMENT_INPUT_TYPES,
 } from '../../constants';
+import {choicesOptions} from "../../../../../SideBar/SettingsPane/SchemaSelector/mockedSchemas";
```
luixlive: Same than before, let's add some fixtures.

luixlive (changes requested):
I still see a few things that I think are worthy to work on before merging this one. But the difficult part I think was the schemas, references, flattening the options, etc... And that is very solid now 💯 

1- The border color (or outline) in the focus state doesn't match the one we've been using for other fields. Here a comparison between a dropdown and a text field:
image

2- The cursor doesn't have a hover state. I see three scenarios where I'd expect the cursor to tell me something but it doesn't:
  i. Hovering the input: since we are allowed to write in the input to filter options, I'd expect a `cursor: text;` there.
  ii. Hovering the caret: the caret has a hover state but the cursor stays as default. I'd expect a `cursor: pointer;`.
  iii. Hovering an option: the options are clickable and I'd expect some feedback from the cursor here too. Again `cursor: pointer;` when hovering an option.

3- There's an issue while using the widget with the keyboard. It's focusable, I can write in it easily, open it with space, navigate the options and chips with the arrows and remove the selection with backspace. However, I'd expect that if I open the options menu and then press escape, to simply close the menu again. Seems like react-select handles that correctly, however we need to stop the propagation of the event because the sidebar tries to close itself:
image

You can see this working correctly with the date and time pickers. When their menus are open, the keydown event of pressing escape stops its propagation to avoid these scenarios. It's not only the sidebar, we may also render these fields in modals that will try to close when pressing escape. 

I remember fixing this for the Priority and Reported By selects (and you can test them) so you should have examples around 😄 

4- I see you added the check mark as Tiffany suggested. I think it would be easy to make it match closer the EFB component as we have been trying to achieve with these new forms. Basically, a bit of a margin between the icon and the option text and changing the color of the icon:
image

EFB:
image

**Extra note:** remember to add the new translations to the spreadsheet!

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/index.js
luixlive: Bump to this comment @gaboDev ! I think it would be better to skip the custom implementation we use in other places in the code given that we are rewriting most of the configuration.

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/index.js
gaboDev: The same Option component is used to render both types of select, and adding a checkmark to multi-option select is not possible since selected options are deleted from the list

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/index.js
gaboDev: Sure 🫡

AlanCalvillo:
looking good, can you add another section to the schema with the following settings:
- two column, one column having single choice and the other multi.
- one inactive field

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/Item/utils/index.js
gaboDev: That one is used in 2 places, didn't want to duplicate logic

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/index.js
luixlive: Exactly, that's what I mean. Since they have been deleted from the list, then why do we need to add this validation? We shouldn't get an `Option` that has `isMulti` as `false` so feels unnecessary. And again, even if we were to render an option that has `isMulti` as `true`, why not to render the check icon? I know it won't happen, but this logic tells me that we don't want the check icon for those cases but I think we would.

AlanCalvillo:
i̶m̶ ̶a̶b̶o̶u̶t̶ ̶t̶o̶ ̶a̶p̶p̶r̶o̶v̶e̶ ̶i̶t̶,̶ ̶j̶u̶s̶t̶ ̶s̶e̶e̶i̶n̶g̶ ̶t̶h̶i̶s̶ ̶o̶d̶d̶ ̶b̶e̶h̶a̶v̶i̶o̶r̶ ̶:̶̶
 w̶h̶e̶n̶ ̶t̶r̶y̶i̶n̶g̶ ̶t̶o̶ ̶r̶e̶m̶o̶v̶e̶ ̶a̶ ̶v̶a̶l̶u̶e̶ ̶i̶n̶ ̶a̶ ̶m̶a̶n̶d̶a̶t̶o̶r̶y̶ ̶f̶i̶e̶l̶d̶,̶ ̶w̶h̶i̶c̶h̶ ̶w̶a̶s̶ ̶p̶r̶e̶v̶i̶o̶u̶s̶l̶y̶ ̶s̶a̶v̶e̶d̶,̶ ̶t̶h̶e̶n̶ ̶t̶r̶y̶i̶n̶g̶ ̶t̶o̶ ̶s̶a̶v̶e̶,̶ ̶t̶h̶e̶ ̶a̶j̶v̶ ̶c̶o̶m̶p̶l̶a̶i̶n̶s̶ c̶o̶r̶r̶e̶c̶t̶l̶y̶,̶ ̶b̶u̶t̶ ̶w̶h̶e̶n̶ ̶i̶ ̶t̶r̶y̶ ̶t̶o̶ ̶a̶d̶d̶ ̶a̶ ̶v̶a̶l̶u̶e̶ ̶b̶a̶c̶k̶ ̶i̶n̶ ̶a̶n̶d̶ ̶t̶h̶e̶n̶ ̶s̶a̶v̶e̶,̶ ̶t̶h̶e̶ ̶f̶o̶r̶m̶ ̶d̶o̶e̶s̶n̶'̶t̶ ̶s̶a̶v̶e̶.̶ ̶(̶s̶e̶e̶ ̶v̶i̶d̶e̶o̶ ̶i̶f̶ ̶w̶o̶r̶d̶i̶n̶g̶ ̶i̶s̶ ̶c̶o̶n̶f̶u̶s̶i̶n̶g̶ ̶)

̶h̶t̶t̶p̶s̶:̶/̶/̶g̶i̶t̶h̶u̶b̶.̶c̶o̶m̶/̶u̶s̶e̶r̶-̶a̶t̶t̶a̶c̶h̶m̶e̶n̶t̶s̶/̶a̶s̶s̶e̶t̶s̶/̶7̶e̶9̶d̶b̶5̶9̶a̶-̶d̶c̶e̶9̶-̶4̶3̶7̶7̶-̶a̶3̶b̶7̶-̶a̶7̶4̶4̶f̶c̶4̶5̶0̶7̶5̶6̶

̶n̶o̶t̶e̶ ̶t̶h̶a̶t̶ ̶t̶h̶i̶s̶ ̶d̶o̶e̶s̶n̶'̶t̶ ̶s̶e̶e̶m̶ ̶t̶o̶ ̶b̶e̶ ̶h̶a̶p̶p̶e̶n̶i̶n̶g̶ ̶i̶n̶ ̶o̶t̶h̶e̶r̶ ̶f̶o̶r̶m̶s̶.̶

seems like this was a false negative, a hypothesis is that the schema changed and fell into this [bug](https://allenai.atlassian.net/browse/ERA-11162)

AlanCalvillo (approved):

luixlive (approved):

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/index.js
luixlive: Just a bump on this @gaboDev I still don't get why it's necessary to have that. Feels like a bad logic. If multi selects show the options, we do want the checkmarks. If they don't show the options, then why do we need this flag at all?

luixlive:
@gaboDev 
I noticed three last super small things that would take you 2 minutes to fix!
image

1- The space between the label and the input doesn't match other fields, like you can see in the screenshot above, the numeric field has more space there (it's tiny, but noticeable).
2- In the single select, if you choose an option, it will appear in the input and you will have a cross icon at the right (like in the screenshot) to remove the selection. If you hover the cross, you don't get a pointer cursor. 
3- I don't see an example of an options with a longer text than the width of the menu. I would be good that you just make sure that it doesn't break the view. We could add ellipsis and a `title` attribute that shows the entire text with a native tooltip when the user hovers the option.

Still, those are minor, and this widget is very solid now. Good job!! ✅

JoshuaVulcan (approved):
Looking and feeling very nice 😎

gaboDev:
> @gaboDev I noticed three last super small things that would take you 2 minutes to fix! 
> 1- The space between the label and the input doesn't match other fields, like you can see in the screenshot above, the numeric field has more space there (it's tiny, but noticeable). 2- In the single select, if you choose an option, it will appear in the input and you will have a cross icon at the right (like in the screenshot) to remove the selection. If you hover the cross, you don't get a pointer cursor. 3- I don't see an example of an options with a longer text than the width of the menu. I would be good that you just make sure that it doesn't break the view. We could add ellipsis and a `title` attribute that shows the entire text with a native tooltip when the user hovers the option.
> 
> Still, those are minor, and this widget is very solid now. Good job!! ✅


Feedback already addressed, proceeding to merge this PR 🫡

PR #1226, opened 2025-02-04
ERA-11176: Patrol "to" date filter disabled after selecting "from" date filter

luixlive:
### What does this PR do?
Fixes a bug that was causing Patrol date filter to make impossible to set an end date.

### Relevant link(s)
- [ERA-11176](https://allenai.atlassian.net/browse/ERA-11176)
- [Env]()

The previous logic in the `max` prop of the end date `DateTimePicker` component was forcing to set a max date even if the max end date was `null`. The new logic checks if the max end date is `null` and in that case leaves the `max` as `undefined`, which is desired.

[ERA-11176]: https://allenai.atlassian.net/browse/ERA-11176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/DateRangeSelector/index.js
+++ b/src/DateRangeSelector/index.js
@@ -71,8 +71,6 @@ const DateRangeSelector = ({
     onFilterSettingsToggle && onFilterSettingsToggle(!filterSettingsOpen);
   }, [filterSettingsOpen, onFilterSettingsToggle]);
 
-  const hasEndMaxDate = typeof endMaxDate !== 'undefined';
-
   const onStartDateTimePickerChange = (dateTime) => {
     setStartDateTime(dateTime);
 
@@ -147,7 +145,7 @@ const DateRangeSelector = ({
                 popperPlacement: placement,
               },
             }}
-            max={formatDateToLocalISO(hasEndMaxDate ? endMaxDate : maxDate)}
+            max={endMaxDate === null ? undefined : formatDateToLocalISO(endMaxDate || maxDate)}
             min={formatDateToLocalISO(startDate || new Date('2000-01-01'))}
             onChange={onEndDateTimePickerChange}
             required={requireEnd}


src/DateRangeSelector/index.js
```diff
@@ -147,7 +145,7 @@ const DateRangeSelector = ({
                 popperPlacement: placement,
               },
             }}
-            max={formatDateToLocalISO(hasEndMaxDate ? endMaxDate : maxDate)}
+            max={endMaxDate === null ? undefined : formatDateToLocalISO(endMaxDate || maxDate)}
```
JoshuaVulcan, line 148: OK, just a minor bit of type checking funk. Oh JavaScript.

Can we use something like lodash's `isNil` to be more resilient against these little drifts?

JoshuaVulcan (dismissed):

fonzy911 (dismissed):
Hi @luixlive, it looks great now:

image

JoshuaVulcan (dismissed):

fonzy911 (approved):

src/DateRangeSelector/index.js
luixlive: Well, actually `isNil` does something we don't want here. It checks if the variable is `null` or `undefined`. If the prop `endMaxDate` was not passed (undefined) we want to set `maxDate`. Which is actually what this line does just at the end. If the parent explicitly passed down `endMaxDate` as `null` then we don't want to set a `max`, that is why we set the `undefined` here.
I know it's not the most intuitive approach, it seems it's been implemented like this for years. You can see the line I removed above was doing basically the same check:
```
const hasEndMaxDate = typeof endMaxDate !== 'undefined';
```

Meaning, if it's not undefined, it is defined, even if it is null. We just needed to do some changes with the new pickers. It feels out of the scope of this (and the pickers) ticket, but if you think it's worth the time I can think on something better. The downside is that this component is reused in a few places so it would involve more changes.

PR #1228, opened 2025-02-07
ERA-10523: Support EFB schemas in ER> Choice List with List input type selected

gaboDev:
### What does this PR do?
- Adds custom component called SelectListGroup for managing options listing
- It adds support for List option for Choice list type field 

### How does it look
- image

### Relevant link(s)
* Tracking tickets: [ERA-10523](https://allenai.atlassian.net/browse/ERA-10523)
* [Hosted demo environments](https://era-10523.dev.pamdas.org)

[ERA-10523]: https://allenai.atlassian.net/browse/ERA-10523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


luixlive (changes requested):
I have some initial impressions by testing the widget without reviewing the code:

1- Should we add a maximum height and make the container scrollable? The example lists seem to long and they increase the size of the whole form a lot. This is probably a question for @tiffanynwong 

2- Keyboard navigation is suboptimal. For example, the grouping div is focusable, but it is useless since there's no interaction the user can do with it. It just focuses because it has a `tabindex="0"` that shouldn't really be there. Only elements that have interactions should be focusable:
image

Also, for some weird reason, only the fist element of the list has a secondary focus state when you click tab that focuses the label only:
image

Taking a look at the [MDN web docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#try_it) I see that only the checkbox itself should be focusable and not the whole div or even the label (which I would expect that to be the default behavior when using native <input type="checkbox">):
image

I would expect our fields to follow those web standards, just with our custom border colors and other styles.

Finally, notice the differences in keyboard navigations in the MDN web docs for radios and checkboxes. We are not following that properly (we don't have the arrows navigation, for example). I think some of these behaviors will enable themselves after the following point (HTML semantics):
image

3- Reading the web standards, and with some help from DeepSeek to summarize other considerations, I see we are missing some key things related to the way we should implement a selectable list:
- The right HTML semantics are:
```
// Multi
<fieldset>
  <legend>Ranger</legend> // Widget-level label should be <legend> and not label
  <div>
    <input id="ranger-luis" name="ranger-luis" checked="false" type="checkbox" />
    <label for="ranger-luis">Luis</label>
  </div>
  <div>
    <input id="ranger-gabo" name="ranger-gabo" checked="true" type="checkbox" />
    <label for="ranger-gabo">Gabo</label>
  </div>
  <div>
    <input id="ranger-alfonso" name="ranger-alfonso" checked="false" type="checkbox" />
    <label for="ranger-alfonso">Alfonso</label>
  </div>
</fieldset>

// Single are idaentical but the inputs are type="radio", have the same name and have a value!
<input id="ranger-luis" name="ranger" checked="false" type="radio" value="luis" />
...
<input id="ranger-gabo" name="ranger" checked="true" type="radio" value="gabo" />
...
<input id="ranger-alfonso" name="ranger" checked="false" type="radio" value="alfonso" />
```
We are not using fieldset which is the standard and very important for assistive technologies to understand our forms. Also, we should be using a `legend` and not a `label` for the wrapper text. This is different form the rest of the components we have been creating. Of course, each input (checkbox or radio) should have its label.

[Here are the MDN web docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio) for the radio buttons. Notice that the attribute `name` works differently for the two types of inputs. If it's a single select (radio buttons) they all have the same name, so the form submission returns something like: `ranger: gabo` (name and value). While multi selects have their own name and no value since the value will be a boolean like: `ranger-luis: on, ranger-gabo: on`.

- Regarding accessibility, it seems we don't really need to be concerned about it if the semantics are implemented properly. Screen readers can detect fieldsets, their legend and associated inputs if they are properly implemented. However I see that you are passing down the aria attributes incorrectly to each input, instead of leaving the right ones at the `fieldset` level (which is not implemented yet). Others need specific customization depending on if multi or single select, like if the widget is required. Right now if the fields is required you are adding the `aria-required` atribute to every single checkbox input. That is like telling the user that they should check every single input in the list! That's definitely not what we want. Also you are setting `aria-describedby` in the input, but it should be in the fieldset.
image

I think this PR has the core of the implementation of the widget, but it has many semantic and a11y errors. The MDN docs are very straightforward and simple to read, and as I have suggested before, ChatGPT or DeepSeek are very helpful guides if something is not clear.

tiffanynwong:
Looking good!

> 1- Should we add a maximum height and make the container scrollable? The example lists seem to long and they increase the size of the whole form a lot. This is probably a question for @tiffanynwong

Good question. For one thing, we don't do that in the form builder preview rn and whatever we do in the rendering should match the form builder preview. Second, I got the impression that users would choose the list type because they want all choices to be seen in the form without having to click into a dropdown list. So I think it's ok without a max height, though I know it's different from what we do currently. 

However, we did get a comment from a user about showing the choices in 2 or 3 columns. That way the user could see all the choices and the list could take up less space in the form. For a one column choice list could we display the choices in two columns like this? This may add a bit of scope because we'd want to update the form builder preview to match, up to @JoshuaVulcan if it's worth doing. 
Screen Shot 2025-02-10 at 10 57 55 AM

Is it possible to get choices with longer text to show the full text instead of truncating? For a data collection app, it's not ideal to not show the entire text of a choice. The tool tip is nice but might not be discoverable enough and the scrolling doesn't quite work either. 
Screen Shot 2025-02-10 at 11 00 58 AM
Feb-10-2025 11-01-41
Like this:
Screen Shot 2025-02-10 at 11 04 47 AM

JoshuaVulcan:
@gaboDev @luixlive @tiffanynwong 
> > 1- Should we add a maximum height and make the container scrollable? The example lists seem to long and they increase the size of the whole form a lot. This is probably a question for @tiffanynwong

> Good question. For one thing, we don't do that in the form builder preview rn and whatever we do in the rendering should match the form builder preview. Second, I got the impression that users would choose the list type because they want all choices to be seen in the form without having to click into a dropdown list. So I think it's ok without a max height, though I know it's different from what we do currently.

I disagree with ^^this; without some bounds to the height, any site with more than 15 options in a choice list wouldn't be able to use this style of display without an IMHO unacceptable amount of vertical scrolling. I'd rather the choice list is max'd to some reasonable height (say, 10 items), and scrolls beyond that. The trick there, for scrolls-within-scrolls, is to ensure there's enough cursor space to empower users to scroll _past_ the choice list easily within the form. 

> However, we did get a comment from a user about showing the choices in 2 or 3 columns. That way the user could see all the choices and the list could take up less space in the form. For a one column choice list could we display the choices in two columns like this? This may add a bit of scope because we'd want to update the form builder preview to match, up to @JoshuaVulcan if it's worth doing.


@tiffanynwong I love the suggestion! I'll make a ticket. It can be post-MVP, let's focus on the current feature set for now. 

> Is it possible to get choices with longer text to show the full text instead of truncating? For a data collection app, it's not ideal to not show the entire text of a choice. The tool tip is nice but might not be discoverable enough and the scrolling doesn't quite work either.

100% agree with this

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/Dropdown.js
```diff
@@ -0,0 +1,79 @@
...
+
+  const handleOnChange = (newValue) => {
+    if ( ( details.multiple && newValue.length === 0 ) || !newValue){
+      return onChange(undefined);
+    }
+    const returnedValue = details.multiple ? newValue.map((item) => item.const ?? item) : newValue.const;
+    return onChange( returnedValue );
+  };
+
+  const onMenuClose = () => setMenuOpen(false);
```
luixlive: nit: These three functions could be inline and keep the component body a bit cleaner 😄  The last one can simply be: 
```
  onKeyDown={(event) => event.key === 'Escape' && isMenuOpen && event.stopPropagation()}
```

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/Dropdown.js
```diff
@@ -0,0 +1,79 @@
...
+  const [isMenuOpen, setMenuOpen] = useState(false);
+  const { t } = useTranslation('components', { keyPrefix: 'choiceList' });
+
+  const selectedValue = details.multiple ? value : details.options.find((item) => item.const === value);
+
+  const handleOnChange = (newValue) => {
+    if ( ( details.multiple && newValue.length === 0 ) || !newValue){
+      return onChange(undefined);
+    }
+    const returnedValue = details.multiple ? newValue.map((item) => item.const ?? item) : newValue.const;
```
luixlive: Why do we have the `??` operator here `item.const ?? item`. We should be sure of the format we use in our data right? Why do we expect sometimes to have the value in `item.const` and sometimes in just `item`?

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/List.js
```diff
@@ -0,0 +1,28 @@
...
+    }}
+    id={id}
+    disabled={disabled}
+    value={value}
+    options={details.options}
+    isMulti={details.multiple}
+    getOptionValue={(option) => option.const}
+    getOptionLabel={(option) => option.title}
+    hasError={hasError}
+    {...otherProps}
```
luixlive: We are using `otherProps` to spread extra properties that are not being handled here, but for some reason you are extracting some of the props like `id`, `value`, `hasError` and `disabled` and then just passing down. We could just levae those at the `otherProps` object and reduce the amount of lines here:
```
const List = ({ details, onChange, ...otherProps }) => <SelectListGroup
  onChange={(newValue) => onChange( !newValue || !newValue?.length ? undefined : newValue )}
  options={details.options}
  isMulti={details.multiple}
  getOptionValue={(option) => option.const}
  getOptionLabel={(option) => option.title}
  {...otherProps}
/>;
```

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/index.js
```diff
@@ -92,18 +21,19 @@ const ChoiceList = ({ details, error, id, onFieldChange, value = '' }) => {
 
   return <div>
     <label className={`${styles.dropdownWrapper} ${hasError ? styles.error : ''}`}>
```
luixlive: This is the unnecessary label that I mentioned. It should be conditionally rendered for dropdowns but not for the lists.

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/index.js
```diff
@@ -92,18 +21,19 @@ const ChoiceList = ({ details, error, id, onFieldChange, value = '' }) => {
...
-      {label}
-
+      { details.inputType === CHOICE_LIST_ELEMENT_INPUT_TYPES.DROPDOWN ?  label : '' }
       <Input
           aria-describedby={hasDescription ? `${id}-description`: ''}
-          aria-errormessage={hasError ? `${id}-description` : ''}
+          aria-errormessage={hasError ? `${id}-description` : undefined}
           aria-invalid={hasError}
           hasError={hasError}
-          aria-required={details.isRequired}
```
luixlive, line 102: We still need to make sure we are setting the required attribute. It just needs to be properly set for dropdowns vs checkboxes vs radios.

src/SelectListGroup/styles.module.scss
```diff
@@ -0,0 +1,27 @@
+@use "../common/styles/vars/colors";
```
luixlive: nit: let's use single quotes in these imports just like everywhere else

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/index.test.js
```diff
@@ -152,7 +152,7 @@ describe('ReportManager - DetailsSection - SchemaForm - fields - ChoiceList', ()
     expect(onFieldChange).toHaveBeenCalledWith('a-choice', '0d553bb7-5c4f-43d7-9b82-a561a668ae64');
   });
 
-  test.only('allow to select multiple options when the choice list is set to multiple selection', async () => {
+  test('allow to select multiple options when the choice list is set to multiple selection', async () => {
     const onFieldChange = jest.fn();
```
luixlive: We should add new tests here to cover the scenarios of schemas for list inputs. I know we have the tests for our new `SelectListGroup` in its own file, but we should test here that we render the field properly based on the schema attributes.

src/SelectListGroup/index.js
```diff
@@ -0,0 +1,126 @@
...
+
+  return optionValue === currentSelectListValue;
+};
+
+const SelectListGroup = ({
+  className = '',
+  disabled = false,
+  getOptionLabel = null,
+  getOptionValue = null,
+  hasError,
```
luixlive: nit: following the official (and our own) wording, I think it would be better to call this prop `invalid` to match the `aria-invalid`. We are not using aria here, which is ok since it's not stable for checkbox groups, but we could keep our language similar 😄

src/SelectListGroup/index.js
```diff
@@ -0,0 +1,126 @@
...
+  label,
+  options,
+  onChange,
+  readOnly,
+  value: selectListValue,
+  ...otherProps
+}) => {
+
+  const optionsState = useMemo(() => options.map((option) => {
+    const label = option.label ?? getOptionLabel?.(option);
```
luixlive, line 36: Just as before, this makes me wonder why our format is not stable. Why some times we expect the label to come from `option.label` and some times we fetch it from the method `getOptionLabel`? Our schemas and the data should follow a stable format.

src/SelectListGroup/index.js
```diff
@@ -0,0 +1,126 @@
...
+  getOptionLabel = null,
+  getOptionValue = null,
+  hasError,
+  id = '',
+  isMulti = true,
+  label,
+  options,
+  onChange,
+  readOnly,
+  value: selectListValue,
```
luixlive: Why do we rename this prop if this name is not being re-defined?

src/SelectListGroup/index.js
```diff
@@ -0,0 +1,126 @@
+import React, { useEffect, useMemo, useRef, useState } from 'react';
+
+import SelectableItem from './SelectableItem';
+
+import styles from './styles.module.scss';
+
+const calculateOptionValue = (option, getOptionValue = null) => option.value ?? getOptionValue?.(option);
+
+const isOptionChecked = (option, currentSelectListValue, isMulti, isListEmpty, getOptionValue = null) => {
```
luixlive: The variable `isListEmpty` makes me thing that there are no options available, not that there is no option checked. Maybe `isAnyOptionChecked` or something like that is a better name.

src/SelectListGroup/index.js
```diff
@@ -0,0 +1,126 @@
...
+    const value = calculateOptionValue(option, getOptionValue);
+    return {
+      isChecked: isOptionChecked(option, selectListValue, isMulti, !selectListValue || selectListValue?.length === 0, getOptionValue),
+      id: `${id}-${value}`,
+      label,
+      value,
+    };
+  }), [options, getOptionLabel, getOptionValue, selectListValue, isMulti, id]);
+
+  const [areOptionsFocused, setAreOptionsFocused] = useState(() => optionsState.map(({ id }) => ({
```
luixlive: Hmm can we have several options focused at the same time? I don't think so right? Then why do we store an array of ids instead of simply storing the id of the focused item or null if none is focused? This way we don't have to iterate an array at all, we just need to hold the value of the focused option or null optimizing storage and processing.

src/SelectListGroup/index.js
```diff
@@ -0,0 +1,126 @@
...
+
+    if (focusedOptionIndex !== autoFocusedOptionIndex){
+      setAutoFocusOptionId(areOptionsFocused[autoFocusedOptionIndex].id);
+    }
+  };
+
+  useEffect(() => {
+    if (autoFocusOptionId){
+      setTimeout(() => {
+        autoFocusRef?.current?.focus();
```
luixlive: I'm really lost trying to understand this `autoFocus` stuff. `autofocus` is a native HTML property to focus an element in the screen automatically when the screen is rendered. Is this related to that? If not, maybe this name is confusing. But still, could you explain the logic of this focus? Why do we focus in an `useEffect` with a delay of 100 hardcoded milliseconds?

src/SelectListGroup/index.test.js
```diff
@@ -0,0 +1,115 @@
+import React from 'react';
+import userEvent from '@testing-library/user-event';
+
+import { render, screen } from '../test-utils';
+
+import SelectListGroup from './';
+
+describe('SelectListGroup', () => {
```
luixlive, line 8: There are many states and props that these tests are not considering: read only, disabled, hasError, etc, etc... We should make sure all of them work properly.

src/SelectListGroup/SelectableItem/index.js
```diff
@@ -0,0 +1,87 @@
...
+  isFocused,
+  label,
+  onClick,
+  readOnly = false,
+  setIsFocused,
+  value,
+  isMulti = true
+}, ref) => {
+
+  const role = isMulti ? 'checkbox' : 'radio';
```
luixlive: Let's not hardcode strings and have constants at the top of the file.

src/SelectListGroup/SelectableItem/index.js
```diff
@@ -0,0 +1,87 @@
+import React, { forwardRef } from 'react';
+
+import styles from './styles.module.scss';
+
+const Ripple = ({ onClick, readOnly, disabled, children }) =>
```
luixlive: Very nice touch! The ripple seems great. However, the click callback doesn't work well. If you click the div of the ripple it will show a pointer cursor but select the first checkbox of the group always.

src/SelectListGroup/SelectableItem/index.js
```diff
@@ -0,0 +1,87 @@
...
+  setIsFocused,
+  value,
+  isMulti = true
+}, ref) => {
+
+  const role = isMulti ? 'checkbox' : 'radio';
+
+  const handleOnChange = (event) => {
+    event?.preventDefault();
+    readOnly || disabled || onClick(value, !isChecked);
```
luixlive: This syntax is not very readable. Using an `if (!readOnly && !disabled)` is more obvious and still simple. However, if we are using a native input with `readOnly` or `disabled` the events should be automatically blocked by the browser and this condition shouldn't be necessary!

src/SelectListGroup/SelectableItem/index.js
```diff
@@ -0,0 +1,87 @@
...
+             ref={ref} />
+    </Ripple>
+
+    <label htmlFor={id}>
+      {label}
+    </label>
+  </div>;
+};
+
+const SelectableItemWithRef = forwardRef(SelectableItem);
```
luixlive: nit: unnecessary extra constant. We can simply `export default forwardRef(SelectableItem);`

luixlive (changes requested):
Nice work Gabo! This is way closer to the final implementation now.

I still see some issues with the design and semantics that are causing some issues with the UI and UX.
1- In Figma, the `legend` is outside of the border:
image

 But in your environment, the legend is inside:
image

Fixing that is very easy, you just need to add the border to a div that wraps just the inputs instead of adding it to the entire `fieldset`. Also, the border radius doesn't match the designs.

2- You are wrapping the `fieldset` with a `label` that is semantically incorrect. We should remove it. Each checkbox input has its label, and the fieldset has its legend, that's all we need.
image

3- If you click anywhere in the list that is not part of a checkbox (for example in the small area in between of two inputs) the first checkbox gets checked / unchecked. I have the theory that this is related to that unnecessary label wrapping everything propagating a click event to the first input it encounters.

4- The inputs are a bit hard to click. You can see the pointer cursor even if you are not exactlly hovering the input (you can hover a bit over the square of the checkbox) however if you click it won't check that option. It will select the first one because of the error mentioned in point number 3. And something similar happens with the label because it has an unnecessary top-margin:
image

Instead of using that margin we should align our elements using something like  flexbox. And we should also fix the cursor pointer to make clear when we are hovering a clickable option or not.

5- If a group of radio inputs is required, we should add the `required` attribute to one of the inputs: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio#additional_attributes
image

It's a good practice to add the required attribute to the first one. We should add that to our widget.

6- The `value` attribute in checkboxes is not really necessary since the default form behavior is to se them as "on". All we need is the id in the `name` attribute to identify the field. We can just get rid of those values.

src/SelectListGroup/index.js
gaboDev: This is the custom shared component, here We are letting the implementator use any chosen format to pass label and value, which is the same pattern react-select allows us to us any data format to render the options, so this is not related with your previous comment about consistency (Which I'm still checking, since it was part of the prev PR 😅 )

src/SelectListGroup/index.js
gaboDev: Just forgot to remove the setTimeOut, I was getting an exception due a logic error, in regards of the autofocus logic, it is used to focus the next/prev input based on the keyboard navigation

luixlive: Ok, if it's the functionality to move the focus based on the keyboard, then I guess it's not `auto` right?

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/index.js
gaboDev: Adding a required attr to the inputs was triggering html native form validations, so I left only the aria prop

luixlive:
Hey @gaboDev I cannot test the widget in the hosted environment. Seems like the deployment has the feature flag off 😞

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/Dropdown.js
```diff
@@ -0,0 +1,75 @@
...
+
+  return <ReactSelect
+        components={{ Option }}
+        classNames={{
+            clearIndicator: () => styles.clearIndicator,
+            container: () => styles.container,
+            control: (state) => `${styles.control} ${ invalid ? styles.dropdownError : '' } ${ state.isFocused ? styles.controlFocused : '' }`,
+            dropdownIndicator: () => styles.cursorPointer,
+            indicatorsContainer: () => invalid && styles.caretError,
+            indicatorSeparator: () => styles.separator,
```
luixlive: nit: this line is adding a style with `display: none !important;`, which means that the separator is still rendered and then hidden with CSS. It would be better to not render it at all by:
```
const CustomIndicatorSeparator = () => null;
...
return <ReactSelect
        components={{ Option, IndicatorSeparator: CustomIndicatorSeparator }}
```

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/Dropdown.js
```diff
@@ -0,0 +1,75 @@
...
+            multiValue: () => styles.multiValue,
+            multiValueRemove: () => styles.multiValueRemove,
+            option: () => styles.cursorPointer,
+            placeholder: () => invalid && styles.error,
+        }}
+        isClearable
+        isDisabled={disabled}
+        inputId={id}
+        isMulti={details.multiple}
+        isSearchable
```
luixlive: nit: `isSearchable` is `true` by default in `react-select` so this prop is not necessary.

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/Dropdown.js
```diff
@@ -0,0 +1,75 @@
...
+
+import styles from './styles.module.scss';
+
+const Option = ({ data, isSelected, isMulti, ...restProps }) => <div>
+  <SelectComponent.Option data={data} isMulti={isMulti} {...restProps}>
+    {
+        isSelected && !isMulti && <CheckIcon className={styles.checkMark} />
+    }
+    <span className={`${styles.optionLabel} ${ !isMulti && !isSelected && styles.singleOption }`}
+          title={data.label}>
```
luixlive: nit: Putting the title in this `span` makes the area where the tooltip appears extremely small. It would be better to add it to the wrapper `div` so the entire option component shows the tooltip:
```
const Option = ({ data, isSelected, isMulti, ...restProps }) => <div title={data.label}>
  <SelectComponent ...>
```

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/index.js
```diff
@@ -91,20 +20,26 @@ const ChoiceList = ({ details, error, id, onFieldChange, value = '' }) => {
   const label = details.isRequired ? `${details.label} *` : details.label;
 
   return <div>
-    <label className={`${styles.dropdownWrapper} ${hasError ? styles.error : ''}`}>
-      {label}
+    {
+      details.inputType === CHOICE_LIST_ELEMENT_INPUT_TYPES.DROPDOWN &&
+        <label className={`${styles.dropdownWrapper} ${hasError ? styles.error : ''}`}>
```
luixlive: Now you are not wrapping the `Input` with the `label` anymore but they are not connected 😬 The DOM doesn't recognize the connection between the label and the input, you can see that because when you click the label the input is not focused. You need to add the prop `htmlFor` in the label and provide the `inputId` you passed to the `react-select` (which is simply the field `id` prop you have in here too).

src/SelectListGroup/SelectableItem/styles.module.scss
```diff
@@ -0,0 +1,131 @@
...
+      background: transparent;
+      transition: background-color 200ms ease-in-out;
+
+      position: absolute;
+      z-index: 1;
+      top: 0;
+      left: 0;
+      margin-left: -0.55rem;
+      margin-top: -0.4rem;
+
```
luixlive, line 56: nit: I see lots of empty lines with no sense around in this file.

luixlive (commented):
Nice job @gaboDev ! I don't want to block this PR any further, I added a couple very low priority nits. The only thing I consider important is the label not being connected to the input since that affects navigation and accessibility.

Now, I finally understood the logic I was asking about last time, this `autoFocus` thing you had and all the keyboard navigation stuff you added. It's not necessary! Browsers handle that navigation by default, that's what I tried to explain in my first pass. You can see that working here:
- Checkbox: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#try_it
- Radio: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio#try_it

See how simple the code really is:
image

The navigation you coded by yourself is not needed and makes this component error-prone. We should leave this to the native functionality. I just tried by myself by pulling this branch and removing all that code and it works just great. Basically you can remove all these props from `SelectableItem`: setIsFocused, isFocused, ref, focusNextSelectableItem, focusPreviousSelectableItem. ANd get rid of the `handleOnKeyDown`, `onFocus`, `onBlur` callbacks. It took me a couple minutes to remove all that code and test the lists working just like they do natively in those MDN docs, which are the standards!

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/index.js
luixlive: Yes, but I see that you are deleting it here and didn't see it in the environment 🤔

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/index.js
gaboDev: Not sure why I missed that 😆

luixlive (approved):
You got it! Semantics look great, it's accessible and easy to navigate 😄 

I non-blocking found an issue that should be very simple to address, which is that now that you are not handling the navigation (which is good) you are not calling `event.preventDefault()` when the user presses enter while focusing a checkbox. This is causing the form to be submitted but if a checkbox is focused I would expect that to be prevented and just to check the box. It should be as easy as adding an `Enter` key listener to intercept the event 👍

PR #1229, opened 2025-02-07
ERA-11088: Support EFB in ER > List v2 event types

luixlive:
### What does this PR do?
Implement the request of event types v2. In order to achieve it, a few things had to be done here:
- Mocks: since the v2 endpoint for getting event types is not ready yet, the implementation in this PR was done consuming mocks from the internal `mock-api` used in `das-web-react`.
  - Enable the `mock-api` in the `docker-compose.yml` file.
  - Uncomment the `mock-api` reverse proxy in the local Nginx `nginx/default.conf`.
  - Rework the `mock-api/Dockerfile.dev` and update to Node 22.
  - Basically, redo our entire mock server. We were using [json-server](https://github.com/typicode/json-server/tree/v0) which has been in beta for years, has over 600 issues in Github and there are lots of basic API scenarios that you can't cover with this framework. So I removed it in favor of a simple custom Express server that captures the desired routes and returns mocks, either static fixtures or programmatically built data.
  - Upgrade the dependencies of the `mock-api`.
  - Add routes and mocks for v2 event types and event type schemas.
- Augment our `src/ducks/event-types.js`: it now fetches event types from the v2 endpoint.
  - Add the new `REACT_APP_MOCK_EVENTTYPES_V2_API` environment variable to enable the mocks.
  - Add the new `API_V2_URL` environment variable to build the endpoint URLs.
  - Fetch the v2 event types in the `fetchEventTypes` action creator and merge them with the v1 event types, adding a `version` property for easy recognition.
- Add new Redux reducer for event categories: the v2 event types don't include the entire category object, just the category value to reference them and decrease the response size. This means now we have to fetch the event categories separately and save them in our local store.
  - New `src/ducks/event-categories.js` that includes a `fetchEventCategories` action creator to request them from the existing endpoint `/api/v1.0/activity/events/categories` and a new reducer that catches the success action to store the event categories.
  - This action creator gets called from `src/App.js` just like other basic ER data needed to run the system like `fetchEventTypes`, `fetchEventsSchema`, `fetchMaps`, `fetchSystemStatus`, and so on...
- Update our event types Redux selectors: specifically, I needed to see the version of the event type in order to do calculations that involves the category, but I also optimized some selectors with memoization as the official docs suggest.
- Update the `src/ducks/event-schemas.js` file: it now processes the new v2 endpoint for fetching v2 event type schemas.

### How does it look
There are no visual changes other than the fact that if you run the environment in your local machine you will see v1 and v2 event types coexisting and rendering their forms.

### Relevant link(s)
* [ERA-11088](https://allenai.atlassian.net/browse/ERA-11088)
* **There is no feature environment** since the changes are only visible if you enable the `mock-api` which only works when running our docker compose file.

- I'd recommend pulling the branch locally and taking the time to navigate the files and understand the new `mock-api` Express server. It's very simple and straightforward.
- Then, I'd go to the `App.js` file and see the addition of the new `fetchEventCategories`.
- Next, the `event-categories.js` duck would make a lot of sense.
- Finally, the implementation of the new `fetchEventTypes`, which gets and merges the V2 event types.
- After that, the rest of the code will make sense. Being the last important part to understand the `fetchEventTypeSchema` action creator, that gets and stores the schemas depending on if their associated event type is v1 or v2.
- The rest of the files are just minor changes needed to make these changes work and boy scout improvements.

[ERA-11088]: https://allenai.atlassian.net/browse/ERA-11088?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.env
```diff
@@ -5,11 +5,12 @@ REACT_APP_MAPBOX_TOKEN=pk.eyJ1IjoidmpvZWxtIiwiYSI6ImNpZ3RzNXdmeDA4cm90N2tuZzhsd3
 REACT_APP_DAS_HOST=''
 REACT_APP_DAS_AUTH_TOKEN_URL=/oauth2/token/
 REACT_APP_DAS_API_URL=/api/v1.0/
+REACT_APP_DAS_API_V2_URL=/api/v2.0/
```
luixlive, line 8: This and the next file are only the addition of new environment variables:
- `REACT_APP_DAS_API_V2_URL`
- `REACT_APP_MOCK_EVENTTYPES_V2_API`
I think they are self-explanatory.

docker-compose.yml
```diff
@@ -8,13 +7,13 @@ services:
...
       - .:/app
     restart: on-failure
-  # mock_api:
-  #   build:
-  #     context: ./mock-api
-  #     dockerfile: Dockerfile.dev
-    # volumes:
-    #   - /app/node_modules
-    #   - ./mock-api:/app
+  mock_api:
```
luixlive, line 10: We re-enable our mock-api for the docker compose command

mock-api/Dockerfile.dev
```diff
@@ -1,10 +1,14 @@
-FROM node:15.0.1-alpine
+FROM node:22-alpine
```
luixlive, line 1: Upgrade to Node 22 and changes in the Dockerfile after moving from `json-server` to Express.

mock-api/api/v1.0/activity/events/index.js
```diff
@@ -0,0 +1,17 @@
+const express = require('express');
+
+const { generateEvent } = require('./utils');
+
+const events = express.Router();
+
+events.get('/', (_, res) => {
```
luixlive, line 7: Our new handler for `/api/v1.0/activity/events`

mock-api/api/v1.0/activity/events/utils.js
```diff
@@ -1,14 +1,9 @@
-const { polygon, featureCollection } =  require('@turf/helpers');
```
luixlive, line 1: This isn't new code, I just updated our legacy `faker` library to the new one and moved the file.

mock-api/api/v2.0/activity/eventtypes/fixtures.js
```diff
@@ -0,0 +1,88 @@
+const eventtypesFixture = {
```
luixlive, line 1: Fixtures for v2.0 event types and event type schemas.

mock-api/api/v2.0/activity/eventtypes/index.js
```diff
@@ -0,0 +1,19 @@
+const express = require('express');
+
+const { eventtypesFixture, eventtypeSchemaFixtures } = require('./fixtures');
+
+const eventtypes = express.Router();
+
+eventtypes.get('/', (_, res) => {
```
luixlive, line 7: Our new handlers for the v2.0 endpoints!

mock-api/app.js
```diff
@@ -0,0 +1,16 @@
+const express = require('express');
```
luixlive, line 1: The root Express server file.

nginx/default.conf
```diff
@@ -2,17 +2,17 @@ upstream web {
   server web:3000;
 }
 
-# upstream mock_api {
-#   server mock_api:8080;
-# }
+upstream mock_api {
```
luixlive, line 5: Enable the reverse proxy to forward all the `http://localhost:9000/api/...` calls to the mock server.

src/App.js
```diff
@@ -1,99 +1,86 @@
-import React, { createContext, memo, useCallback, useEffect, useContext, useState } from 'react';
```
luixlive, line 1: Lots of changes because I removed the old `connect` function and replaced it with `useDispatch` and `useSelector`. The only new thing here is calling the new `fetchEventCategories` action creator.

src/DetailViewComponents/ActivitySection/ContainedReportListItem/index.js
```diff
@@ -48,10 +49,10 @@ const ContainedReportListItem = ({ cardsExpanded, onCollapse, onExpand, report }
   }, [dispatch, report.id, reportFromEventStore]);
 
   useEffect(() => {
-    if (!reportSchemas) {
+    if (!!eventType && !reportSchemas) {
       dispatch(fetchEventTypeSchema(report.event_type, report.id));
```
luixlive, line 53: Since we need to read the version of the event type before getting the schemas (we need to know if we should use the v1.0 or v2.0 schemas endpoint) now we don't fetch the schema until the event type is loaded.

src/ReportManager/ReportDetailView/index.js
```diff
@@ -663,10 +668,10 @@ const ReportDetailView = ({
   ]);
 
   useEffect(() => {
-    if (!!reportForm && !reportSchemas) {
+    if (!!reportForm && !!eventType && !reportSchemas) {
       dispatch(fetchEventTypeSchema(reportForm.event_type, reportForm.id));
```
luixlive, line 672: Same than before, we shouldn't get the schema until we have the event type loaded.

src/ReportTypeMultiSelect/index.js
```diff
@@ -51,7 +58,7 @@ const ReportTypeMultiSelect = (props) => {
 
   const filteredEventTypes = filter.length ? filterEventTypes(eventTypes, filter) : eventTypes;
 
-  const itemsGroupedByCategory = mapReportTypesToCategories(filteredEventTypes);
+  const itemsGroupedByCategory = mapEventTypesToCategories(filteredEventTypes, eventCategories);
```
luixlive, line 61: Now the `mapEventTypesToCategories` helper needs the event categories too since v2.0 event types don't include the full category objects, so we need a way to map them.

src/ducks/event-categories.js
```diff
@@ -0,0 +1,37 @@
+import axios from 'axios';
```
luixlive: Our new duck for event-categories!

src/ducks/event-schemas/index.js
```diff
@@ -0,0 +1,122 @@
+import axios from 'axios';
```
luixlive, line 1: This duck is not new but I moved it to a folder so it can have a `utils` file. The new stuff here is all that is related to the v2.0 schemas fetching.

src/ducks/event-schemas/sanitizeSchemas.js
```diff
@@ -0,0 +1,313 @@
+import isString from 'lodash/isString';
```
luixlive, line 1: Nothing of this is new code, just moved it to a place where it makes more sense and causes less confusion.

src/ducks/event-types.js
```diff
@@ -1,35 +1,42 @@
...
 
+// Actions
 export const FETCH_EVENT_TYPES_SUCCESS = 'FETCH_EVENT_TYPES_SUCCESS';
 
-export const fetchEventTypes = () => dispatch => axios.get(EVENT_TYPE_API_URL)
-  .then((response) => {
-    dispatch(fetchEventTypesSuccess(response));
-  });
+// Action creators
+export const fetchEventTypes = () => async (dispatch) => {
```
luixlive: Here we fetch and merge the event types v2 with the event types v1.

src/selectors/event-types.js
```diff
@@ -1,11 +1,44 @@
 import { createSelector } from 'reselect';
 
-const getEventTypes = ({ data: { eventTypes } }) => eventTypes;
```
luixlive, line 3: Huge improvement, refactoring and addition to our event-types related selectors. We really need to do this for the rest of ours selectors and even add new memoized selectors with `reselect`. Seems like a very nice step for client side optimizations.

src/utils/event-types.js
```diff
@@ -44,35 +44,56 @@ export const calcIconColorByPriority = (priority) => {
   }
 };
 
-export const mapReportTypesToCategories = eventTypes => eventTypes
+export const mapEventTypesToCategories = (eventTypes, eventCategories) => {
```
luixlive: This method needed to be re-done with the addition of event types v2.0 and the event categories independent store. Of course we need to check the version of each event type to know how to handle it. I also did an optimization on the amount of O(n) operations (foreach, map, reduce...) calls and I transformed our categories into an object so we don't need to iterate an array every single time we want to fetch one of them by value. Even though I didn't measure it, I can say confidently that this method should be way faster.

mock-api/api/v2.0/activity/eventtypes/fixtures.js
```diff
@@ -0,0 +1,88 @@
...
+    },
+  ],
+  status: {
+    code: 200,
+    message: 'OK',
+  },
+};
+
+const eventtypeSchemaFixtures = {
+  animal_control_v2: {
```
netomo, line 65: not sure how exactly is being used the `eventtypeSchemaFixtures`, I have added two endpoints for schemas:

One serves a multiple schemas at the same time, it supports the same query parameters as the normal list of event types, but it returns a dict, keyed by the event types values...

The othe serves only one schema, uses the same url as the event type detail but with the addition of the `schema/` to the url e.g. `/api/v2.0/activity/eventtypes/fire_v2/schema/`

The body in both cases would not be wrapped by this structure `{data: {}, status: {code: 200, message: 'ok'}}`, it would be a standard json directly

luixlive: image

Here you can see how it is used. Each property of this object is a different response returned by the endpoint when the user calls `/api/v2.0/activity/eventtypes/:eventTypeValue`.

Nows, regarding the body not being wrapped... Hmm you shared with me this image over Slack where we can see the JSON stringified and wrapped just like these mocks:
image

netomo: Got it! Nice mock implementation and sorry about the confusion on that, I shared it just as a preview, the schema body itself would be just a json, not wrapped and not as string.

.env
gaboDev: Following the current convention, shouldn't `REACT_APP_MOCK_EVENTTYPES_V2_API` be `REACT_APP_MOCK_EVENT_TYPES_V2_API` instead?

gaboDev (approved):
LGTM 🥳

.env
luixlive: Well I'd agree in separating the two words. But I based the name of the variable on the name of the endpoint which is `/eventtypes` as a single word. That made sense to me.

PR #1235, opened 2025-02-20
ERA-11055: track styles: settings option to select current timezone (client-side only, just for tracks when this is toggled on)

luixlive:
### What does this PR do?
Add the time zone selector in the new "Time of Day Settings" menu inside the map's subject track legend.

### How does it look
image

### Relevant link(s)
* [ERA-11055](https://allenai.atlassian.net/browse/ERA-11055)
* [Env](https://era-11055.dev.pamdas.org)

`src/SubjectTrackLegend/TimeOfDaySettings/TimeZoneSelect/index.js` has all the important stuff really.

Make sure to also understand the trivial additions in `src/ducks/tracks/index.js` to support the new `timeOfDayTimeZone` setting in the tracks reducer.

[ERA-11055]: https://allenai.atlassian.net/browse/ERA-11055?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tiffanynwong:
It's beautiful 😍. Just some small things. 

1. can we change the time zone text from blue to dark grey? That way the text won't look clickable. 
Screenshot 2025-02-20 at 11 06 58
2. Can we use the smaller carret we use in the event forms instead of the big chevron. Looks too much like the expand/collapse button above it.
Screenshot 2025-02-20 at 11 09 12
Screenshot 2025-02-20 at 11 09 27

3. In the long format name, is it possible to include the country? Searching by city is really nice but also a lot of our users aren't in cities so they might try country first especially for countries that have one timezone. 
Screenshot 2025-02-20 at 11 13 05

luixlive:
Thanks @tiffanynwong I'll work on the first two items.

> 3\. In the long format name, is it possible to include the country? Searching by city is really nice but also a lot of our users aren't in cities so they might try country first especially for countries that have one timezone.

Regarding this one, that will actually be very complicated 😅  Those texts are not hardcoded or calculated by us. They are returned from the native Javascript API [Intl.DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat) and are part of the IANA standard: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones . The standard does not include a country, so it wouldn't just be difficult to calculate that, it's breaking the standard. Since they have never been part of the standard, user's shouldn't be used to know any time zone by country name anyway 🤔 

Edit: Also, some of those time zone names are used by several countries. A time zone doesn't correspond to a city or a country, it corresponds to a region, and as such they may be used by several cities and countries at the same time. So It wouldn't really make sense to try to map them to a single country.

tiffanynwong:
Ok no worries, it would have been nice to add if it was an option the library provided. But no need to do it since it doesn't.

src/SubjectTrackLegend/TimeOfDaySettings/TimeZoneSelect/index.js
```diff
@@ -0,0 +1,164 @@
+import React, { useEffect, useMemo, useRef } from 'react';
```
luixlive: New time zone select component:
image

src/SubjectTrackLegend/TimeOfDaySettings/TimeZoneSelect/index.js
```diff
@@ -0,0 +1,164 @@
...
+      {data.title}
+
+      <br />
+
+      <span className={styles.description}>{data.description}</span>
+    </p>
+  </components.Option>
+</div>;
+
+const getTimeZoneOffsetParts = (ianaTimeZone, locales) => {
```
luixlive: This helper method receives an IANA time zone and the user's locale from the `i18n` object, and with that information it uses the `Intl.DateTimeFormat` API to calculate the time zone parts we need to build the `Option` component of the selector. For example:
**Input**
- `ianaTimeZone`: `'America/Mexico_City'`
- `locales`: `'es'`

**Output**
```
{
  hours: 6,
  ianaTimeZone: 'America/Mexico_City',
  longTimeZoneName: 'Hora Estándar Central',
  minutes: 0,
  sign: '-',
}
```

That information is enough to sort the time zones by offset and build the `Option` labels like: `(UTC-06:00) America / Mexico City - Hora Estándar Central`.

src/SubjectTrackLegend/TimeOfDaySettings/TimeZoneSelect/index.js
```diff
@@ -0,0 +1,164 @@
...
+      ianaTimeZone,
+      longTimeZoneName,
+      minutes: match[3] ? parseInt(match[3]) : 0,
+      sign: match[1],
+    };
+  }
+  return { hours: 0, ianaTimeZone, longTimeZoneName, minutes: 0, sign: '+' };
+};
+
+const compareTimeZoneOffsetParts = (timeZoneOffsetPartsA, timeZoneOffsetPartsB) => {
```
luixlive: A compare method for the `timeZoneParts.sort()`. It sorts the time zone by two dimensions: offset amount and then alphabetically.

src/SubjectTrackLegend/TimeOfDaySettings/TimeZoneSelect/index.js
```diff
@@ -0,0 +1,164 @@
...
+
+const TimeZoneSelect = () => {
+  const dispatch = useDispatch();
+  const { i18n, t } = useTranslation('tracks', { keyPrefix: 'subjectTrackLegend.timeOfDaySettings.timeZoneSelect' });
+
+  const timeOfDayTimeZone = useSelector((state) => state.view.trackSettings.timeOfDayTimeZone);
+
+  const inputRef = useRef();
+
+  const options = useMemo(() => Intl.supportedValuesOf('timeZone')
```
luixlive, line 96: Here we build the `Select` options. `Intl.supportedValuesOf('timeZone')` has wide support across browsers and it returns all the supported values of IANA time zones. Getting the time zones this way (programmatically) will keep us up to date to any changes to the standard, not like hardcoding the time zones ourselves in a JSON file.

The array of IANA time zones is mapped to get their parts, then sorted and finally we build the options array with 4 properties:
- `description`: custom property where we store the `longTimeZoneName` so we can render it greyed out below the title.
- `label`: `react-select` property. By default, this is the one rendered in the option text, however, we customized the component to not render this one. We still need it because this is the text used for search, so in here we want to include everything (UTC offset, IANA time zone and long time zone name).
- `title`: custom property where we store the UTC offset and the humanized IANA time zone to render as the title of the option text.
- `value`: `react-select` property of the value that will be associated to this option. Of course, we want the non-humanized IANA time zone here since that is how we store the value in Redux.

image

src/SubjectTrackLegend/TimeOfDaySettings/TimeZoneSelect/index.js
```diff
@@ -0,0 +1,164 @@
...
+        value: timeZoneOffsetParts.ianaTimeZone,
+      };
+    }), [i18n.language]);
+
+  const value = useMemo(
+    () => options.find((option) => option.value === timeOfDayTimeZone),
+    [options, timeOfDayTimeZone]
+  );
+
+  useEffect(() => {
```
luixlive, line 123: If there is no time zone selected yet, we set the user's time zone as the default.

src/ducks/tracks/index.js
```diff
@@ -126,6 +126,7 @@ export const TRACK_LENGTH_ORIGINS = { EVENT_FILTER: 'EVENT_FILTER', CUSTOM_LENGT
 export const SET_TRACK_SETTINGS_DEFAULT_CUSTOM_TRACK_LENGTH = 'SET_TRACK_SETTINGS_DEFAULT_CUSTOM_TRACK_LENGTH';
 export const SET_TRACK_SETTINGS_IS_TIME_OF_DAY_COLORING_ACTIVE = 'SET_TRACK_SETTINGS_IS_TIME_OF_DAY_COLORING_ACTIVE';
 export const SET_TRACK_SETTINGS_LENGTH = 'SET_TRACK_SETTINGS_LENGTH';
+export const SET_TRACK_SETTINGS_TIME_OF_DAY_TIME_ZONE = 'SET_TRACK_SETTINGS_TIME_OF_DAY_TIME_ZONE';
```
luixlive, line 129: We add a new action, action creator, default state and reducer condition to process and store changes to the `timeOfDayTimeZone`.

tiffanynwong (approved):
Looks good 👍

gaboDev (approved):
LGTM! 🚀 

Friendly reminder about updating out i18n spreadsheet 🫡

PR #1236, opened 2025-02-21
ERA-11059: Update track+point selectors to calc time zone fractions for use by track color expression

gaboDev:
### What does this PR do?
- Adds logic to calculate time-of-day range of track points for color expression feature

### Relevant link(s)
* Tracking tickets: [ERA-11059](https://allenai.atlassian.net/browse/ERA-11059)
* [Hosted demo environments](https://era-11059.dev.pamdas.org)

[ERA-11059]: https://allenai.atlassian.net/browse/ERA-11059?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/SubjectTrackLegend/constants.js
```diff
@@ -0,0 +1,23 @@
+export const TIME_OF_DAY_RANGE_LEVELS = {
```
luixlive: I have two suggestions regarding these constants:
1- The naming `LEVEL` makes it sound like a video game haha, I get it's not straightforward to come with a good name, but I feel like something like `PERIOD` is a better fit.
2- Instead of having an object that maps an index to an index and then using the keys of it to map the hour ranges, why not we simply stick with a single array where the indexes are the keys and the ranges are the values:
```
export const TIME_OF_DAY_PERIODS = ['12:01 - 15:00', '15:01 - 18:00', ...];
```

src/SubjectTrackLegend/constants.js
```diff
@@ -0,0 +1,23 @@
...
+export const TIME_OF_DAY_RANGES = {
+  [TIME_OF_DAY_RANGE_LEVELS.LEVEL_1]: '12:01 - 15:00',
+  [TIME_OF_DAY_RANGE_LEVELS.LEVEL_2]: '15:01 - 18:00',
+  [TIME_OF_DAY_RANGE_LEVELS.LEVEL_3]: '18:01 - 21:00',
+  [TIME_OF_DAY_RANGE_LEVELS.LEVEL_4]: '21:01 - 00:00',
+  [TIME_OF_DAY_RANGE_LEVELS.LEVEL_5]: '00:01 - 03:00',
+  [TIME_OF_DAY_RANGE_LEVELS.LEVEL_6]: '03:01 - 06:00',
+  [TIME_OF_DAY_RANGE_LEVELS.LEVEL_7]: '06:01 - 09:00',
+  [TIME_OF_DAY_RANGE_LEVELS.LEVEL_8]: '09:01 - 12:00',
+  [TIME_OF_DAY_RANGE_LEVELS.LEVEL_9]: '12:01 - 15:00'
```
luixlive: This one is repeated. In the legend we are repeating it to make it obvious that it circles back, but logically we don't want to have two ranges for the same period.

src/SubjectTrackLegend/TimeOfDaySettings/index.js
```diff
@@ -10,18 +10,56 @@ import { ReactComponent as InformationIcon } from '../../common/images/icons/inf
...
-  { color: 'fandangoPink', key: 2, text: '18:01 - 21:00' },
-  { color: 'purplePlum', key: 3, text: '21:01 - 00:00' },
-  { color: 'majorelleBlue', key: 4, text: '00:01 - 03:00' },
-  { color: 'lapisLazuli', key: 5, text: '03:01 - 06:00' },
-  { color: 'spanishGreen', key: 6, text: '06:01 - 09:00' },
-  { color: 'green', key: 7, text: '09:01 - 12:00' },
-  { color: 'titaniumYellow', key: 8, text: '12:01 - 15:00' },
+  {
+    color: 'titaniumYellow',
+    key: TIME_OF_DAY_RANGE_LEVELS.LEVEL_1,
```
luixlive: I think these keys should stay "hardcoded" (not really because they are in a constant) here. These keys are just for React to recognize the components rendered in an iteration, they are not part of the logic of our time periods. And since the first and last item of the array are the same, we need a key to differentiate them here, but that won't be the case anywhere else.

src/SubjectTrackLegend/utils/index.js
```diff
@@ -0,0 +1,55 @@
+import {
```
luixlive: Why are we storing these utils under `SubjectTrackLegend` if they are not being used by that component? They are related to the business logic but it would be better to store them where they are used. Or if they are being used by several files, probably in our `src/utils`.

src/SubjectTrackLegend/utils/index.test.js
```diff
@@ -0,0 +1,14 @@
+import { getTimeOfDayRangeLevelBasedOnTime } from './';
+
+import { TIME_OF_DAY_RANGE_LEVELS } from '../constants';
+
+describe('SubjectTrackLegend - utils', () => {
+  test('calculate proper time of day range based on time', () => {
+    expect(
```
luixlive: This is nice, could we add a couple more scenarios with other time zones / dates. It could be part of the same test, just some more expects to cover more ground.

src/utils/datetime.test.js
```diff
@@ -18,4 +18,9 @@ describe('format dates', () => {
     expect(format(date, 'MMMM dd, YYYY')).toBe('diciembre 20, 1993');
   });
 
+  test('converts date to time based on time zone', () => {
+    const date = new Date('2025-02-21T03:00:07.940Z');
+    expect( getTimeInTimezone( date, 'America/Monterrey') ).toBe('21:00');
```
luixlive, line 23: Same than before, adding a couple more tests would be nice.

src/selectors/tracks/index.js
```diff
@@ -91,10 +92,38 @@ const selectSubjectTracks = createSelector(
     .map((subjectId) => tracks[subjectId])
 );
 
+
 export const selectSubjectTracksTrimmedToTrackTimeEnvelope = createSelector(
```
luixlive: Two things regarding this selector:
1- The name should be updated to tell the fact that it also calculates the time of day period (I know it's big enough but better to be extremely explicit) `selectSubjectTracksTrimmedToTrackTimeEnvelopeWithTimeOfDayPeriod` maybe?
2- We are not checking if the time of day functionality is active and still we are running the calculations for every single track each time this selector runs. We should read the `trackSettings` object and check if `isTimeOfDayColoringActive` before running the calculations.

luixlive (changes requested):
The core implementation is there and makes sense 💯 I added a couple of suggestions.

src/SubjectTrackLegend/utils/index.js
```diff
@@ -0,0 +1,55 @@
...
+    parseInt(hour)
+  );
+};
+
+export const getTimeOfDayRangeLevelBasedOnTime = (datetimeString, timeZone) => {
+  const trackPointTimeDate = buildDateBasedOnStringTime(getTimeInTimezone(new Date(datetimeString), timeZone));
+  let timeOfDayRangeLevel = null;
+
+  for (const [rangeLevel, range] of Object.entries(TIME_OF_DAY_RANGES)) {
+    const { from, to } = buildTimeOfDayRangeObjectFromRangeString(range);
```
luixlive: Ok, I wanted to separate this comment from the review because it's a big thing. I think this method could be optimized considerably.

Remember that this `getTimeOfDayRangeLevelBasedOnTime` will run for every single track. Subjects could have thousands of them. And ten you iterate up to 9 times in this `for` and in each one you  `buildTimeOfDayRangeObjectFromRangeString` for every single time period! That's a lot of processing!

I'll suggest to do something way simpler. We still need to iterate all the tracks of course, that is inevitable, but we could ease the processing involved in this function if we change the parsing string to dates, then extracting times, then doing date comparisons, etc... We could do a way simpler math and a set of if / else. What I suggest is to get the time (hours and minutes) of the track time in the timezone selected and get the total amount of minutes it has:
```
const trackTimeInTimeZone = getTimeInTimezone(new Date(track.time), timeOfDayTimeZone);
const [hourString, minutesString] = trackTimeInTimeZone.split(':');
const trackTimeMinutesInTimeZone = (parseInt(hour) * 60) + parseInt(minutes);
```

And then you can have a method with the minutes of each period!
```
const getTrackTimeOfDayPeriod = (trackTimeMinutes) => {
  if (totalMinutes >= 721 && totalMinutes <= 900) return 0; // 12:01 - 15:00
  if (totalMinutes >= 901 && totalMinutes <= 1080) return 1; // 15:01 - 18:00
  if (totalMinutes >= 1081 && totalMinutes <= 1260) return 2; // 18:01 - 21:00
  ...
}
```

And of course, you can have all of these minutes as part of the constants! 
```
export const TIME_OF_DAY_PERIODS = {
  0: {
    key: 0,
    rangeString: '12:01 - 15:00',
    rangeMinutesLow: 721,
    rangeMinutesHigh: 900,
  },
  ...
};
```

It's just something that came to my mind to make this method faster, and a bit easier since I'm pretty sure it will cut a lot of lines. But if you can think of something even better that would be great! Just remember, we are going to run this method thousdands of times for some sites, maybe more.

src/utils/tracks.js
```diff
@@ -356,3 +358,42 @@ export const addSocketStatusUpdateToTrack = (tracks, newData) => {
...
+  const [hour, min] = timeString.split(':');
+  return {
+    hour: parseInt(hour),
+    min: parseInt(min)
+  };
+};
+
+export const getTimeOfDayRangeLevelBasedOnTime = (datetimeString, timeZone) => {
+  const trackTimeInTimeZone = parseTimeStringToObject( getTimeInTimezone(new Date(datetimeString), timeZone) );
+  const trackTotalMinInTZ = (trackTimeInTimeZone.hour * 60) + trackTimeInTimeZone.min;
```
luixlive: There is a catch here for the hour `00:00`. In the constants you defined its corresponding amount of minutes as 1440, but this math operation will actually result in 0. Mathematically, the right value is 0, but it makes more sense to assign 1440 so the conditional of `trackTotal <= 1440` works correctly. What we just need to do here is to make sure that if the result is 0, we assign 1440:
```
// Instead of hardcodign 1440 we could use the constant.
const trackTotalMinInTZ = ((trackTimeInTimeZone.hour * 60) + trackTimeInTimeZone.min) || 1440;
```

src/selectors/tracks/index.js
```diff
@@ -3,7 +3,8 @@ import { differenceInCalendarDays, subDays } from 'date-fns';
 import uniq from 'lodash/uniq';
 
 import { TRACK_LENGTH_ORIGINS } from '../../ducks/tracks';
-import { trimTrackDataToTimeRange } from '../../utils/tracks';
+
+import { getTimeOfDayRangeLevelBasedOnTime, trimTrackDataToTimeRange } from '../../utils/tracks';
```
luixlive: We are still using the word `level` here, we should change it to period for consistency. `getTimeOfDayPeriod` or something like that.

src/selectors/tracks/index.test.js
```diff
@@ -1,7 +1,7 @@
 import {
   selectTrackTimeEnvelope,
   selectHeatmapSubjectTracksTrimmedToTrackTimeEnvelope,
-  selectSubjectTracksTrimmedToTrackTimeEnvelope,
+  selectSubjectTracksTrimmedToTrackTimeEnvelopeWithTimeOfDayPeriod,
```
luixlive, line 4: We should augment this tests to validate the right addition of the time of day period to the tracks.

src/utils/tracks.js
```diff
@@ -356,3 +358,42 @@ export const addSocketStatusUpdateToTrack = (tracks, newData) => {
   }
   return tracks;
 };
+
+const parseTimeStringToObject = (timeString) => {
```
luixlive: nit: very small opinion on a trivial matter, but I don't think we will be reusing this and it's a very simple logic so I would say it's not necessary to have a separate function.

src/utils/tracks.js
```diff
@@ -356,3 +358,42 @@ export const addSocketStatusUpdateToTrack = (tracks, newData) => {
...
+    secondTDPeriod,
+    thirdTDPeriod,
+    fourthTDPeriod,
+    fifthTDPeriod,
+    sixthTDPeriod,
+    seventhTDPeriod,
+    eighthTDPeriod,
+  ] = TIME_OF_DAY_PERIODS;
+
+  if ( trackTotalMinInTZ >= firstTDPeriod.rangeMinutesMin && trackTotalMinInTZ <= firstTDPeriod.rangeMinutesMax ) return 0;
```
luixlive: I feel like all these repetitive lines could become a single `.findIndex`:
```
return TIME_OF_DAY_PERIODS.findIndex(
  (timeOfDayPeriod) => trackMinutesInTimeZone >= timeOfDayPeriod.rangeMinutesMin && trackMinutesInTimeZone <= timeOfDayPeriod.rangeMinutesMax
);
```

Also a nit: `trackTotalMinInTZ` took me a while to understand that `min` is minutes and not `minimum` so maybe `minutes` would be better.

src/SubjectTrackLegend/constants.js
```diff
@@ -0,0 +1,42 @@
+export const TIME_OF_DAY_PERIODS = [
```
luixlive: Still not convinced if this constant belongs here since we need the range minutes in other place. We could move these to the more global `src/constants`.

luixlive (approved):
Nice ✅  I still have some comments, but nothing to not approve the PR, the logic is sound.

src/selectors/tracks/index.js
```diff
@@ -91,10 +92,38 @@ const selectSubjectTracks = createSelector(
...
 
-export const selectSubjectTracksTrimmedToTrackTimeEnvelope = createSelector(
-  [selectSubjectTracks, selectTrackTimeEnvelope],
-  (subjectTracks, trackTimeEnvelope) => subjectTracks.map(
-    // Trim each subject tracks to the track time envelope.
-    (subjectTracks) => trimTrackDataToTimeRange(subjectTracks, trackTimeEnvelope.from, trackTimeEnvelope.until)
+
+export const selectSubjectTracksTrimmedToTrackTimeEnvelopeWithTimeOfDayPeriod = createSelector(
+  [selectSubjectTracks, selectTrackTimeEnvelope, selectTrackSettings],
+  (subjectTracks, trackTimeEnvelope, { timeOfDayTimeZone }) => subjectTracks.map(
```
luixlive: I just noticed something I missed before approving! I had already mentioned this in my first pass and I think this is important. We should receive the `trackSettings` object and only calculate the `timeOfDayPeriod`s if the functionality is active. We shouldn't be adding all this processing if the user is not going to use it.
```
(subjectTracks, trackTimeEnvelope, trackSettings) => subjectTracks.map((subjectTrack) => {
  ...
  timeOfDayRangeLevel: trackSettings.isTimeOfDayColoringActive
    ? getTimeOfDayRangeLevelBasedOnTime(properties.time, trackSettings.timeOfDayTimeZone)
    : undefined,
});
```

Also, we are still using the word levels in `timeOfDayRangeLevel`. `timeOfDayPeriod` is consistent and it makes more sense 🤠

JoshuaVulcan (approved):

PR #1238, opened 2025-02-25
Fix broken tests in develop branch

luixlive:


src/ducks/event-schemas/index.test.js
```diff
@@ -20,7 +20,7 @@ const server = setupServer(
   http.get(`${EVENT_TYPE_SCHEMA_API_URL}/snare_rep`, () => HttpResponse.json({ data: snareSchemaV1 })),
   http.get(
     EVENT_TYPE_SCHEMA_V2_API_URL('snare_rep'),
-    () => HttpResponse.json({ data: JSON.stringify(snareSchemaV2) })
```
luixlive, line 23: This makes sense, Ernesto asked to not wrap the response of schemas as usual and just leave the pure JSON.

src/selectors/tracks/index.test.js
```diff
@@ -76,14 +76,6 @@ describe('Selectors - Tracks', () => {
         until: null,
       });
     });
-
-    test('builds the track time envelope if the track length origin is the event filter', () => {
```
luixlive, line 80: Removed this test since its unstable. It depends on the server timezone, thinking on a creative solution to keep this branch of the code tested could be done later.

JoshuaVulcan (approved):

PR #1243, opened 2025-02-27
ERA-11057: track styles: track color expression

gaboDev:
### What does this PR do?
- Adds time-of-day coloring expression to track points 
- Adds batch support to useMapSource and useMapLayer hooks

### How does it look
image

### Relevant link(s)
* Tracking tickets: [ERA-11057](https://allenai.atlassian.net/browse/ERA-11057)
* [Hosted demo environments](https://era-11057.dev.pamdas.org)

[ERA-11057]: https://allenai.atlassian.net/browse/ERA-11057?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/hooks/index.js
```diff
@@ -74,46 +74,7 @@ export const useMapEventBinding = (eventType = 'click', handlerFn = noop, layerI
...
-        });
-      }
-    };
-  }, [sourceId, map]);
-
-  return map?.getSource(sourceId);
-};
-
-export const useMapLayer = (layerId, type, sourceId, paint, layout, config = {}) => {
+export const useMapLayerZZZZ = (layerId, type, sourceId, paint, layout, config = {}) => {
```
luixlive: What is this `ZZZZ` stuff?

src/hooks/useMapLayer/index.js
```diff
@@ -0,0 +1,167 @@
...
+          const { id, type, sourceId, paint = {}, layout = {}, filter, condition, before } = config;
+          const conditionValue = condition ?? true;
+          const beforeValue = before || defaultConfig.before;
+
+          if (map.getSource(sourceId) && conditionValue && !map.getLayer(id)){
+            const layerObj = {
+              id,
+              source: sourceId,
+              type,
+              layout: layout || {},
```
luixlive: You are already defining layout as `layout = {}` so adding a default here again is redundant.

src/hooks/useMapLayer/index.js
```diff
@@ -0,0 +1,167 @@
...
+              paint: paint || {}
+            };
+
+            // Only add filter if it's defined and is an array
```
luixlive: It's nice to add comments explaining logic that is not obvious, but this one is very trivial and doesn't tell me anything that I can't understand by myself by looking at the code. Something more interesting would be to explain why it needs to be an array. Maybe it's just the way Mapbox expects it, you could explain that.

src/hooks/useMapLayer/index.js
```diff
@@ -0,0 +1,167 @@
...
+  // Update layout properties for existing layers
+  useEffect(() => {
+    if (map) {
+      layersConfigs.forEach(config => {
+        if (config?.id && config.layout){
+          const { id, layout, condition } = config;
+
+          if (( condition ?? true ) && map.getLayer(id) && layout) {
+            Object.entries(layout).forEach(([key, value]) => {
+              map.setLayoutProperty(id, key, value);
```
luixlive: A personal recommendation and something I always try to do is to name the parameters I provide to a method (either if its local or an external API method like this one) as they are called within the method itself. It makes it easier to read documentation and when navigating the code to keep track of which variable corresponds to which. A simple example here would be to call these params `layerId`, `name` and `value` as the docs suggest: https://docs.mapbox.com/mapbox-gl-js/api/map/#map#setlayoutproperty

src/selectors/tracks/index.js
```diff
@@ -97,35 +100,18 @@ export const selectSubjectTracksTrimmedToTrackTimeEnvelopeWithTimeOfDayPeriod =
...
-                timeOfDayPeriod: isTimeOfDayColoringActive
-                  ? getTimeOfDayPeriodBasedOnTime(properties.time, timeOfDayTimeZone)
-                  : undefined
-              }
-            };
-          })
+      return isTimeOfDayColoringActive
+        ? {
+          ...trimmedTrackData,
+          twoPointLineStringTrackPoints: buildFeatureCollectionOfTwoPointLineStringSegments(trimmedTrackData.track, timeOfDayTimeZone)
```
gaboDev: We need to split all track point by segments formed by a line string object of 2 points having: the coors and times of point A and B, also assigning a startColor and endColor based on the time of each point

This object will allow us to set an specific gradient color to each line, being a workaround of the issue of [MapBox not being able to apply dynamically data-drive stop colors for a gradient line](https://github.com/mapbox/mapbox-gl-js/issues/8977)

src/hooks/useMapLayer/index.js
```diff
@@ -0,0 +1,167 @@
...
+    }
+  }, [map, layersConfigs, layerConfig]);
+
+  // Update paint properties for existing layers
+  useEffect(() => {
+    if (map) {
+      layersConfigs.forEach(config => {
+        if (config?.id && config?.paint){
+          const { id, paint, condition } = config;
+          if (( condition ?? true ) && map.getLayer(id) && paint) {
```
luixlive: Ok I see four things repeating over and over again in these useEffects that should be updated,  instead of adding a comment for each I'll summarize them here:
1. `condition ?? true` doesn't consider the case for `undefined`. If `condition` is either `null` or `undefined` I would say that there is no condition so it is truthy by default. The right way to check for that is `condition !== false`. This way we make sure that the only falsy value is `false`.
2. `layerConfig` is in the dependency array unnecessarily, impacting memoization checks performance.
3. Most of these `useEffects` have an `if` nested immediately below another `if`. This isn't clean code, we should remove one of those ifs: `if (config?.condition !== false && map.getLayer(config?.id) && config?.paint) { ... }` is logically equivalent to the two ifs you have making the code simpler and avoiding repetition of conditions.
4. A suggestion I already made: call the arguments we pass to the Mapbox API the same way they call them internally so the reading of the docs and the type suggestions done by VS code are trivial to analyze.

src/TracksLayer/track.js
```diff
@@ -53,24 +61,55 @@ const TrackLayer = ({ before, id, lineLayout, linePaint, onPointClick, showTimep
...
-  useMapSource(pointSourceId, trackData.points);
-
-  useMapLayer(
-    layerId,
-    'line',
+  const {
+    sourcesConfigs,
+    layersConfigs,
+    hasTimeOfDaySegments
+  } = useMemo(() => generateMapSourcesAndLayersBasedOnTwoLineTrackPointsSegments(
```
gaboDev: Here, two things are happening internally:

1. We grab the two-line segments calculated on the selector and create groups of the all possible combination of pairs of color stops (64), this with the purposes of reduce the amount of sources/layers created
2. For each pair, We create layer/source config objects to be passed to the map

src/hooks/useMapLayer/index.js
```diff
@@ -0,0 +1,167 @@
...
+        if (( condition ?? true ) && map.getLayer(id)) {
+          map.setLayerZoomRange(id, minZoomValue, maxZoomValue);
+        }
+      });
+    }
+  }, [map, layersConfigs, defaultConfig, layerConfig]);
+
+  // Cleanup on unmount
+  useEffect(() => {
+    const refs = layerIdsRef.current;
```
luixlive: This variable feels unnecessary, you are renaming a ref to ref and it's even counterintuitive since refs have the current property and you are removing it. We could simply call `layerIdsRef.current.forEach((layerId) => ...);`

gaboDev:
Most of the changes of this PR are related with the refactor of useMapSource and useMapLayer, since the approach for the coloring expression required some grouping it was necessary to adapt those hooks to be use with batches of object configs.

The actual approach can be seen in these files:

1. src/selectors/tracks/index.js
2. src/TracksLayer/track.js
3. src/TracksLayer/utils/index.js

src/hooks/useMapLayer/index.js
```diff
@@ -0,0 +1,167 @@
...
+  useEffect(() => {
+    if (map) {
+      layersConfigs.forEach(config => {
+        if (!config || !config.id) return;
+
+        const { id, condition, minZoom, maxZoom } = config;
+        const minZoomValue = minZoom || defaultConfig.minZoom || MIN_ZOOM;
+        const maxZoomValue = maxZoom || defaultConfig.maxZoom || MAX_ZOOM;
+
+        if (( condition ?? true ) && map.getLayer(id)) {
```
luixlive: Most of these `useEffects` include in their conditions this exact validation, I guess we could abstract it out to a method:
```
const shouldUpdateMapLayer = (config) => config?.condition !== false && config?.id && map.getLayer(config.id);
```

src/hooks/useMapSource/index.js
```diff
@@ -0,0 +1,74 @@
...
+
+  useEffect(() => {
+    // Initialize sources that don't exist yet
+    if (map) {
+      sourcesConfigs.forEach(config => {
+        if (!!config?.id && !!config?.data){
+          const { id, data, options = {} } = config;
+          const sourceConfig = { ...defaultConfig, ...options };
+
+          if (!map.getSource(id)) {
```
luixlive: Ok, no need to add a bunch of comments here too. I see some of the comments I added for the other hook will apply here too. Saving conditionals, cleaning dependency arrays, may be abstracting a function to do the repetitive checks...

src/hooks/useMapSource/index.js
```diff
@@ -0,0 +1,74 @@
...
+  useEffect(() => {
+    let timeouts = [];
+    sourcesConfigs.forEach(config => {
+      if (!!config?.id && !!config?.data){
+        const { id, data, options = {} } = config;
+        if (!!options.enabled){
+          const timeout = window.setTimeout(() => {
+            const source = map?.getSource?.(id);
+            if (source) {
+              source?.setData?.(data);
```
luixlive: I think this conditional chain in `source?.` isn't necessary, we just checked that source is defined in the `if`

src/hooks/useMapSource/index.js
```diff
@@ -0,0 +1,74 @@
...
+  }, [map, sourcesConfigs, defaultConfig]);
+
+  // Update data for existing sources
+  useEffect(() => {
+    let timeouts = [];
+    sourcesConfigs.forEach(config => {
+      if (!!config?.id && !!config?.data){
+        const { id, data, options = {} } = config;
+        if (!!options.enabled){
+          const timeout = window.setTimeout(() => {
```
luixlive: Why do we need to run this within a `timeout`? It has a smell...

src/hooks/useMapSource/index.js
```diff
@@ -0,0 +1,74 @@
...
+            if (map?.getSource(id)) {
+              map.removeSource(id);
+            }
+          });
+        });
+      }
+    };
+  }, [map]);
+
+  return sourcesConfigs.map((source) => map?.getSource(source.id));
```
luixlive: We should be consistent. Evey time we iterate the `sourcesConfigs` we call the iterated value `config`, but in here we call it `source`. Why not to call it explicitly `sourceConfig` in singular? This way it's explicit and consistent. Same applies for the other hook. Also, everytime we iterate it we check that the config is defined and has an id. In here we don't do that check, we just call `source.id` like if the presence of that property was warrantied... Then, why are we doing checks in other places? If we know for sure we are going to receive it, we shouldn't do validations anywhere. If we don't, then we should do validations everywhere.

src/BaseLayerRenderer/TileLayerRenderer.js
```diff
@@ -25,7 +26,7 @@ const SourceComponent = ({ id, tileUrl, sourceConfig }) => {
     ...sourceConfig,
   }), [sourceConfig, tileUrl]);
 
-  useMapSource(id, null, config);
+  useMapSource({ id, data: {} }, config);
```
luixlive: It'd be nice to support sending `data: null` (or simply not sending it at all) and add the right check validations / defaults in the hook so we don't have to send an empty object.

src/hooks/useMapSource/index.js
```diff
@@ -0,0 +1,74 @@
+import { useContext, useEffect, useMemo, useRef } from 'react';
+
+import { MapContext } from '../../App';
+
+const useMapSource = (sourceConfig, defaultConfig = { type: 'geojson' }) => {
```
luixlive: I find very confusing to have two parameters, both called config. I think there should be a better naming for both, or at least one of them. Same applies for the other hook.

src/TracksLayer/track.js
```diff
@@ -41,8 +48,9 @@ const TIMEPOINT_LAYER_PAINT = {
 
 const TrackLayer = ({ before, id, lineLayout, linePaint, onPointClick, showTimepoints, trackData }) => {
   const map = useContext(MapContext);
+  const { isTimeOfDayColoringActive } = useSelector(selectTrackSettings);
 
-  const trackId = id || trackData.track.features[0].properties.id;
+  const trackId = id || (trackData.track?.features?.[0]?.properties?.id || 'unknown-track');
```
luixlive: Why do we need now to add these conditional chaining and default?

src/TracksLayer/utils/index.js
```diff
@@ -0,0 +1,81 @@
...
+
+    const key = `${segment.properties.startColor}|${segment.properties.endColor}`;
+    if (!segmentsByColorPair[key]) {
+      segmentsByColorPair[key] = [];
+    }
+    segmentsByColorPair[key].push(segment);
+  });
+
+  if (segmentsWithMissingColors > 0){
+    console.warn('Segments with missing colors:', segmentsWithMissingColors);
```
luixlive: Do we want to keep this logs in the console? What's the purpose?

src/TracksLayer/utils/index.js
```diff
@@ -0,0 +1,81 @@
...
+      },
+      layout: layerLayout,
+      options: layerOptions
+    });
+  });
+
+  return {
+    sourcesConfigs: sources,
+    layersConfigs: layers,
+    hasTimeOfDaySegments: sources.length > 0
```
luixlive: I feel like we shouldn't return this variable. It's easy to calculate it in the other file, the responsibility of this method is just to return the sources and layer configs. Also, the name of this method suggests that it creates the layers and sources, but it just calculates it configurations. An alternative name that explains better its purpose: `getTimeOfDaySourceAndLayerConfigurations`.

src/selectors/tracks/index.js
luixlive: I understand the reasoning behind this naming `twoPointLineStringTrackPoints` (for both the property and the method) right now, but it worries me that it will not be intuitive when we come back  two months later and read this code. It feels a bit obscure. Could you think of a name a bit more explicit about its relation with the time of day feature? `buildTimeOfDayFeatureCollection` maybe. And of course, add comments inside the method explaining that the feature collection consists of several line strings connecting each two coordinates.

luixlive (changes requested):
Just a first pass. The intention is clear. I haven't tested the code yet, we would need some nice data in the environment. Also I'll wait for the missing tests you mentioned. But for now I have several small suggestions to try to make this piece of code as readable as possible since it's a very specific feature.

src/hooks/index.js
gaboDev: That was removed already

src/hooks/useMapLayer/index.js
gaboDev: This one is a recommendation from the linter actually

src/hooks/useMapLayer/index.js
gaboDev: For the point 2 and some other comment, not sure If the diff didn't get updated both those were solved before reviewing, I will take a look to the other points 🫡

src/TracksLayer/utils/index.js
gaboDev: It was added just for debugging purposes, removing it

src/hooks/useMapSource/index.js
```diff
@@ -0,0 +1,72 @@
...
+  const map = useContext(MapContext);
+  const sourceIdsRef = useRef([]);
+  const sourcesConfigs = useMemo(() => Array.isArray(sourceConfig) ? sourceConfig : [sourceConfig], [sourceConfig]);
+
```
JoshuaVulcan: Why not just have a plural name (such as `useMapSources`) and have it always accept an array of objects?  That would be clearer and simpler than accepting these two formats for `sourceConfig`, and take the mystery/magic out of naming it `useMapSource` but secretly allowing more than one source at a time in the hook

src/hooks/useMapSource/index.test.js
```diff
@@ -0,0 +1,90 @@
...
+
+    expect(map.getSource).toHaveBeenCalledTimes(3); // Get called 3 times by: adding source check, updating data, returning the source
+    expect(map.getSource).toHaveBeenCalledWith(sourceConfig.id);
+    expect(source.setData).toHaveBeenCalledTimes(1);
+    expect(source.setData).toHaveBeenCalledWith(sourceConfig.data);
+
+    jest.useRealTimers();
+  });
+
+});
```
JoshuaVulcan: Missing tests for multi-source scenarios

src/selectors/tracks/index.js
```diff
@@ -4,15 +4,18 @@ import uniq from 'lodash/uniq';
 
 import { TRACK_LENGTH_ORIGINS } from '../../ducks/tracks';
 
-import { getTimeOfDayPeriodBasedOnTime, trimTrackDataToTimeRange } from '../../utils/tracks';
+import {
+  trimTrackDataToTimeRange,
+  buildFeatureCollectionOfTwoPointLineStringSegments
```
JoshuaVulcan: I think Luis already mentioned, but `buildFeatureCollectionOfTwoPointLineStringSegments` is a pretty stanky and confusing name

src/selectors/tracks/index.js
JoshuaVulcan: `trackSegments` would suffice as a name here. that's really all it is.

JoshuaVulcan: @luixlive @gaboDev I would more recommend a very accurate-and-generic name. There's nothing time-of-day specific here, and as we increase style customization it might be pretty desirable and recyclable to have these segmented tracks. I like something simple and clean like `buildTrackSegments(track)` IMHO.

src/utils/tracks.js
```diff
@@ -363,7 +363,81 @@ export const getTimeOfDayPeriodBasedOnTime = (datetimeString, timeZone) => {
   const [hour, min] = getTimeInTimezone(new Date(datetimeString), timeZone).split(':');
   const trackTotalMinutesInTZ = ( (parseInt(hour) * 60) + parseInt(min) ) || 1440;
 
-  return TIME_OF_DAY_PERIODS.findIndex((timeOfDayPeriod) =>
+  const period = TIME_OF_DAY_PERIODS.find((timeOfDayPeriod) =>
     trackTotalMinutesInTZ >= timeOfDayPeriod.rangeMinutesMin && trackTotalMinutesInTZ <= timeOfDayPeriod.rangeMinutesMax
   );
+
+  return period ?? TIME_OF_DAY_PERIODS[0];
```
JoshuaVulcan: `return period ?? TIME_OF_DAY_PERIODS[0];`

A default here means we're straight-up presenting potentially incorrect information about the time of day for a segment....that doesn't feel particularly safe to me

src/utils/tracks.js
```diff
@@ -363,7 +363,81 @@ export const getTimeOfDayPeriodBasedOnTime = (datetimeString, timeZone) => {
...
+        endTime
+      },
+      geometry: {
+        type: 'LineString',
+        coordinates: [startCoors, endCoors]
+      }
+    });
+  }
+
+  return {
```
JoshuaVulcan: This file already imports the `featureCollection` utility fn from turf, why not just `return featureCollection(segments)` ? Seems cleaner to me.

JoshuaVulcan (changes requested):
Pretty light-touch first pass, mostly concerned with supplementing existing comments and looking at code style/naming conventions. It's close!

src/selectors/tracks/index.js
luixlive: Well, it sounds like we want to define "segment" like a feature collection of line strings joining pairs of points. I'm ok with that, it would be nice to document it somewhere, even just as a comment in the method so its clear what we mean with that.

But I differ about something else. You say @JoshuaVulcan that there is nothing related to the time of day feature, but there is: the start color and end color properties added to these "segments":
image

That is part of the method and it is specifically related to that feature. That's why I was thinking on a name that links this method with the feature. If we want to put a generic name to it I would recommend to not add those properties there, add them separately. But I don't know if it's really worth it and if we are going to reuse this method later. It sounds like we can leave it as is for now and then refactor if it's necessary.

src/selectors/tracks/index.js
```diff
@@ -97,35 +100,18 @@ export const selectSubjectTracksTrimmedToTrackTimeEnvelopeWithTimeOfDayPeriod =
...
-              ...otherFeaturesProps,
-              properties: {
-                ...properties,
-                timeOfDayPeriod: isTimeOfDayColoringActive
-                  ? getTimeOfDayPeriodBasedOnTime(properties.time, timeOfDayTimeZone)
-                  : undefined
-              }
-            };
-          })
+      return isTimeOfDayColoringActive
```
luixlive, line 91: One extra comment. I just realized while working on another ticket that you added this conditional but never actually tested it. This file already has unit tests, we are just missing one test to make sure that the time of day segments are added or not based on `isTimeOfDayColoringActive`. Let's add it please 😄

src/hooks/useMapSource/index.js
gaboDev: Got it!

src/selectors/tracks/index.js
gaboDev: Sure, I agree with `trackSegments` for the prop and `buildTrackSegments` for the method

src/utils/tracks.js
gaboDev: That one came from a test, not sure where, anyways I removed it and did several test to see if any time got no period object and indeed, there were some of them not getting set that property, so I added a validation and works great now!

src/selectors/tracks/index.js
gaboDev: Done!

src/hooks/useMapLayers/index.js
```diff
@@ -0,0 +1,168 @@
...
+
+const shouldUpdateMapLayer = (layerConfig, map) => {
+  return !!(
+    !!layerConfig?.id
+    && hasLayerCondition(layerConfig)
+    && !!map.getLayer(layerConfig.id)
+  );
+};
+
+const useMapLayers = (layerConfigsBatch = [], defaultConfig = {}) => {
```
luixlive: I still have the concern about having two configurations: the individual layer configuration and the default configuration. I feel it counter-intuitive. I'd just leave the layer configurations and the implementation can do the job of injecting common configs just before providing them to this hook. What do you think? 🤔

src/hooks/useMapLayers/index.js
```diff
@@ -0,0 +1,168 @@
+import { useContext, useEffect, useRef } from 'react';
+
+import { MapContext } from '../../App';
+import { MAX_ZOOM, MIN_ZOOM } from '../../constants';
+
+const hasLayerCondition = (layerConfig) => layerConfig?.options?.hasOwnProperty('condition')
+  ? layerConfig.options.condition
+  : true;
+
+const shouldUpdateMapLayer = (layerConfig, map) => {
```
luixlive: The `return` isn't needed and the logical expression has some unnecessary `!!` coercions since you are already wrapping it with one: 
```
const shouldUpdateMapLayer = (layerConfig, map) => !!(layerConfig?.id && hasLayerCondition(layerConfig) && map.getLayer(layerConfig.id))
```

src/hooks/useMapLayers/index.js
```diff
@@ -0,0 +1,168 @@
+import { useContext, useEffect, useRef } from 'react';
+
+import { MapContext } from '../../App';
+import { MAX_ZOOM, MIN_ZOOM } from '../../constants';
+
+const hasLayerCondition = (layerConfig) => layerConfig?.options?.hasOwnProperty('condition')
```
luixlive: The name of this method implies that it will return true if the layer has a condition property (independent of its value). Maybe `assertLayerCondition`? And here we sill have the issue about the condition being `null`.

src/hooks/useMapLayers/index.js
```diff
@@ -0,0 +1,168 @@
...
+
+  useEffect(() => {
+    if (map){
+      layerConfigsBatch.forEach(layerConfig => {
+        if (
+          layerConfig?.id
+          && layerConfig?.type
+          && layerConfig?.sourceId
+          && map.getSource(layerConfig.sourceId)
+          && layerConfig?.options?.condition !== false
```
luixlive: Wait, now I'm confused, why aren't we using the new method `hasLayerCondition`. I like the direct comparison to `false`, but then why sometimes we use the method and sometimes we don't? 
Also, why aren't we using `shouldUpdateMapLayer` here?

src/hooks/useMapLayers/index.js
```diff
@@ -0,0 +1,168 @@
...
+            id,
+            type,
+            sourceId,
+            paint = {},
+            layout = {},
+            options: {
+              filter,
+              before
+            } = {}
+          } = layerConfig;
```
luixlive: nit: I feel like extracting these variables from the object just adds a bunch of unnecessary lines to the file.

src/hooks/useMapLayers/index.js
```diff
@@ -0,0 +1,168 @@
...
+            paint: paint
+          };
+
+          const filterValue = filter || defaultConfig.filter;
+          if (Array.isArray(filterValue)) {
+            layerObj.filter = filterValue;
+          }
+
+          // Handle line-gradient and line-color conflict
+          if (type === 'line' && paint?.['line-gradient'] && paint?.['line-color']) {
```
luixlive: Two things:
1- I see a console warn here, is this information that important to be printed? Will it become just noise?
2- Is this hook's responsibility to handle this "conflict"? I feel like the implementator should be careful to not add both styles if the conflict with each other, instead of deleting a random property here.

src/hooks/useMapLayers/index.js
```diff
@@ -0,0 +1,168 @@
...
+  useEffect(() => {
+    if (map) {
+      layerConfigsBatch.forEach(layerConfig => {
+        const before = layerConfig?.options?.before || defaultConfig.before;
+        if (
+          layerConfig?.id
+          && before
+          && map.getLayer(layerConfig.id)
+        ){
+          map.moveLayer(layerConfig.id, before);
```
luixlive: Shouldn't we check if we really need to call this method? I mean, this useEffect will be called with any change in the config objects and for every single layer with a `before` property it will call the `moveLayer`. But, do we need to? Maybe they are already well located in the layer stack right? I don't know how much this could impact performance.

src/hooks/useMapLayers/index.js
```diff
@@ -0,0 +1,168 @@
...
+      });
+    }
+  }, [map, layerConfigsBatch, defaultConfig]);
+
+  useEffect(() => {
+    if (map) {
+      layerConfigsBatch.forEach(layerConfig => {
+        if ( shouldUpdateMapLayer(layerConfig, map) ) {
+          const { options: { minZoom, maxZoom } = {} } = layerConfig;
+          map.setLayerZoomRange(
```
luixlive: Same thing, maybe the zoom is already in the right level, but we are not checking that. We are just iterating and calling the `setLayerZoomRange` for every single layer when there is a single change in the config objects. Considering that now we are supporting batches that could have dozens of layers, this may have some performance impact.

src/hooks/useMapLayers/index.js
```diff
@@ -0,0 +1,168 @@
...
+            minZoom || defaultConfig.minZoom || MIN_ZOOM,
+            maxZoom || defaultConfig.maxZoom || MAX_ZOOM
+          );
+        }
+      });
+    }
+  }, [map, layerConfigsBatch, defaultConfig]);
+
+  useEffect(() => {
+    const refs = layerIdsRef.current;
```
luixlive, line 106: Still confused on why we redefine a ref into a more generic name ref

src/hooks/useMapSources/index.js
```diff
@@ -0,0 +1,65 @@
...
+      });
+    }
+  }, [map, sourceConfigsBatch, defaultConfig]);
+
+  useEffect(() => {
+    let timeouts = [];
+    sourceConfigsBatch.forEach(sourceConfig => {
+      const source = map?.getSource?.(sourceConfig?.id);
+      if (sourceConfig?.id && sourceConfig?.data && source){
+        const timeout = window.setTimeout(() => {
```
luixlive: bump to my original question, why do we use timeouts here?

luixlive (commented):
Not blocking this PR, I think @JoshuaVulcan has a better understanding on the reasoning behind and if he likes the approach I think it's ok too. Still, added some more comments on very minor things that I think could improve performance and code readability. Nice work, this ticket was a tricky one!

src/selectors/tracks/index.js
luixlive: Just consider the fact that it has logic related to the time of day period feature. `trackSegments` sounds very generic and doesn't tell what is really going on. We could remove that logic from there to keep it generic, or make sure that the name expresses the intent.

luixlive (approved):
Approving just for unblocking, but please read my latest suggestions, and I'd also recommend to wait for a second pass from @JoshuaVulcan since he has more insight on the reasoning behind this approach.

src/hooks/useMapLayers/index.js
gaboDev: shouldUpdateMapLayer implies that the layer should exist which is not required here, and regarding condition, I forgot to use the method here 😄

src/hooks/useMapLayers/index.js
gaboDev: Nice observation I can take a look to it

src/hooks/useMapLayers/index.js
gaboDev: There is a linter warn about accessing a ref in a unmount callback, so just to get ride of it I declared this outside that callback, I can renaming the variable or add a disable-lint line

src/hooks/useMapLayers/index.js
JoshuaVulcan: IMHO this should just be removed entirely. Resolutions in style conflicts should be fixed upstream by the implementor.

src/hooks/useMapLayers/index.js
JoshuaVulcan: @gaboDev this is a good point -- while I think it is generally useful to give a "default" config plus overrides per-layer, the interfaces by which you can do this are pretty unclear and not easy to reason through. Can you improve code legibility to better-document and more plainly state what the configuration objects accept and which takes precedence?

src/hooks/useMapLayers/index.js
JoshuaVulcan: At least it describes the shape of the expected configuration, which would otherwise only be intuited by reading the whole function thoroughly. 🤔

src/hooks/useMapLayers/index.js
JoshuaVulcan: Don't disable the linter for that line, it exists for a reason 😅

JoshuaVulcan (changes requested):
@gaboDev there's some good feedback from @luixlive to address -- especially open questions around the efficiency of iterating long lists to support small changes on individual members of those lists.

Additionally, looks like there's some funk in the processing/rendering of the tracks themselves; namely, if I reduce my track length to a temporal period which should have 0 track points, some time-of-day track data remains drawn on the map. See here

time-of-day-segments-not-cleaning-up

src/hooks/useMapLayers/index.js
gaboDev: At the end, calculating layers for segment pairs was the only place where the default config param was being used, so it was just matter of inject that global options to each layer (which was kind of already happened) so global default configs for a batch of layers was not really necessary, I think the implementator should be in charge of provide a proper layer config object per batch so We left out of the hook this responsibility

src/hooks/useMapLayers/index.js
gaboDev: The add layer effect is way more cleaner now, this destructuring was removed from it

src/hooks/useMapLayers/index.js
gaboDev: Removed!

src/hooks/useMapLayers/index.js
gaboDev: I removed both effect since where practically just increasing performance with no specific purpose

src/hooks/useMapSources/index.js
gaboDev: Removed already, sorry missed that in the last iteration

gaboDev:
> Additionally, looks like there's some funk in the processing/rendering of the tracks themselves; namely, if I reduce my track length to a temporal period which should have 0 track points, some time-of-day track data remains drawn on the map. See here
> 
> time-of-day-segments-not-cleaning-up

This bug was related with old layers refs being still in the map while not being present in the layer config batch, I just add some logic to remove those layers when necessary

@JoshuaVulcan

JoshuaVulcan (approved):
@gaboDev approved to not block, but we have an ongoing Slack conversation on the value of maintaining both the full track line + two-point segments, vs just holding the two-point segments in memory.

PR #1244, opened 2025-02-28
ERA-11239: track styles: patrol tracks

luixlive:
### What does this PR do?
Re-do the `PatrolTrackLegend` component so it has the look and feel like the new `SubjectTrackLegend`.

### How does it look
Desktop:
image

Mobile:
image

### Relevant link(s)
* [ERA-11239](https://allenai.atlassian.net/browse/ERA-11239)
* [Env](https://era-11239.dev.pamdas.org/)

1- Start by reviewing the new component `src/TrackLegend/index.js`. It's code is around 80% of what `SubjectTrackLegend` used to be. I just removed the subject internal details so new abstraction is not tied to subjects, and we can also use it to render patrol tracks.
2- After understanding the code of `TrackLegend`, reviewing `PatrolTrackLegend` and `SubjectTrackLegend` will be trivial.
3- The "last" important thing to review is the refactoring and optimization of patrols related selectors in `src/selectors/patrols.js`.
4- The rest of the files are translations and selector naming updates.

[ERA-11239]: https://allenai.atlassian.net/browse/ERA-11239?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Map/index.js
```diff
@@ -369,12 +369,6 @@ const Map = ({
     );
   }, [dispatch]);
 
-  const onPatrolTrackLegendClose = useCallback(() => {
```
luixlive, line 372: This dispatch is now handled directly by `PatrolTrackLegend`.

src/PatrolTrackLegend/index.js
```diff
@@ -1,110 +1,65 @@
-import React, { memo, useCallback } from 'react';
```
luixlive, line 1: Implementation of the new `TrackLegend` component for patrols:
image

src/PatrolTrackLegend/index.js
```diff
@@ -1,110 +1,65 @@
...
-      displayTitle={displayTitle}
-      iconId={iconId}
-      onRemovePatrolClick={(event) => dispatch(updatePatrolTrackState({
-        pinned: trackState.pinned.filter((value) => value !== event.target.value),
-        visible: trackState.visible.filter((value) => value !== event.target.value),
-      }))}
-      patrolData={patrolData}
-    />}
-  /> : null;
+  const patrolTrackState = useSelector((state) => state.view.patrolTrackState);
```
luixlive, line 19: Extremely simple component, we just need to select the `patrolTrackState` and `patrolsWithTrackData` values from Redux to calculate the description and the items array (patrols being tracked).

src/SubjectTrackLegend/index.js
```diff
@@ -1,125 +1,65 @@
-import React, { useCallback, useEffect, useMemo, useState } from 'react';
```
luixlive, line 1: There isn't a noticeable change here, but instead of implementing the legend directly here, now we implement the new `TrackLegend` component passing down the subjects:
image

src/SubjectTrackLegend/index.js
```diff
@@ -1,125 +1,65 @@
...
 
-const SubjectTrackLegend = ({ subjectTracksCount }) => {
+const SubjectTrackLegend = () => {
   const dispatch = useDispatch();
   const { t } = useTranslation('tracks', { keyPrefix: 'subjectTrackLegend' });
 
-  const timeOfDayTrackingEnabled = useFeatureFlag(FEATURE_FLAG_LABELS.TIME_OF_DAY_TRACKING);
-
-  const isTimeOfDayColoringActive = useSelector((state) => state.view.trackSettings.isTimeOfDayColoringActive);
   const subjectStore = useSelector((state) => state.data.subjectStore);
```
luixlive, line 24: The implementation is very similar to the `PatrolTrackLegend`. Here we select `subjectStore`, `subjectTrackState`, `subjectTracksTrimmedToTrackTimeEnvelope` and `trackTimeEnvelope` to build the description and items (tracked subjects).

src/TrackLegend/TracksList/index.js
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
```
luixlive, line 1: This is not really a new component, it's just the abstraction of what used to be `SubjectTrackList` so now it's agnostic of the items that are listed. They could either be subjects or patrols.

src/TrackLegend/index.js
```diff
@@ -0,0 +1,204 @@
+import React, { useCallback, useEffect, useState } from 'react';
```
luixlive, line 1: This is the new abstraction that renders track legends. It needs to be implemented and provided with the items, description, callbacks and the flags to render or not certain buttons. `SubjectTrackLegend` and `PatrolTrackLegend` use this component to render their legends, they provide subjects and patrols as items respectevely.

The code is very similar to what used to be `SubjectTrackLegend`, but I removed all the logic related to subjects so now the "items" are injected as dependencies (props) and the component stays agnostic.

src/ducks/patrols.js
```diff
@@ -262,45 +261,6 @@ export const uploadPatrolFile = (event_id, file, onUploadProgress = (event) => c
...
 };
 
-export const INITIAL_PATROLS_STATE = {
-  count: null,
-  next: null,
-  previous: null,
-  results: [],
-};
-
-const patrolsReducer = (state = INITIAL_PATROLS_STATE, action) => {
```
luixlive, line 272: This reducer wasn't needed anymore. We can rely entirely on the `patrolStore`. And honestly, it was old nasty code, we were basically storing an Axios response object in our store and barely using it (nothing we couldn't achieve using the `patrolStore` directly).

src/selectors/patrols.js
```diff
@@ -1,155 +1,143 @@
-import uniq from 'lodash/uniq';
```
luixlive, line 1: Total refactor and optimization of our patrol selectors.

src/utils/patrols.js
```diff
@@ -228,15 +228,6 @@ export const actualEndTimeForPatrol = (patrol) => {
     : null;
 };
 
-export const getLeaderForPatrol = (patrol, subjectStore) => {
```
luixlive, line 231: Unused code.

src/selectors/tracks/index.js
```diff
@@ -64,68 +62,45 @@ export const selectHeatmapSubjectTracksTrimmedToTrackTimeEnvelope = createSelect
   )
 );
 
-const selectPatrolTracksLeaderIds = createSelector(
```
luixlive, line 67: This selector is no longer needed since the logic to show / hide the subjects tracks when a patrol is being tracked and the subject is the leader of the patrol was a little bit cleaned (lots could be done to improve this flow still, it's buggy, but this is a start).

src/TrackLegend/index.js
```diff
@@ -0,0 +1,204 @@
...
+  const onCollapseMenu = () => setExpandedMenu(null);
+
+  const onExpandMenu = (menu) => {
+    if (!expandedMenu) {
+      // If no menu is currently expanded, we just expand the requested one.
+      setExpandedMenu(menu);
+    } else {
+      // If there is a menu expanded, we first collapse it and then expand the new one.
+      onCollapseMenu();
+      setTimeout(() => setExpandedMenu(menu), BOOTSTRAP_DEFAULTS.COLLAPSE_TRANSITION_TIME);
```
gaboDev, line 61: Would be worth it to clear this timeout on unmount?

src/TrackLegend/index.js
```diff
@@ -0,0 +1,204 @@
...
+    }
+  };
+
+  const onActivateTimeOfDayColoring = () => {
+    // When activating the time of day coloring, we also expand its menu.
+    dispatch(setIsTimeOfDayColoringActive(true));
+    onExpandMenu(MENUS.TIME_OF_DAY_SETTINGS);
+  };
+
+  const onDectivateTimeOfDayColoring = useCallback(() => {
```
gaboDev: little type in here, it should be `onDeactivateTimeOfDayColoring`

src/TrackLegend/index.js
```diff
@@ -0,0 +1,204 @@
...
+    dispatch(setIsTimeOfDayColoringActive(false));
+
+    // When deactivating the time of day coloring, we collapse its menu if it was expanded.
+    if (expandedMenu === MENUS.TIME_OF_DAY_SETTINGS) {
```
gaboDev: Several ifs statements are repeating the same comparisons of the const at the beginning of the component `isTimeOfDaySettingsExpanded` `isTrackSettingsExpanded` `isTracksListExpanded` Why are we using those here?

src/SubjectTrackLegend/index.js
```diff
@@ -1,125 +1,65 @@
...
+  }, [subjectTracksTrimmedToTrackTimeEnvelope, t, trackTimeEnvelope.from, trackTimeEnvelope.until]);
 
-  const onActivateTimeOfDayColoring = () => {
-    // When activating the time of day coloring, we also expand its menu.
-    dispatch(setIsTimeOfDayColoringActive(true));
-    onExpandMenu(MENUS.TIME_OF_DAY_SETTINGS);
-  };
+  // Build the items array with the description, icon, id and title of each tracked subject.
+  const items = useMemo(() => subjectTracksTrimmedToTrackTimeEnvelope.map((subjectTracks) => {
+    const id = subjectTracks.track.features[0].properties.id;
```
gaboDev: Here seems like a good place to use:
`const [firstFeature] = subjectTracks.track.features;`
instead of repeating `subjectTracks.track.features[0]` just to make the code more readable

gaboDev: Also, is it possible to `subjectTracks.track.features[0]` to does not exist? If yes, We should add some default values or validations

gaboDev (changes requested):
Just left a couple of minor comments, the only reason I'm not approving is that I don't see test coverage for `TrackLegend` implementation in `PatrolTrackLegend` and `SubjectTrackLegend` is there a reason for that?

src/TrackLegend/index.js
luixlive: Actually, it's impossible to unmount the component without first finishing the timeout. If you go to the last few lines of this component, it is wrapped by `DelayedUnmount` by the same amount of time of this timeout: `BOOTSTRAP_DEFAULTS.COLLAPSE_TRANSITION_TIME`. This means that even if the component is about to be unmounted, it will at least wait for this time to elapse, making the cleaning task unnecessary 😄

src/TrackLegend/index.js
luixlive: Duh, didn't think of it 🤓  Updating them!

src/SubjectTrackLegend/index.js
luixlive: On the first comment: agreed, I'll update it.
On the second one: it shouldn't, that data comes from a selector that is already running validations and filtering just the valid data. If we ever hit that issue, I wouldn't say the solution is adding a check here but fixing the selector.

luixlive:
> Just left a couple of minor comments, the only reason I'm not approving is that I don't see test coverage for `TrackLegend` implementation in `PatrolTrackLegend` and `SubjectTrackLegend` is there a reason for that?

@gaboDev  Nice observation! Actually I forgot to add a comment regarding this myself. I decided not to add tests to those. I tried, but the necessary mocked data in the store for the selectors to build the right scenarios is so complex and big, that the tests would be extremely fragile and hard to comprehend. I know this is not orthodox, we should try to add tests anyway, but for real, the amount of data was insane, and in every scenario it changes considerably. In my opinion, components shouldn't rely on these super complex data selection. Our selectors are doing too much. I would say that the right fix is to update our reducers to actually do some pre-processing before storing the data and avoid having to do all that processing in our selectors. That would optimize time and memory, while also simplifying the data mocking in the tests. But that was definitely out of the scope. This PR was big enough.

gaboDev (approved):

PR #1252, opened 2025-03-07
ERA-11057: track styles: track color expression

gaboDev:
### What does this PR do?
- Fixes track clearing bug, it fixes the clean-up process of map layers at unmounting time 

### Relevant link(s)
[ERA-11057](https://allenai.atlassian.net/browse/ERA-11057)

- The bug was reproducible by:

1. turning on time-of-day styles
2. adjusting the track length days
3. clearing tracks

The app was crashing due map sources being deleted while there were map layers actively using that source, meaning the cleanup process of map layers was not working properly when clicking the 'Clear tracks' button of the track settings legend. 

[ERA-11057]: https://allenai.atlassian.net/browse/ERA-11057?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/hooks/useMapLayers/index.js
+++ b/src/hooks/useMapLayers/index.js
@@ -24,7 +24,6 @@ const useMapLayers = (layerConfigsBatch = []) => {
           map.removeLayer(id);
         }
       });
-      layerIdsRef.current = existingLayers.slice();
 
       // Add new layers
       layerConfigsBatch.forEach(layerConfig => {
--- a/src/TracksLayer/track.js
+++ b/src/TracksLayer/track.js
@@ -1,4 +1,4 @@
-import { memo, useContext, useEffect, useMemo, useRef } from 'react';
+import { memo, useContext, useMemo } from 'react';
 import PropTypes from 'prop-types';
 
 import { LAYER_IDS, MAP_ICON_SCALE } from '../constants';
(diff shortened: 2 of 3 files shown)


src/hooks/useMapLayers/index.js
```diff
@@ -24,7 +24,6 @@ const useMapLayers = (layerConfigsBatch = []) => {
           map.removeLayer(id);
         }
       });
-      layerIdsRef.current = existingLayers.slice();
```
JoshuaVulcan, line 27: @gaboDev could you clarify how this fixes the issue, and what the original intent of assigning this ref's value here was? I want to make sure we're not losing any business logic by simply removing this line, rather than moving it.

gaboDev: The idea was to keep layerIdsRef in sync with the existing layers, but at the end, keeping them in sync was causing old layers to keep active and not being removed simply because they were not listed on the ref

gaboDev: So, it is no necessary to keep the ref updated because its responsibility it is to clean up all unused layer even the one not listed in the config batches but existing ones in the map

JoshuaVulcan (approved):

PR #1253, opened 2025-03-07
ERA-11236: Support EFB schemas in ER > Location Field

luixlive:
Fist of all: sorry for the enormous PR. In order to achieve a fully a11y a keyboard-navigatable location widget, a few of our shared components needed to be refactored, and then I had to update the places where those were implemented too. The good news: the benefits of these new components will be noticed in a few different places of the app 😄 

### What does this PR do?
Re-do our location picker component to be more accessible and use the new version in the `SchemaForm` component.

### How does it look
image

### Relevant link(s)
* [ERA-11236](https://allenai.atlassian.net/browse/ERA-11236)
* [Env](https://era-11236.dev.pamdas.org)

I'd recommend to pull the branch and start in the new `LocationPicker`. From there you can navigate to its child components: `MenuPopover`, `GpsInput`, `PickMapLocationButton` and `GetUserLocationButton`. Then,`GpsInput` also has a child that was refactored: `GpsFormatToggle`. After reviewing the code and testing those components, you will be over two thirds of the review.

Then you can go to `SchemaForm/fields/Location/index.js`, where the new `LocationPicker` is used to render the location fields of the new schemas. I added a new mocked schema "Example for locations" that will be helpful for testing the different configurations of the input.

[ERA-11236]: https://allenai.atlassian.net/browse/ERA-11236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/CursorGpsDisplay/index.js
```diff
@@ -1,4 +1,5 @@
 import React, { useCallback, useContext, useEffect, useRef, useState } from 'react';
+import Button from 'react-bootstrap/Button';
```
luixlive: Updates to match the new version of our GPS iput in the cursor display:
image

src/DatePicker/index.js
```diff
@@ -84,12 +84,14 @@ const DatePicker = ({
   const onYearInputKeyDown = (event) => {
     switch (event.key) {
     case 'ArrowDown':
-      event.preventDefault();
-
-      // Decrease the year when the user presses the down arrow.
-      if (isValidYearInput(year) && parseInt(year) > 0) {
-        const yearMinusOne = (parseInt(year) - 1).toString().padStart(4, '0');
-        onYearChange(yearMinusOne);
+      if (!readOnly) {
```
luixlive, line 87: Trivial changes done to our `DatePicker` so it behaves correctly when it is set in mode `readOnly`. We were passing down the prop to the inputs but the keyboard events like increasing or decreasing the year, month and day with up and down arrows were still functional.

src/GetUserLocationButton/index.js
```diff
@@ -0,0 +1,75 @@
+import React, { forwardRef, memo, useState } from 'react';
```
luixlive, line 1: Total refactor of our button to retrieve the user location:
image

src/GpsFormatToggle/index.js
```diff
@@ -12,55 +12,79 @@ import styles from './styles.module.scss';
 
 const gpsFormatTracker = trackEventFactory(GPS_FORMAT_CATEGORY);
 
-const GpsFormatChoice = ({ currentFormat, format }) => {
+const GpsFormatToggle = ({
```
luixlive: Refactor of the component to be fully accessible and keyboard-navigatable. Now the choices are a semantically correct list of radio inputs that are styled to look as tabs. You can focus the fieldset with tab and change the selection with the arrows:
image

src/GpsInput/index.js
```diff
@@ -17,107 +13,111 @@ import GpsFormatToggle from '../GpsFormatToggle';
 
 import styles from './styles.module.scss';
 
-const GpsInput = ({ buttonContent, lngLat, onButtonClick, onValidChange, tooltip, ...restProps }) => {
+const GpsInput = ({
```
luixlive, line 16: Another big refactor to add a11y and keyboard navigation:
image

src/LocationPicker/MenuPopover/index.js
```diff
@@ -0,0 +1,153 @@
+import React, { forwardRef, useEffect, useRef } from 'react';
```
luixlive, line 1: The menu of our new location picker. Fully accessible and easy to navigate. I has a focus trap so pressing tab doesn't take you away from the scope of the menu:
image

src/LocationPicker/index.js
```diff
@@ -0,0 +1,130 @@
+import React, { forwardRef, memo, useImperativeHandle, useRef, useState } from 'react';
```
luixlive, line 1: Our new accessible and navigable `LocationPicker`, with a conditionally rendered copy button and a jump to location button. It consists of a beautifully styled, wrapped and handled `<input type="text" /> ` so it is semantically correct, recognized by forms and by screen readers, while also being very simple to use with the mouse and keyboard:
image

src/MapMarkerDropper/index.js
```diff
@@ -3,17 +3,20 @@ import throttle from 'lodash/throttle';
 import PropTypes from 'prop-types';
 
 import { addMapImage } from '../utils/map';
+import { MAP_INTERACTION_CATEGORY, trackEventFactory } from '../utils/analytics';
```
luixlive, line 6: Update the map marker dropper to use the new `PickMapLocationButton`:
image

src/PatrolDetailView/PlanSection/index.js
```diff
@@ -20,7 +20,7 @@ import { setMapLocationSelectionPatrol } from '../../ducks/map-ui';
 import { useMatchMedia } from '../../hooks';
 
 import DatePicker, { EMPTY_DATE_VALUE } from '../../DatePicker';
-import LocationSelectorInput from '../../EditableItem/LocationSelectorInput';
```
luixlive, line 23: Updates to the patrol's detail view to support the new `LocationPicker`:
image

src/PatrolDetailView/index.js
```diff
@@ -259,12 +259,10 @@ const PatrolDetailView = () => {
   }, [patrolForm]);
 
   const onPatrolEndLocationChange = useCallback((endLocation) => {
-    const updatedEndLocation = !!endLocation ? { latitude: endLocation[1], longitude: endLocation[0] } : null;
```
luixlive, line 262: The new `LocationPicker` doesn't follow the Mapbox uncommon standard of defining locations like `[lng, lat]`. It's simple and less verbose, but it doesn't match the location data type as it is expected by any type of form. The new picker follows a more standard an readable format: `{ latitude: '', longitude: '' }`.

src/PickMapLocationButton/index.js
```diff
@@ -0,0 +1,104 @@
+import React, { forwardRef, memo, useContext, useEffect, useState } from 'react';
```
luixlive: Another totally re-done component. It's previous version wasn't functional by itsel! It required the implementator to hide the sidebar and modals so the map is available to the user. This new version is totally independent, verbose and leaves the styling of the button to the parent so it's very flexible:
image

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/SortableList/Item/FormModal/index.js
```diff
@@ -36,7 +39,7 @@ const FormModal = ({
       aria-labelledby="formModal-title"
       backdrop={false}
       centered
-      className={isNestedModal ? styles.noBackground : ''}
+      className={`${styles.formModal} ${isNestedModal ? styles.noBackground : ''} ${!canShowModals ? styles.hide : ''}`}
```
luixlive, line 42: Added a new class to hide the modal if the store value `state.view.modals.canShowModals` is `false`. Otherwise, location inputs inside a collection modal would be useless, we need to hide the modal so the user can choose a location then show the modal again.

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/SortableList/Item/utils/index.js
```diff
@@ -54,14 +54,14 @@ export const getHumanizedValue = (field, value, defaultHumanizedValue, language,
     return isValid(parsedDate) ? format(parsedDate, formatStr) : defaultHumanizedValue;
 
   case FORM_ELEMENT_TYPES.LOCATION:
-    return `${value.latitude}°, ${value.longitude}°`;
+    return calcGpsDisplayString(value.latitude, value.longitude, gpsFormat);
```
luixlive, line 57: Instead of always showing the title of a location field as latitude and longitude, now we use the system GPS format set by the user to follow their preferences.
image

src/ReportManager/DetailsSection/SchemaForm/fields/Location/index.js
```diff
@@ -0,0 +1,44 @@
+import React, { memo } from 'react';
```
luixlive, line 1: The new `Location` field of the `SchemaForm` which is basically a trivial implementator of the `LocationPicker` adding all the right `aria` attributes, a label, a description and an error message:
image

src/ReportManager/DetailsSection/index.js
```diff
@@ -34,7 +34,7 @@ import {
 import AreaSelectorInput from './AreaSelectorInput';
 import DatePicker, { EMPTY_DATE_VALUE } from '../../DatePicker';
 import GeometryPreview from './AreaSelectorInput/GeometryPreview';
-import LocationSelectorInput from '../../EditableItem/LocationSelectorInput';
+import LocationPicker from '../../LocationPicker';
```
luixlive, line 37: Update the our report details section to use the new `LocationPicker`:
image

src/TextCopyBtn/index.js
```diff
@@ -1,14 +1,23 @@
...
 import { ReactComponent as ClipboardIcon } from '../common/images/icons/clipboard-icon.svg';
 
-import { showToast } from '../utils/toast';
-
 import styles from './styles.module.scss';
 
-const TextCopyBtn = ({ className, getText, icon, label, permitPropagation, successMessage, text, ...restProps }) => {
+const TOAST_AUTOCLOSE_TIME = 2000;
+
+const TextCopyBtn = ({
```
luixlive, line 11: And the last refactored component to be accessible and navigable:
image

src/TimePicker/OptionsPopover/index.js
```diff
@@ -131,6 +131,8 @@ const OptionsPopover = ({
   const onItemSelection = (time) => {
     onChange(time);
     onClose();
+
+    optionsPopoverButtonRef.current.focus();
```
luixlive, line 135: Minor improvements to our time picker, like only focusing the input when closing the popover if it was closed by pressing escape or selecting an item (not by clicking somewhere else). And adding a nice animation for focusing the selected item of the list.

src/TimePicker/index.js
```diff
@@ -124,7 +124,7 @@ const TimePicker = ({
   const onHourInputBlur = () => {
     // If the hour input is blurred and the user left a single digit we autofill the first one with a zero, unless we
     // moved the focus programatically after the user typed a valid hour.
-    if (shouldAutofillHourOnBlurRef.current && shouldCompleteFirstHourDigitWithZero(hour)) {
+    if (!readOnly && shouldAutofillHourOnBlurRef.current && shouldCompleteFirstHourDigitWithZero(hour)) {
```
luixlive, line 127: Just like with the `DatePicker`, added the missing conditions to forbid the keyboard functionality if the field is `readOnly`.

tiffanynwong:
It's beautiful 🥲. No comments nice work!!

src/LocationPicker/index.js
```diff
@@ -0,0 +1,128 @@
...
+        disabled={disabled}
+        text={displayValue}
+        title={t('textCopyButtonLabel')}
+      />}
+
+      <button
+        aria-label={t('jumpToLocationButtonLabel')}
+        className={styles.jumpToLocationButton}
+        disabled={!value || disabled}
+        onClick={() => jumpToLocation([value.longitude, value.latitude])}
```
gaboDev, line 100: There are several places where the `value` object is being used to access `longitude` and `latitude`, I see that prop is a required one, but in some places where `LocationPicker` is being implemented that value prop is set to null in some cases, so in this particular line, should We avoid to call `jumpToLocation` fn with a `[undefined, undefined]` param ?

I think We should add a validation here

src/GpsInput/index.js
```diff
@@ -17,107 +13,111 @@ import GpsFormatToggle from '../GpsFormatToggle';
...
       setIsValid(true);
-      setLastKnownValidValue(inputValue);
-      onValidChange(inputValue);
+      onChange(null);
     } else {
       try {
-        const locationObject = calcActualGpsPositionForRawText(inputValue, gpsFormat);
+        const locationObject = calcActualGpsPositionForRawText(event.target.value, gpsFormat);
         const isLocationValid = validateLngLat(locationObject.longitude, locationObject.latitude);
         if (!isLocationValid) {
```
gaboDev: We can simplify this if-else statement by assigning `isLocationValid` to `setIsValid` and then just add a simple if to call `onChange` conditionally

src/PickMapLocationButton/index.js
```diff
@@ -0,0 +1,104 @@
...
+  const map = useContext(MapContext);
+
+  const [isPickingMapLocation, setIsPickingMapLocation] = useState(false);
+
+  const onButtonClick = () => {
+    // Update the store so the application is aware that user is picking a location in the map, it hides the modals and
+    // the sidebar.
+    dispatch(setIsPickingLocation(true));
+    dispatch(setModalVisibilityState(false));
+    dispatch(hideSideBar());
```
gaboDev: This is just an idea, at some point it would be great to have a dispatch in charge of putting the app in that 'mode' of hiding several UI elements to let the user use the map freely

I took a look if there are another places where these dispatches are being called at the same time like here, but since this is the only place I decided to add this just as an idea

gaboDev: Just noticed, all those dispatches are being constantly re-invoked together, Wouldn't be a good idea to fit them into a toggle function? something like `toggleUIElement(true)` ?

gaboDev (changes requested):
Added a couple of comments, nothing serious, from there I'm ready to approve 🫡

src/LocationPicker/index.js
luixlive: Nice observation! Yes, `value` is required because this must be a controlled component. However we must have a way to tell: the user hasn't set a value yet. Since the value is an object `{ latitude: number, longitude: number }` we don't set the empty value with an empty string, so I decided to use `null`. That means that, as you say, we should be careful to consider the `null` scenario to avoid exceptions when trying to read the `latitude` and `longitude` properties. But don't worry, that is handled! If you take a look at all the places where we read those properties in this file, we first check that the value is defined:
```
const displayValue = value ? calcGpsDisplayString(value.latitude, value.longitude, gpsFormat) : '';

...

{value && <TextCopyBtn ... />}

...

<input
  ...
  value={value ? `${value.latitude},${value.longitude}` : ''}
/>
```

And in this specific line, we shouldn't worry about jumping to the location with undefined values, because that happens on the `onClick` event. But if the value is `null` it won't be clickable because it will be disabled! 
```
disabled={!value || disabled}
```

Same applies to the `MenuPopover` that receives the `value` as a prop too. So we should be good 😄

src/PickMapLocationButton/index.js
luixlive: Nice suggestion, I decided to create the following method:
```
  const setAppToShowMapMode = useCallback((showMapMode = true) => {
    dispatch(setIsPickingLocation(showMapMode));
    dispatch(setModalVisibilityState(!showMapMode));
    dispatch(showMapMode ? hideSideBar() : showSideBar());
  }, [dispatch]);
```

gaboDev (approved):

PR #1255, opened 2025-03-13
Hotfix: client-side memory issues

JoshuaVulcan:
### What does this PR do?
- Fixes a memory leak issue with unrevoked object URLs being used in the generation of cleaned SVGs for map icon presentation.
- Also adds a bit of memoization for additional performance gains.

I had a call with some admins from Lewa and SWT this morning about performance issues, and noticed our map image object URLs being recalculated at a high frequency. While looking into memoizing those calculated values for a performance gain, I noticed the object URLs weren't being revoked...in this context, this causes a memory leak. And if called that frequently, it could really pile up. So I fixed that, then added the memoization, and some test coverage to match. 

I'd like to hotfix this into the current release as a first test to see if it helps out the folks at Lewa.


src/TrackLegend/TimeOfDaySettings/TimeZoneSelect/index.test.js
```diff
@@ -60,17 +60,18 @@ describe('TrackLegend - TimeOfDaySettings - TimeZoneSelect', () => {
...
 
-    expect(setTimeOfDayTimeZone).not.toHaveBeenCalled();
+  //   expect(setTimeOfDayTimeZone).not.toHaveBeenCalled();
 
-    userEvent.click(screen.getAllByRole('option')[100]);
+  //   userEvent.click(screen.getAllByRole('option')[100]);
 
-    expect(setTimeOfDayTimeZone).toHaveBeenCalledTimes(1);
-    // This test may break if someday the IANA standard updates.
-    expect(setTimeOfDayTimeZone).toHaveBeenCalledWith('America/Guadeloupe');
```
luixlive: I think we can keep this test uncommented and having partial coverage if all we do is just remove this line. We would at least make sure that `setTimeOfDayTimeZone` is called.

src/utils/img.js
```diff
@@ -11,42 +12,83 @@ const imgNeedsHostAppended = url => {
   return true;
 };
 
-export const imgElFromSrc = (src, width = 30, height = null) => new Promise((resolve, reject) => {
-  let img = new Image();
+const imageCache = new Map();
```
luixlive, line 15: Having a global variable as a cache store has a smell to me. Maybe it's time to re-work the approach of these methods? Couldn't we have a hook with a state variable with the "cache" that passes down through context the `imgElFromSrc` method? Something like that sounds like the React approach, easier to memoize, debug, etc...

src/utils/img.js
```diff
@@ -11,42 +12,83 @@ const imgNeedsHostAppended = url => {
...
+
+const generateImageCacheKey = (src, width, height) => {
+  const w = width === null ? 'null' : (width === undefined ? 'undefined' : width);
+  const h = height === null ? 'null' : (height === undefined ? 'undefined' : height);
+  return `${src}:${w}:${h}`;
+};
+
+export const imgElFromSrc = (src, baseUnit = null) => {
```
luixlive, line 23: Could we use more readable method and variable names? `imgElFromSrc`, `w`, and `h` names aren't very descriptive. And specifically with the method, just by reading the name (even if we use complete words) doesn't really tell what it does, maybe adding a verb to the name could help.

src/utils/img.js
```diff
@@ -11,42 +12,83 @@ const imgNeedsHostAppended = url => {
...
+  if (!src) {
+    return Promise.reject('no src provided');
+  }
+
+  const cacheKey = generateImageCacheKey(src, baseUnit, null);
+
+  if (imageCache.has(cacheKey)) {
+    return imageCache.get(cacheKey);
```
luixlive, line 31: Maybe I don't have a full understanding of what is going on, but it feels like this is a patch and not the solution to the issue. Why do we need to cache the images? Shouldn't we instead ask ourselves why is this method being called so many times being asked for the same resources?

src/utils/img.js
```diff
@@ -11,42 +12,83 @@ const imgNeedsHostAppended = url => {
...
-      const smallest = Math.min(naturalHeight, naturalWidth) || baseUnit;
-      const widthIsLarger = largest === naturalWidth;
-      const aspectRatio = smallest / largest;
-      if (widthIsLarger) {
+  const imagePromise = new Promise((resolve, reject) => {
+    const cleanupAndResolve = () => {
+      if (baseUnit && img.naturalWidth && img.naturalHeight) {
+        const widthIsLarger = img.naturalWidth > img.naturalHeight;
+
+        if (widthIsLarger || !widthIsLarger) {
```
luixlive: What is this? 🤯  This is logically equivalent to `if (true)` hahaha. This condition will always be true, it doesn't make sense to have it wrapped.

src/utils/img.js
```diff
@@ -57,5 +99,7 @@ export const calcUrlForImage = imagePath => {
     return imagePath;
   }
   const appendString = !!DAS_HOST ? `${DAS_HOST}/` : '';
-  return `${appendString}${imagePath}`.replace(/^http:\/\//i, 'https://').replace('.org//', '.org/');
+  const final = `${appendString}${imagePath}`.replace(/^http:\/\//i, 'https://').replace('.org//', '.org/');
```
luixlive: Personally, I don't think this variable is helping a lot. `final` doesn't tell me anything about the content of this string. If the intention was to make clearer what is being returned, we could be a bit more verbose.

src/utils/img.test.js
```diff
@@ -0,0 +1,214 @@
+import { imgElFromSrc, calcImgIdFromUrlForMapImages, calcUrlForImage, registerActiveURL } from './img';
```
luixlive, line 1: Thanks for the tests!!

luixlive (changes requested):
I see some logical errors and have a suggestion to instead of using utility methods with global variables as cache mechanism to go for a more React-y approach. Maybe a context provider handling the caching and retrieval of images. Also, I read this code and kind of get a sense of what is doing but it feels very technical and not verbose enough to fully tell me what is doing. Maybe we could use this PR for a bit of a refactoring on the variable names, conditions, adding comments, etc... To make it cleaner and more readable 👍

src/utils/img.js
JoshuaVulcan: It's less efficient to do it that way, but we can certainly React-ify it more if you would like. I would prefer just using a factory function to create a nice closure to wrap the cache in.

src/utils/img.js
JoshuaVulcan: The render cycle for the map clusters builds the image strings each time, which is normal since that's how the functional programming nature of React componentry works, and then has traditionally relied on browser caching mechanisms to not over-request images which have already been fetched.

Since we're creating object URLs for some images, this causes a memory leak; the object URLs are being created each time, and are held in memory indefinitely.  Having our own cache here prevents `createObjectURL` from being called more than needed, and thus patches the leak.

src/utils/img.js
JoshuaVulcan: Just a WIP typo. This isn't part of the latest change set.

src/utils/img.js
luixlive: Well, it shouldn't be less efficient if it's properly memoized. Maybe we are talking about a couple renders and a couple milliseconds, but React re-renders hundreds of times and the cache mechanism will make those milliseconds be negligible. I think the React approach is a better fit four our React application, but I also understand it's out of the scope of a hotfix, just wanted to share the idea.

src/utils/img.js
JoshuaVulcan: Got it

JoshuaVulcan:
> I see some logical errors and have a suggestion to instead of using utility methods with global variables as cache mechanism to go for a more React-y approach. Maybe a context provider handling the caching and retrieval of images. Also, I read this code and kind of get a sense of what is doing but it feels very technical and not verbose enough to fully tell me what is doing. Maybe we could use this PR for a bit of a refactoring on the variable names, conditions, adding comments, etc... To make it cleaner and more readable 👍

I agree with 100% of this -- can we do it outside of the scope of this hotfix? I'd like to apply this change in isolation to a couple of prod sites to test if it resolves a memory leak I noticed recently

src/utils/img.js
JoshuaVulcan: Totally! Maybe in a separate technical debt task? This is a hotfix branch and these are by and large existing function names

src/utils/img.js
```diff
@@ -11,42 +12,81 @@ const imgNeedsHostAppended = url => {
...
+  const h = height === null ? 'null' : (height === undefined ? 'undefined' : height);
+  return `${src}:${w}:${h}`;
+};
+
+export const imgElFromSrc = (src, baseUnit = null) => {
+  if (!src) {
+    return Promise.reject('no src provided');
+  }
+
+  const cacheKey = generateImageCacheKey(src, baseUnit, null);
```
Chouffe, line 28: It looks like `baseUnit` is not always the width of the image:

```js
if (baseUnit && img.naturalWidth && img.naturalHeight) {
        const widthIsLarger = img.naturalWidth > img.naturalHeight;

        const aspectRatio = widthIsLarger ?
          img.naturalHeight / img.naturalWidth :
          img.naturalWidth / img.naturalHeight;

        if (widthIsLarger) {
          img.width = baseUnit;
          img.height = Math.round(baseUnit * aspectRatio);
        } else {
          img.height = baseUnit;
          img.width = Math.round(baseUnit * aspectRatio);
        }
      } else if (baseUnit) {
        img.width = baseUnit;
        img.height = baseUnit * aspectRatio;
      } else {
        img.height = baseUnit;
        img.width = baseUnit * aspectRatio;
      }
``` 

Does it make sense to update the `generateImageCacheKey` function to use `baseUnit` instead of width, height parameters?

src/utils/img.js
```diff
@@ -11,42 +12,81 @@ const imgNeedsHostAppended = url => {
...
-});
+
+
+
+      resolve(img);
+    };
+
+    img.addEventListener('load', cleanupAndResolve, { once: true });
+
+    img.onerror = (e) => {
```
Chouffe, line 66: Nice cleanup here

src/utils/img.js
Chouffe: How often and how much data can/should be held by this cache? Would a regular js map be good enough for a cache? Or should we use like a bounded in-memory cache to avoid another memory leak - the cache can potentially grow indefinitely here. 

Also, we could reach for persistent solutions like LocalStorage/SessionStorage.

Chouffe (commented):
Overall looks like a nice and pragmatic fix to the memory leak that is described in the PR description.

The main question I have is whether a regular js Map is enough as an in-memory cache because it is unbounded. If we know that only a small number of Images can be generated overall, it should be fine, but if bazillions of images are potentially generated we probably want to use something more robust than a Map to avoid another memory-leak.

JoshuaVulcan:
> Overall looks like a nice and pragmatic fix to the memory leak that is described in the PR description.
> 
> The main question I have is whether a regular js Map is enough as an in-memory cache because it is unbounded. If we know that only a small number of Images can be generated overall, it should be fine, but if bazillions of images are potentially generated we probably want to use something more robust than a Map to avoid another memory-leak.

@Chouffe great question! The reason I used to a `Map` to begin with was actually how nice it was to work with when busting cache items beyond a certain limit (I think I'd originally set it to be like 100 or 200 in my first pass at the code).  But in real-world terms, there's a maximum of a few hundred possible for an ER site; as such, I removed the bounding/lifecycle management stuff and just kept it simple.

PR #1256, opened 2025-03-15
bugfix:missing required prop for track rendering

JoshuaVulcan:
### What does this PR do?
- Hotfixes a missing required property which got merged into the develop branch. 

### Relevant link(s)
* https://allenai.atlassian.net/browse/ERA-11295


.env.development
```diff
@@ -1,5 +1,4 @@
-REACT_APP_DAS_HOST=https://root.dev.pamdas.org
-REACT_APP_GA_TRACKING_ID=UA-128569083-12
```
Chouffe, line 2: This env variable is not needed anymore?

.env.production
```diff
@@ -1,6 +1,5 @@
 GENERATE_SOURCEMAP=false
 REACT_APP_DAS_HOST=''
 REACT_APP_ROUTE_PREFIX=/
-REACT_APP_GA_TRACKING_ID=UA-128569083-1
```
Chouffe, line 4: Same question here.

src/TracksLayer/index.js
```diff
@@ -35,12 +35,13 @@ const TracksLayer = ({ onPointClick, showTimepoints }) => {
 
   return subjectTracksWithPatrolTrackShownFlag.length > 0
     ? subjectTracksWithPatrolTrackShownFlag.map((subjectTracks) => <TrackLayer
-        key={`track-layer-${subjectTracks.track.features[0].properties.id}`}
-        linePaint={{ 'line-opacity': subjectTracks.patrolTrackShown ? 0.4 : 1 }}
-        onPointClick={onTimepointClick}
-        showTimepoints={showTimepoints}
-        trackData={subjectTracks}
-      />)
+      id={subjectTracks.track.features[0].properties.id}
```
Chouffe, line 38: This is the culprit of the bug I suppose. 
Looks good to me overall!

Not knowing the exact business logic, a question comes to mind when reviewing this:
Is it possible to have an empty `features` list? In that case, how do we handle it?

Chouffe (approved):
Overall LGTM! Nice fix. I left a question regarding default behavior and error handling.

.env.development
JoshuaVulcan: Correct! We transitioned to GA4, so the Universal Analytics ID is no longer needed.

.env.production
JoshuaVulcan: Same answer 😆

src/TracksLayer/index.js
JoshuaVulcan: Good question! By the time the selector chain gets to this point, the data has been pretty thoroughly cleaned and prepared. I don't think it's possible to hit this point having a `track` property but without a linestring feature inside that track.

PR #1268, opened 2025-03-31
ERA-10349: Support EFB schemas in ER > Location Field [3]

luixlive:
### What does this PR do?
Fix bug when pressing escape while choosing location

### DIFF
--- a/src/SideBar/index.js
+++ b/src/SideBar/index.js
@@ -65,6 +65,7 @@ const SideBar = () => {
 
   const sideBarRef = useRef();
 
+  const isPickingLocation = useSelector((state) => state.view.mapLocationSelection.isPickingLocation);
   const sideBar = useSelector((state) => state.view.sideBar);
 
   const [showEventsBadge, setShowEventsBadge] = useState(false);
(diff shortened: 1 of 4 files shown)


src/SideBar/index.js
```diff
@@ -147,20 +148,21 @@ const SideBar = () => {
     }
   }, [sidebarOpen, currentTab, socket, isReportDetailsViewActive]);
 
+  // NOTE: This is getting unmaintainable. Is it really a good practice to use escape like a navigation key?
```
luixlive, line 151: This feature has been causing many issues since we added it. This `useEffect` adds a key down listener in the `SideBar` to navigate back if `Escape` is pressed. I wonder if it's worth it to keep adding patches everytime we find a new unwanted side effect this is adding. Pressing `Escape` to navigate is not common. I know that key is used to close views, and given our navigation is related to a SideBar it kind of makes sense to expect the side bar to close when pressing it. But our SideBar is not just a presentational element to close, its tabs and its items are connected to the router, causing the mentioned side effects like the one I'm fixing it when pressing escape while picking a location in the map.

JoshuaVulcan (approved):

PR #1273, opened 2025-04-03
[ERA-11281] Message Badge Display Consistency

JoshuaVulcan:
### What does this PR do?
- Adds two layout props to the message badge layer. One prevents the badge from disappearing if another icon overlaps it, and the other prevents the message badge from making a different layer's icons disappear if they overlap. In short, to ignore all collision detection and be consistently-shown.

### How does it look
- N/A

### Relevant link(s)
* https://allenai.atlassian.net/browse/ERA-11281
* No hosted env; this is tiny, can we just test in dev?

### DIFF
--- a/src/MessageBadgeLayer/index.js
+++ b/src/MessageBadgeLayer/index.js
@@ -24,11 +24,13 @@ const calcMapMessages = (messages = [], subjectFeatureCollection) => {
   const subjectFeaturesWithUnreadMessages =
     subjectFeatureCollection.features
       .map(feature =>
-        ({ feature, messages: messages
-          .filter(msg =>
-            extractSubjectFromMessage(msg)?.id === feature.properties.id
-          )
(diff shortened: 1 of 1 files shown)


src/MessageBadgeLayer/index.js
```diff
@@ -48,6 +50,8 @@ const LAYER_ID = `${SOURCE_ID}_LAYER`;
 
 const messageBadgeLayout = {
   'icon-anchor': 'bottom-left',
+  'icon-ignore-placement': true,
+  'icon-allow-overlap': true,
   'icon-image': 'message-badge',
   'icon-offset': [2, -0.35],
   'icon-size': 0.5,
```
JoshuaVulcan, line 57: Lines 53-54 are the only substantive changes. The rest is just auto-formatting.

luixlive (approved):

PR #1274, opened 2025-04-03
ERA-11154: Do not add a collection item when pressing "cancel" upon first creation

gaboDev:
### What does this PR do?
- Adds logic to the Collection component to remove new empty collection items when user cancels an addition. 

### Relevant link(s)
* Tracking tickets: [ERA-11154](https://allenai.atlassian.net/browse/ERA-11154)
* [Hosted demo environments](https://era-11154.dev.pamdas.org)

[ERA-11154]: https://allenai.atlassian.net/browse/ERA-11154?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/ReportManager/DetailsSection/SchemaForm/fields/Collection/index.js
```diff
@@ -108,15 +112,22 @@ const Collection = ({
...
   };
 
-  const setIsItemFormModalOpen = (itemIndex) => (isItemFormModalOpen) => setItems([
-    ...items.slice(0, itemIndex),
-    { ...items[itemIndex], isFormModalOpen: isItemFormModalOpen },
-    ...items.slice(itemIndex + 1),
-  ]);
+  const setIsItemFormModalOpen = (itemIndex) => (isItemFormModalOpen) => {
+    setItems((currentItems) => {
+      const itemToUpdate = currentItems[itemIndex];
```
gaboDev, line 117: This is the fix for the bug I found, once the modal was cancel after leaving it empty (not adding info and not saving it) that empty item was deleted at `onItemDelete` fn, but here `items` array  was still having that empty object causing the modal not appear for the next added items

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/SortableList/Item/FormModal/index.test.js
```diff
@@ -44,6 +44,7 @@ describe('ReportManager - DetailsSection - SchemaForm - fields - Collection - So
         renderField={renderField}
         rightColumn={[]}
         title="Item 3"
+        isDeletable
```
luixlive: We should add a test to make sure that the trash icon renders and doesn't render depending on this new prop

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/SortableList/Item/index.js
```diff
@@ -29,6 +29,7 @@ const Item = ({
   isDragging = false,
   isDragOverlay = false,
   isFormModalOpen = false,
+  isItemRecentAdded = false,
```
luixlive: Typo: `wasItemRecentlyAdded`

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/SortableList/Item/FormModal/index.js
```diff
@@ -15,6 +15,7 @@ const FormModal = ({
   focusLocationMarker,
   formData,
   isOpen,
+  isDeletable,
```
luixlive: I think the name of this prop is incorrect. The item is still deletable, we just don't want the trash button. We could keep the name of the variable of the parent `shouldDeleteOnCancel` or create a custom prop like `hideDeleteButton`. Both are more explicit on the intent.

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/SortableList/Item/index.js
```diff
@@ -94,9 +91,23 @@ const Item = ({
     onChange({ ...formData, [fieldId]: value }, updatedErrors);
   };
 
-  const onDeleteItem = () => {
-    onDelete();
+  const onDeleteItem = (wasRecentAddedItemCanceled = false) => {
```
luixlive: I think passing this argument around is unnecessary. We can just read the ref in here: 
```
const onDeleteItem = () => {
  onDelete(shouldDeleteOnCancelRef.current); // <-- We know for sure that it was deleted through cancel
  setIsFormModalOpen(false);
};
```

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/SortableList/index.js
```diff
@@ -142,24 +142,25 @@ const SortableList = ({
     >
       <SortableContext items={items.map((item) => item.id)} strategy={verticalListSortingStrategy}>
         {items.map((item, index) => <SortableItem
-          blurLocationMarker={blurLocationMarker}
```
luixlive: 😮 I think you'r editor did something weird here haha, a big indentation just out of nowhere! It makes even confusing to review if you did any change in the props of the componente. I think you didn't, which makes even weirder why this file is being modified at all.

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/index.js
```diff
@@ -66,7 +66,7 @@ const Collection = ({
     );
   };
 
-  const onItemDelete = (itemIndex) => () => {
+  const onItemDelete = (itemIndex) => (wasRecentAddedItemCanceled) => {
```
luixlive: Here we have the `recent` -> `recently` typo again.
But instead I'd suggest to rename this variable with a name of its intention and not the context. Saying that a recently added item being canceled trigger the method is ok, but saying: I want you to delete the item AND decrease the last added item id, is awesome. This way the argument is reusable in other contexts and its clearer what the intention of the argument is.
I'd suggest calling it something like `(decreaseLastAddedItemId = false) =>`

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/index.js
luixlive: Ok, but I think it would be better to do the check outside of the setter function so we don't trigger a render right?
```
// We don't trigger setItems unless we really need to change its value
const setIsItemFormModalOpen = (itemIndex) => (isItemFormModalOpen) => items[itemIndex] && setItems([
    ...items.slice(0, itemIndex),
    { ...items[itemIndex], isFormModalOpen: isItemFormModalOpen },
    ...items.slice(itemIndex + 1),
  ]);
```

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/index.js
```diff
@@ -108,15 +112,22 @@ const Collection = ({
...
+          { ...itemToUpdate, isFormModalOpen: isItemFormModalOpen },
+          ...currentItems.slice(itemIndex + 1)
+        ];
+    });
+  };
 
   const setIsItemFormPreviewOpen = (itemIndex) => (isItemFormPreviewOpen) => setItems([
     ...items.slice(0, itemIndex),
-    { ...items[itemIndex], isFormPreviewOpen: isItemFormPreviewOpen },
+    { ...items[itemIndex], isFormPreviewOpen: isItemFormPreviewOpen, isRecentAdded: false },
```
luixlive: Hmmm why do we have this here 🤔 The form preview has nothing to do with the modal. It seems like we are mixing the logic here. The recently added here should only be set as `true` when adding a new item, and its internal ref will be set to false when clicking Done. Why opening the form preview is interfering with that logic? This has a smell, or maybe it's just unnecessary code.

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/index.js
```diff
@@ -130,7 +141,15 @@ const Collection = ({
...
     lastAddedItemIdRef.current += 1;
     onFieldChange(id, [...value, {}], updatedError);
-    setItems([...items, { id: lastAddedItemIdRef.current, isFormModalOpen: true, isFormPreviewOpen: false }]);
+    setItems([
+      ...items,
+      {
+        id: lastAddedItemIdRef.current,
+        isFormModalOpen: true,
+        isFormPreviewOpen: false,
+        isRecentAdded: true
```
luixlive: `wasRecentlyAdded`

luixlive (changes requested):
The approach is almost there, I added a few suggestions.

Besides of what I already mentioned, I don't see any tests 😬  We should add tests for every `.js` or `.jsx` file modified here, unless there is a very good reason not to or if the changes don't imply any logic like addin a new CSS class.

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/index.js
gaboDev: I got your point here, but the idea is to work with the latest version of `items` state variable (which basically was the root cause of the issue) so making a comparing outside the callback leave us in the same place.

I think returning the same object without modifications from the `setState` callback wont trigger a re-render, right?

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/index.js
gaboDev: My bad, I should have placed this code inside `setIsItemFormModalOpen` as a reset for the flag, also remember the ref is destroyed since it is inside `<Item/>` which is rendered based on `items` array, meaning `item.wasItemRecentlyAdded` is used as an initializer of the ref

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/SortableList/Item/FormModal/index.js
```diff
@@ -96,17 +97,18 @@ const FormModal = ({
...
-        aria-label={t('deleteButton', { itemTitle: title } )}
-        className={styles.deleteButton}
-        onClick={onDeleteItem}
-        title={t('deleteButton', { itemTitle: title } )}
-        type="button"
-      >
-        <TrashCanIcon />
-      </button>
-
+    <Modal.Footer className={`${styles.footer} ${!hideDeleteButton ? styles.alignEvenly : styles.alignRight}`}>
```
luixlive: nit: good conditional practices, put first the truthy resolution `hideDeleteButton ? styles.alignRight : styles.alignEvenly` makes it easier for our brains to rapidly scan the code branches over each condition.

luixlive (approved):
Nice! A little non-blocker, but ✅

JoshuaVulcan (approved):
Works well for me.  Some minor conflicts in the `.env` to resolve!

luixlive:
I won't block the PR, but just noticed that it still lack a few tests in some of the modified files. I just see a single test file modified, but several JS files were modified with new code that should be tested @gaboDev

PR #1276, opened 2025-04-08
feat:connecting trawl lines between linked buoys

JoshuaVulcan:
### What does this PR do?
- Sketches some basic trawl lines on the map for Buoy devices.

### How does it look
Screenshot 2025-04-08 at 11 12 59 AM

### Relevant link(s)
* 🤷 

- The logic is a little funny; we can't rely on the coordinates provided by the `devices` array in a subject's `additional` properties, so instead we match the `device_id`s listed to their subject names and then use the last known position for that subject. @andrejiron21 @doug-nimblegravity should definitely add story work to keep that `devices` data better refreshed to avoid weird data sync issues in the future.

### DIFF
--- a/src/BuoyTrawlLineLayer/index.js
+++ b/src/BuoyTrawlLineLayer/index.js
@@ -1,8 +1,8 @@
 import React from 'react';
 import { useSelector } from 'react-redux';
+import { lineString } from '@turf/turf';
 
 import { getMapSubjectFeatureCollectionWithVirtualPositioning } from '../selectors/subjects';
-import { lineString } from '@turf/turf';
 import useMapLayers from '../hooks/useMapLayers';
 import useMapSources from '../hooks/useMapSources';
 
--- a/src/BuoyTrawlLineLayer/index.js
+++ b/src/BuoyTrawlLineLayer/index.js
@@ -1,0 +1,70 @@
+import React from 'react';
+import { useSelector } from 'react-redux';
+
+import { getMapSubjectFeatureCollectionWithVirtualPositioning } from '../selectors/subjects';
+import { lineString } from '@turf/turf';
+import useMapLayers from '../hooks/useMapLayers';
+import useMapSources from '../hooks/useMapSources';
+
+const lineLayout = {
+  'line-join': 'round',
+  'line-cap': 'round',
+};
+
+const linePaint = {
+  'line-color': 'black',
+  'line-opacity': 0.7,
+  'line-gap-width': 1,
+  'line-width': 2,
+};
+
+const subjectIsBuoyLineEligible = (subjectFeature = {}, _index, allSubjects = []) => {
+  const subject = subjectFeature.properties;
+
+  const is_buoy = subject.subject_subtype === 'ropeless_buoy_device';
+  if (!is_buoy) return false;
+
+  const devices = subject.additional?.devices ?? [];
+  const is_line = devices.length > 1;
+
+  if (!is_line) return false;
+
+  const line_contains_valid_subjects = devices.every(({ device_id }) => allSubjects.find(({ properties }) => properties.name === device_id));
+
+  return line_contains_valid_subjects;
+};
+
+const createTrawlLineGeoJSON = (buoySubjectFeatures) => {
+  return buoySubjectFeatures.reduce((accumulator, { properties }) => {
+    const coordinates =
+      properties.additional.devices.map(({ device_id }) =>
+        buoySubjectFeatures.find(({ properties }) =>
+          properties.name === device_id)?.geometry?.coordinates ?? []
+      );
+
+    accumulator.features.push(lineString(coordinates));
+    return accumulator;
+
+  }, { type: 'FeatureCollection', features: [] });
+};
+
+const BuoyLineLayer = (_props) => {
+  const mapSubjects = useSelector(getMapSubjectFeatureCollectionWithVirtualPositioning);
+
+  const buoySubjects = mapSubjects.features.filter(subjectIsBuoyLineEligible);
+  const trawlLineGeoJSON = createTrawlLineGeoJSON(buoySubjects);
+
+  useMapSources([{ id: 'trawl-lines-source', data: trawlLineGeoJSON }]);
+  useMapLayers([{
+    id: 'trawl-lines-layer',
+    type: 'line',
+    sourceId: 'trawl-lines-source',
+    layout: lineLayout,
+    paint: linePaint,
+  }]);
+
+  return null;
+
+};
+
+export default BuoyLineLayer;
--- a/src/BuoyTrawlLineLayer/index.js
+++ b/src/BuoyTrawlLineLayer/index.js
@@ -39,7 +39,7 @@ const subjectIsBuoyLineEligible = (subjectFeature = {}, _index, allSubjects = []
 
 const createTrawlLineGeoJSON = (buoySubjectFeatures) => {
   return buoySubjectFeatures.reduce((accumulator, { properties }) => {
-    const coordinates =
+    const coordinates = // build the coordinates from each subject's location
       properties.additional.devices.map(({ device_id }) =>
         buoySubjectFeatures.find(({ properties }) =>
           properties.name === device_id)?.geometry?.coordinates ?? []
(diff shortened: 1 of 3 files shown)


src/BuoyTrawlLineLayer/index.js
```diff
@@ -0,0 +1,70 @@
...
+  'line-color': 'black',
+  'line-opacity': 0.7,
+  'line-gap-width': 1,
+  'line-width': 2,
+};
+
+const subjectIsBuoyLineEligible = (subjectFeature = {}, _index, allSubjects = []) => {
+  const subject = subjectFeature.properties;
+
+  const is_buoy = subject.subject_subtype === 'ropeless_buoy_device';
```
luixlive: This string should be a constant. Maybe we won't need it anywhere else so at least putting it at the top of the file like: `BUOY_SUBJECT_SUBTYPE` is enough.

Second question: why snake case here (and in all other variables of this method)? It's not our usual approach 🤔

src/BuoyTrawlLineLayer/index.js
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import { useSelector } from 'react-redux';
+
+import { getMapSubjectFeatureCollectionWithVirtualPositioning } from '../selectors/subjects';
+import { lineString } from '@turf/turf';
```
luixlive: nit: this import should go in the first block.

src/BuoyTrawlLineLayer/index.js
```diff
@@ -0,0 +1,70 @@
...
+
+  const is_buoy = subject.subject_subtype === 'ropeless_buoy_device';
+  if (!is_buoy) return false;
+
+  const devices = subject.additional?.devices ?? [];
+  const is_line = devices.length > 1;
+
+  if (!is_line) return false;
+
+  const line_contains_valid_subjects = devices.every(({ device_id }) => allSubjects.find(({ properties }) => properties.name === device_id));
```
luixlive: This variable name sounds odd, it almost sounds like a boolean for a condition: `if (lineContainsValidSubjects) ...` but it is not, it's just a list of subjects contained by the line. I guess it needs to be reworded, or maybe just return the one-liner since it's kind of easy to read.

src/BuoyTrawlLineLayer/index.js
```diff
@@ -0,0 +1,70 @@
...
+
+  return line_contains_valid_subjects;
+};
+
+const createTrawlLineGeoJSON = (buoySubjectFeatures) => {
+  return buoySubjectFeatures.reduce((accumulator, { properties }) => {
```
luixlive, line 41: Honestly, I've stared at this function for a while haha. I feel like we could make it a bit easier to understand maybe by changing some variable names or not extracting properties from objects like: `({ properties }) => properties.additional...` -> `(buoySubjectFeature) => buoySubjectFeature.properties.additional` etc...

src/BuoyTrawlLineLayer/index.js
```diff
@@ -0,0 +1,70 @@
...
+  }, { type: 'FeatureCollection', features: [] });
+};
+
+const BuoyLineLayer = (_props) => {
+  const mapSubjects = useSelector(getMapSubjectFeatureCollectionWithVirtualPositioning);
+
+  const buoySubjects = mapSubjects.features.filter(subjectIsBuoyLineEligible);
+  const trawlLineGeoJSON = createTrawlLineGeoJSON(buoySubjects);
+
+  useMapSources([{ id: 'trawl-lines-source', data: trawlLineGeoJSON }]);
```
luixlive: We have constants for sources and layers that we should use to keep our code consistent and the ids accessible. Also, I don't know if you tested the properly addition, update and removal of these sources and layers (seeing Chromes console for warnings and errors), but as I mentioned recently, these hooks don't follow Mapbox sources and layers flows correctly from beginning to end. That's why I think it would be better to stop using them and instead use the map API directly.

luixlive (changes requested):

src/BuoyTrawlLineLayer/index.js
JoshuaVulcan: The snake case and non-constantified strings are just me doing this as a quickie/prototype.

src/BuoyTrawlLineLayer/index.js
JoshuaVulcan: This is a boolean for a condition -- it's just using `Array.prototype.every` to check device presence in the subject list.

luixlive: Ohh right, I just understood the every and where this is being used 👍

src/BuoyTrawlLineLayer/index.js
JoshuaVulcan: Summary version:

We _can_ say that the devices listed for a buoy feature's `additional.devices` should be the names of the subjects comprising that trawl.

We _can't_ say that the coordinates included in the additional data `devices` array data is up-to-date. 

So the eligibility-checker function first ensures a trawl line is completely valid and comprised of available subjects for the user.

Then  `createTrawlLineGeoJSON` maps those device names back to their subject's last known position and builds out the linestring coordinates from there.

src/BuoyTrawlLineLayer/index.js
JoshuaVulcan: I did test absence/presence/filtering etc, and timesliding, and it all works as expected.

I kept the names out of our constants file because I'd like this to be as self-isolated as possible. I will however make the strings constants within this file.

src/BuoyTrawlLineLayer/index.js
luixlive: That makes sense. Maybe a couple tweaks or a brief comment could make that more obvious in the code for future reference.

JoshuaVulcan: Added a descriptive inline comment

src/BuoyTrawlLineLayer/index.js
luixlive: Oh with tested I meant manually. In unit testing we usually use mocks for the map and other methods and the real flow of addition, interaction and removal (or not removal, causing memory leaks) of sources and layers are obscured by all the mocking. Anyway, we agreed this is out of the scope.

About the layers and sources being isolated, why's that 🤔 ? I know this is a feature for the specific scenario of buoy, but I don't get why we don't want to treat it as part of our app if now it is. Just like stationary subjects, or patrols, which are features not everyone use. I feel I'm missing some information or something haha.

src/BuoyTrawlLineLayer/index.js
```diff
@@ -0,0 +1,73 @@
...
+          properties.name === device_id)?.geometry?.coordinates ?? []
+      );
+
+    accumulator.features.push(lineString(coordinates));
+    return accumulator;
+
+  }, { type: 'FeatureCollection', features: [] });
+};
+
+const BuoyLineLayer = (_props) => {
```
luixlive, line 54: nit: why the `(_props) => {` instead of simply `() => {`?

luixlive (approved):
The source + layer is very simple and the processing of the source data is clear and well tested 👍

src/BuoyTrawlLineLayer/index.js
JoshuaVulcan: > Oh with tested I meant manually.

Me too :-) 

"Features not everyone uses" can be considered a spectrum -- with this Buoy work being at one pretty extreme end of the spectrum, "features designed for a specific site and fringe use-cases".  It's _so_ specifically for a special sub-project, that I'd rather not pollute the broader app with random `BUOY_` variables and references.

src/BuoyTrawlLineLayer/index.js
JoshuaVulcan: just a matter of preference!

src/BuoyTrawlLineLayer/index.js
JoshuaVulcan: fixed!

src/BuoyTrawlLineLayer/index.js
luixlive: I mean, it's a default linter rule (`no-unused-vars`) that we explicitly skip by prefixing the `_`. It feels weird to add it intentionally for nothing 🤔

PR #1277, opened 2025-04-08
Carrying over 2.113 hotfixes into 2.114

JoshuaVulcan:
### What does this PR do?
- 2.114 was cut before 2.113's hotfixes were pulled down into develop. We should make sure the hotfixes aren't absent from our release.

### How does it look
- N/A

### Relevant link(s)
- N/A


luixlive (commented):
Is the message badge layer a hotfix from the previous release too? I don't think so right? Is it a new hotfix?

JoshuaVulcan:
> Is the message badge layer a hotfix from the previous release too? I don't think so right? Is it a new hotfix?

It's coming over from 113, it was pushed as a hotfix.

PR #1286, opened 2025-04-24
Merging release into dev

JoshuaVulcan:


luixlive (approved):
We should work on removing that feature flag since this is now on by default in prod.

PR #1293, opened 2025-05-12
Stable generated track colors for subjects without `stroke` property specified

JoshuaVulcan:
### What does this PR do?
- Uses a Mapbox style expression to assign a random but stable track color for subjects who do not have explicitly assigned subject tracks. 
- The generated color value is seeded based on the UUID of the subject
- The color range is within boundaries that prevent them from being too dark or too light to be seen on low-contrast backgrounds as best possible.

### Relevant link(s)
* 🤷 
* https://joshua-random-track-colors.dev.pamdas.org/
* 

- It's a one-liner that can benefit a lot of users. Let's test in dev.

- This is a long-desired feature which was recently yet again requested. This is a near-no-code-change way to achieve it.

### DIFF
--- a/src/TracksLayer/track.js
+++ b/src/TracksLayer/track.js
@@ -13,11 +13,21 @@ import useMapLayers from '../hooks/useMapLayers';
 
 const { TRACKS_LINES, SUBJECT_SYMBOLS } = LAYER_IDS;
 
+const STABLE_RANDOM_TRACK_COLOR_BASED_ON_ID = [
+  'rgb',
+  ['random', 64, 224, ['concat', ['get', 'id'], '-r']],
+  ['random', 64, 224, ['concat', ['get', 'id'], '-g']],
+  ['random', 64, 224, ['concat', ['get', 'id'], '-b']]
+];
+
 const TRACK_LAYER_LINE_PAINT = {
   'line-color': [
     'case',
-    ['has', 'stroke'], ['get', 'stroke'],
-    'orange',
+    ['all',
+      ['has', 'stroke'],
+      ['!=', ['get', 'stroke'], '']
+    ], ['to-color', ['get', 'stroke']],
+    STABLE_RANDOM_TRACK_COLOR_BASED_ON_ID,
   ],
   'line-width': ['step', ['zoom'], 1, 8, ['get', 'stroke-width']],
 };


JoshuaVulcan:
@tiffanynwong do we need to do any user testing or feedback gathering here? I think it'll generally be a welcome change, but I'm wondering if people shouldn't have to opt-in so they're not surprised by new colors for their subjects.

tiffanynwong:
We actually just heard this request from a user today, who said they'd need this in a month to evaluate if ER is the right tool to visualize their vessel data (through an integration w/ Gundi, that's done). I think we've heard this request enough that it's fine to just release it, especially since it won't affect subjects that have been assigned a color. 

How many unique colors do you have? Chris from CSIRO Australia said anything from 12-100 colors would be a big help.

src/TracksLayer/track.js
```diff
@@ -13,11 +13,21 @@ import useMapLayers from '../hooks/useMapLayers';
 
 const { TRACKS_LINES, SUBJECT_SYMBOLS } = LAYER_IDS;
 
+const STABLE_RANDOM_TRACK_COLOR_BASED_ON_ID = [
```
luixlive, line 16: Why not to choose a default color instead of assigning them randomly? I may be missing something but I think that as a user I would be a bit annoyed, or even I would think that something in my data changed, if every time I open a subject's track I get a different color. Also the name `stable random track color` is kind of confusing 🤔

luixlive (approved):
I left a question / suggestion, but nothing to block the release ✅

src/TracksLayer/track.js
JoshuaVulcan: >  every time I open a subject's track I get a different color

@luixlive  That's not the behavior of this.

It is **random** because the color is generated by a hashing function, and **stable** because the hash is seeded by the user ID -- meaning each subject UUID will generate a unique color, and the _same hash output will be generated on each render for that subject_.  In other words, subjects will be uniquely colored, and that color will stay the same until overridden in the admin. 

This guarantees each subject has a unique track color (i.e. is discernible from one-another on a map, instead of all being the default orange color and hard to disentangle during visual analysis)

luixlive: Duh, missed the uuid-seeded part. Ok, I guess it makes sense to have unique colors instead of a single default for all of them 👍  Thanks for the clarification!

luixlive: The name will still cause me itches later but I understand there's no easy way to name that haha.

JoshuaVulcan: SAME 💯  I'm all ears for a better suggestion

PR #1297, opened 2025-05-19
chore:add ecoscope downloader link to web UI

JoshuaVulcan:
### What does this PR do?
- Adds a link

### How does it look
- N/A

### Relevant link(s)
* https://allenai.atlassian.net/browse/ERA-11323


JoshuaVulcan:
Screenshot 2025-05-19 at 2 57 43 PM

public/locales/es/menu-drawer.json
```diff
@@ -36,6 +36,7 @@
       "content": "Eventos exportados solo incluiran aquellos que coincidan con los criterios del filtro actualmente aplicado en la pestaña de Reportes.",
       "title": "Eventos de campo"
     },
+    "ecoscopeDownloaderLink": "Descargador de Ecoscopio",
```
luixlive: Do we want to translate the word "Ecoscope"? I would say no, since its the name of the application.

public/locales/ne-NP/menu-drawer.json
```diff
@@ -36,6 +36,7 @@
       "content": "घटना ट्याबमा हाल सेट गरिएका फिल्टर मापदण्डसँग मेल खाने घटनाहरु मात्रै निर्यात घटनामा समावेश हुनेछन् ।",
       "title": "फिल्ड रिपोर्ट"
     },
+    "ecoscopeDownloaderLink": "इकोस्कोप डाउनलोडर",
```
luixlive, line 39: Same, seems like "Ecoscope" was translated

src/GlobalMenuDrawer/index.js
```diff
@@ -269,6 +270,9 @@ const GlobalMenuDrawer = () => {
         <h6>{t('exportsHeader')}</h6>
 
         {modals.map((modal) => <button key={modal.title} onClick={() => onModalClick(modal)}>{modal.title}</button>)}
+        <a className={styles.ecoscopeDownloaderLink} target="_blank" rel="noreferrer" href='https://ecoscope.io/en/latest/ecoscope_gui.html#downloads' aria-label="Ecoscope Downloader">
```
luixlive: I see the translations but I don't think we are using them. The aria label and the text seem to be hardcoded.

luixlive (changes requested):
I don't think you created an env to test this. It's ok since it's very simple, but from your screenshot I feel like the icon is too big, and I'm not sure if it's me but I see it misaligned with the text 🤔

PR #1298, opened 2025-05-21
FE for ERA-11542

gaboDev:


luixlive:
@gaboDev seems like this ticket was closed so can we close this PR?

PR #1305, opened 2025-06-26
feat: subject group track toggle button. test: coverage for map-ui du…

JoshuaVulcan:
…cks and subjectgrouplist selectors

### What does this PR do?
- Allows user to toggle tracks to a visible, pinned, or off state for entire subject groups and subgroups at a time.

### Relevant link(s)
* https://allenai.atlassian.net/browse/ERA-11525
* https://era-11525.dev.pamdas.org/

- The SubjectGroupList files are the primary changes. 
- 
- Easy quick win for usability. Please note I backed off a lot of refactors and foundational changes in favor of keeping this light-touch and consistent with the current code.


src/SideBar/styles.module.scss
```diff
@@ -489,10 +490,16 @@
 }
 
 .trigger {
+  align-items: center;
```
luixlive, line 495: Given that you are touching these styles, and that the scope is related, it would be nice to add hover and focus states to these buttons (at the very least the tracks and heatmap, since there are other things missing that too in the list). It took me several blind tabs to get to the jump to location button, which is the only thing with a focus state here (and honestly, the blue outline is not beautiful haha):
image

src/SubjectGroupList/Content.js
```diff
@@ -84,13 +102,39 @@ const ContentComponent = (props) => {
...
+
+      await fetchTracksIfNecessary(unloadedSubjectTrackIDs);
+
+      setTrackLoadingState(false);
+    }
+
+    if (groupTrackingDataState.track === TRACKING_CONTROL_STATES.FULLY_PINNED) {
+      return hideSubjectTracks(...subjectIDsWithTrackingData);  // turn off all;
+    }
+    if (
```
luixlive, line 129: The way this if is defined is kind of funky, took me a few reads to actually understand it haha. A bit more verbose but IMHO clearer: 
```JavaScript
if (
  groupTrackingDataState.track === TRACKING_CONTROL_STATES.PARTIALLY_PINNED
  || groupTrackingDataState.track === TRACKING_CONTROL_STATES.FULLY_VISIBLE) {
```

You think it just as you read it: if the group state of the tracks is partially pinned or fully visible... Now, in my opinion, the best way to achieve the intention of this code-block in a very readable way:
```JavaScript
switch (groupTrackingDataState.track) {
case TRACKING_CONTROL_STATES.FULLY_PINNED:
  // If all tracks are pinned, hide them all.
  return hideSubjectTracks(...subjectIDsWithTrackingData);

case TRACKING_CONTROL_STATES.PARTIALLY_PINNED:
case TRACKING_CONTROL_STATES. FULLY_VISIBLE:
  // If all tracks are visible, or a few are pinned, pin them all.
  return pinSubjectTracks(...subjectIDsWithTrackingData);

default:
  // Otherwise, just show them all.
  return showSubjectTracks(...subjectIDsWithTrackingData);
}
```

And in a different thought, related to this comment, I wonder if we could have a better name for `groupTrackingDataState`. Reading `groupTrackindDataState.track` is a bit confusing 🤔

src/SubjectGroupList/Content.js
```diff
@@ -27,30 +28,47 @@ const mapLayerTracker = trackEventFactory(MAP_LAYERS_CATEGORY);
...
     {listLevel > 0 && <h6>{itemTitle}</h6>}
-    {showHeatmapControl && <HeatmapToggleButton className={listStyles.toggleButton} loading={loadingTracks}
-      heatmapVisible={groupIsFullyHeatmapped}
-      heatmapPartiallyVisible={groupIsPartiallyHeatmapped}
-      onButtonClick={onGroupHeatmapToggle} showLabel={false} />}
+    {showTrackingControls && <>
+      <TrackToggleButton
+        loading={loadingTracks}
+        onClick={onTrackButtonClick}
+        className={`${(partiallyPinned || partiallyVisible) ? listStyles.partialTrackButton : ''}`}
```
luixlive, line 58: I feel like the `partialTrackButton` class, which modifes the opacity, doesn't change the UI enough for a user to be able to tell: oh this group is visible partially:
image

Maybe adding a dashed border vs a full border helps? Or simply decreasing more the opacity.

src/SubjectGroupList/Content.js
```diff
@@ -146,7 +191,7 @@ const ContentComponent = (props) => {
   </Collapsible>;
 };
 
-const mapStateToProps = (state, ownProps) => subjectGroupHeatmapControlState(state, ownProps);
+const mapStateToProps = (state, ownProps) => subjectGroupTrackingControlsState(state, ownProps);
```
luixlive: Not blocker since it's out of scope, but it would be nice to move these old `connect` to `useDispatch` and `useSelector` 😄

src/SubjectGroupList/selectors.js
```diff
@@ -1,34 +1,100 @@
 import { createSelector } from 'reselect';
 import intersection from 'lodash/intersection';
 
-import { getHeatmapEligibleSubjectsFromGroups } from '../utils/subjects';
+import { getSubjectsWithViewableTrackingDataFromGroups } from '../utils/subjects';
+
+export const TRACKING_CONTROL_STATES = {
```
luixlive: We don't usually export constants from a selectors file. I think it would be better to define this in a `constants.js` and import it here.

src/SubjectGroupList/selectors.js
```diff
@@ -1,34 +1,100 @@
...
+    if (fullyPinned) trackState.track = TRACKING_CONTROL_STATES.FULLY_PINNED;
+    else if (partiallyPinned) trackState.track = TRACKING_CONTROL_STATES.PARTIALLY_PINNED;
+    else if (fullyVisible) trackState.track = TRACKING_CONTROL_STATES.FULLY_VISIBLE;
+    else if (partiallyVisible) trackState.track = TRACKING_CONTROL_STATES.PARTIALLY_VISIBLE;
+
+    return trackState;
+  },
+);
+
+export const subjectGroupTrackingControlsState = createSelector(
```
luixlive: This selector is not doing anything, it's composing other selectors but not really processing anything, just adding memoization overhead 🤔 The only "processing" it does is `!!eligibleSubjects.length` the other values are just grouped in an object and returned. Feels like we could simply select these three pieces directly from the component.

src/ducks/map-ui.js
```diff
@@ -119,6 +119,42 @@ export const toggleTrackState = (id) => (dispatch, getState) => {
 
 };
 
+export const pinSubjectTracks = (...subjectIDs) => (dispatch, getState) => {
```
luixlive: I see we are reading the reducer state in the action creator, which is kind of breaking the flux flow right? Why not to simply send the action and leave the reducer the responsibility of generate the new track state data? In theory, `getState` should be avoided as much as possible to not cause side effects between reducers, but sometimes it's really necessary. The funny part here is that we are using `getState` to read the data of the same reducer haha. We just need to send an action like `PIN_SUBJECT_TRACKS` and the reducer has everything it needs to calculate the new state in there 👍

luixlive (changes requested):
Nice testing coverage! The functionality works fine 💯 
image

Just provided some feedback on the Redux patterns (flux and selectors) and a couple non-blocking minor requests for UI and code style improvements 😄

src/SubjectGroupList/selectors.js
```diff
@@ -1,34 +1,100 @@
...
+    eligibleSubjects.length
+    && intersection(eligibleSubjects, visibleTracks).length === eligibleSubjects.length,
+);
+
+export const groupTracksPartiallyVisible = createSelector(
+  [visibleTrackIDs, groupTracksFullyVisible, visibleTrackingDataSubjectIDsForGroup],
+  (visibleTracks, tracksFullyVisible, eligibleSubjects) =>
+    eligibleSubjects.length > 0
+    && !tracksFullyVisible
+    && intersection(eligibleSubjects, visibleTracks).length,
```
luixlive, line 48: I think Copilot suppressed a good recommnedation. Here you are returning the length of the intersection. Which is truthy, and probably that is why it works, but it may cause issues later. We should return a boolean like in the other selectors. Maybe something like: 
```JavaScript
(visibleTracks, tracksFullyVisible, eligibleSubjects) =>
    eligibleSubjects.length > 0
    && !tracksFullyVisible
    && intersection(eligibleSubjects, visibleTracks).length > 0, // Make it a boolean
```

src/SubjectGroupList/Content.js
JoshuaVulcan: It would be lovely! I had to stop myself halfway through to keep the scope+size manageable.

src/SubjectGroupList/Content.js
JoshuaVulcan: This seems like a stylistic difference; IMHO the way I have it now is more legible and more extensible 😅 . Agreed the name isn't ideal

src/SubjectGroupList/Content.js
JoshuaVulcan: I'll decrease the opacity for now; let's take another look

src/ducks/map-ui.js
JoshuaVulcan: This is a thunk, an extremely common Redux pattern which is used widely throughout this codebase.
https://redux.js.org/usage/writing-logic-thunks

src/SubjectGroupList/selectors.js
JoshuaVulcan: I don't think the API contract of `intersection` is likely to shift away from returning an Array, nor is JS likely to change its Boolean representation of zero and non-zero values, but I'm happy to make this more explicit.

src/ducks/map-ui.js
luixlive: Oh yes I get that. What I mean is that the idea of thunks is to expand the capabilities of the normal action -> reducer flow. Like adding the possibility of running asynchronous actions in the middle (like network calls) or reading the state of another portion of the store (through getState). But in here we are using getState to read the same portion of the store the reducer already has. The reducer could simply receive the action type and action payload and calculate its new state by itself.

src/SubjectGroupList/selectors.js
luixlive: JS truthy and falsy values will remain the same, and given the way we use this value right now, this works. But from a future dev perspective, it would be odd to receive an array from a selector that is supposed to return a boolean.

src/SubjectGroupList/selectors.js
JoshuaVulcan: RIP to `subjectGroupTrackingControlsState`

docker-compose.yml
```diff
@@ -14,17 +14,17 @@ services:
...
-      context: .
-      dockerfile: Dockerfile.dev
-    volumes:
-      - /app/node_modules
-      - .:/app
-    # environment:
-    #   - CI=true
-    command: yarn test
-    restart: on-failure
+  # web_test:
```
luixlive, line 17: 💯 hahaha

src/SubjectGroupList/Content.js
```diff
@@ -24,39 +27,69 @@ const COLLAPSIBLE_LIST_DEFAULT_PROPS = {
...
+  const hiddenSubjectIDs = useSelector(({ data: { mapLayerFilter: { hiddenSubjectIDs } } }) => hiddenSubjectIDs);
 
   const { t } = useTranslation('layers', { keyPrefix: 'layerList' });
 
   const [loadingTracks, setTrackLoadingState] = useState(false);
   const [collapsibleShouldBeOpen, setCollapsibleOpenState] = useState(false);
 
+
+  /* 
+    addHeatmapSubjects, removeHeatmapSubjects, hideSubjectTracks, pinSubjectTracks, showSubjectTracks
```
luixlive: 👀

luixlive (approved):
Nice! I still see a lot of possible improvements around these old layer collapsible components and the way we handle the data. But I agree it's well beyond the scope.

src/SubjectGroupList/Content.js
JoshuaVulcan: gone now!

PR #1307, opened 2025-06-30
ERA-9754: adds data-testids for e2e subject tests

AlanCalvillo:
### What does this PR do?
- just adds some data-testid attributes that are used in https://github.com/PADAS/er-web-automation-playwright/pull/201

### Context
It was difficult to interact with the texts and buttons related to subjects since they were dynamic. Adding a data-testid stablize Playwright interactions.

### DIFF
--- a/src/SubjectPopup/index.js
+++ b/src/SubjectPopup/index.js
@@ -54,7 +54,7 @@ const SubjectPopup = ({ data }) => {
   return <>
     <div className={styles.header}>
       <div>
-        <div className={styles.defaultStatusProperty} data-testid='subject-popup-name'>
+        <div className={styles.defaultStatusProperty} data-testid="subject-popup-name">
           {properties.default_status_value && <>
             {properties.image && <img alt={t('subjectIconAlt', { name: properties.name })} src={properties.image} />}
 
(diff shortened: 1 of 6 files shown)


src/SubjectPopup/index.js
```diff
@@ -54,7 +54,7 @@ const SubjectPopup = ({ data }) => {
   return <>
     <div className={styles.header}>
       <div>
-        <div className={styles.defaultStatusProperty}>
+        <div className={styles.defaultStatusProperty} data-testid='subject-popup-name'>
```
gaboDev: I know We have a lot of places where there is mix of double quotes and single ones usages

I see that most of the time We use double quotes for fixed props like: `type="button"`
For simple strings like params or so: `t('defaultTitle')`
At least We can try to have that consistency 🥳

gaboDev (approved):
LGTM 🚀

src/SubjectPopup/index.js
AlanCalvillo: ah, that makes sense. i noticed the mix of usage so i didn't know which one to use. i better had to ask, though. i'll update them.

PR #1308, opened 2025-07-01
ERA-11695: Sort Map Layers Alphabetically and by Recent Updates

luixlive:
### What does this PR do?
Add new tabbed design to the Map Layers Sidebar panel and add new sorting buttons for Subjects.

### How does it look
image

### Relevant link(s)
* [ERA-11695](https://allenai.atlassian.net/browse/ERA-11695)
* [Env](https://era-11695.dev.pamdas.org)

**Notes**
In the file `src/SideBar/MapLayersTab/index.js` you will notice that the `Tab` components have the props `mountOnEnter` and `unmountOnExit`. This is not ideal, since it takes more resources to constantly mount and unmount the whole layer lists every time the user changes the tab. They are necessary though, because we use the `Collapse` component which needs its items to be in view when calculating its height in order to properly work. Since by typing in the search bar we trigger the opening of the collapses of the filtered layers, the ones that open when their tab is not selected get corrupted. This is the easiest solution (mount and unmount) so the `Collapse` mounts and unmounts only when in view and properly calculates its height. A better solution would be to redesign our lists and not use `Collapse`s anymore.

[ERA-11695]: https://allenai.atlassian.net/browse/ERA-11695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/FeatureLayerList/styles.module.scss
luixlive, line 1: You will notice that a bunch of files were "deleted" and others just got moved. In reality, all were just moved, but while doing so, I did refactors in the files that were related to the task I was working on, specifically the "Subjects" tab. But I also updated a few things in styles classes and other stuff. Some of them git marks them as deleted (and then added as new files) and in some of them it correctly flags them as moved to a new file folder.

src/SearchBar/index.js
```diff
@@ -29,7 +29,7 @@ const SearchBar = ({ className = '', onClear = null, value, ...otherProps }) =>
       {...otherProps}
     />
 
-    {onClear && <button
+    {onClear && value && <button
```
luixlive, line 32: I noticed that some translations have longer placeholders that may take the entire width of the input and leave a weird empty space for the clear button that won't even show up until we have a value, truncating the placeholder in vain. So I decided to only render the clear button once we have a value (which results in exactly the same UX since anyway we were hiding the button until there is a value).

Before:
image

After:
image

src/SideBar/MapLayersTab/AnalyzersTab/index.js
```diff
@@ -1,38 +1,33 @@
...
+import * as styles from '../styles.module.scss';
 
-const COLLAPSIBLE_LIST_DEFAULT_PROPS = {
-  lazyRender: false,
-  transitionTime: 1,
-};
 const mapLayerTracker = trackEventFactory(MAP_LAYERS_CATEGORY);
 
-// eslint-disable-next-line react/display-name
-const AnalyzerLayerList = memo(({
```
luixlive: I updated the component and file names, and refactor _a bit_ of the code for the new tab components, instead of keeping the old components directly below "src/" that were not being reused and had old Redux `connect` pattern. However, they still behave and render the same, just now within their corresponding tabs:
image

src/SideBar/MapLayersTab/EventsTab/index.js
```diff
@@ -0,0 +1,45 @@
...
+import CheckMark from '../../../Checkmark';
+import HeatmapToggleButton from '../../../HeatmapToggleButton';
+
+import * as styles from './styles.module.scss';
+import * as mapLayersStyles from '../styles.module.scss';
+
+const mapLayerTracker = trackEventFactory(MAP_LAYERS_CATEGORY);
+const reportsTracker = trackEventFactory(REPORTS_CATEGORY);
+
+const EventsTab = () => {
```
luixlive, line 18: image

src/SideBar/MapLayersTab/FeaturesTab/index.js
```diff
@@ -1,37 +1,33 @@
...
+import * as styles from '../styles.module.scss';
 
-const COLLAPSIBLE_LIST_DEFAULT_PROPS = {
-  lazyRender: false,
-  transitionTime: 1,
-};
 const mapLayerTracker = trackEventFactory(MAP_LAYERS_CATEGORY);
 
-// eslint-disable-next-line react/display-name
-const FeatureLayerList = ({
```
luixlive: image

src/SideBar/MapLayersTab/Filters/index.js
```diff
@@ -0,0 +1,217 @@
...
+} from '../../../ducks/map-layer-filter';
+import { TAB_KEYS } from '../utils/constants';
+
+import SearchBar from '../../../SearchBar';
+
+import * as styles from './styles.module.scss';
+
+const mapLayersTracker = trackEventFactory(MAP_LAYERS_CATEGORY);
+
+const Filters = ({ tab }) => {
```
luixlive, line 28: Not really new, we previously had the now deleted `MapLayerFilter`, but it was basically rewritten from scratch so our filters are fully accessible and keyboard navigatable. We use the new `SearchBar` and conditionally render the sortable buttons just in the "Subjects" tab. Also notice that the "Events" tab doesn't render this component at all since there's no sense filtering in there.
image

src/SideBar/MapLayersTab/SubjectsTab/index.js
```diff
@@ -0,0 +1,185 @@
...
+};
+
+const lastUpdateCompareFunction = (sortDirection) => (subjectA, subjectB) => {
+  if (subjectB.updated_at > subjectA.updated_at) {
+    return sortDirection === SORT_DIRECTION.down ? 1 : -1;
+  }
+  return sortDirection === SORT_DIRECTION.down ? -1 : 1;
+};
+
+const SubjectsTab = () => {
```
luixlive, line 67: Our new `SubjectsTab` which is an evolution of `SubjectGroupList`, but now with the capabilities of sorting, ungrouping, and rendering correctly in its own tab. Also, some of the utilities it imported from other files are now directly here since they are not reused anywhere else:
image

src/SideBar/MapLayersTab/index.js
```diff
@@ -0,0 +1,78 @@
...
+
+import AnalyzersTab from './AnalyzersTab';
+import EventsTab from './EventsTab';
+import Filters from './Filters';
+import FeaturesTab from './FeaturesTab';
+import SubjectsTab from './SubjectsTab';
+
+import * as styles from './styles.module.scss';
+
+const MapLayersTab = () => {
```
luixlive, line 16: Now our router renders the `MapLayersTab` when the route is `/layers`. This holds our tabbed component that renders the different map layer lists. It's code is very straightforward, just please notice the comment explaining why we needed to set `mountOnEnter` and `unmountOnExit`. This impacts the performance of the rendering of the tabs when switching, which is not great, but it's the only way to make our new tabs work properly with the collapse items each list holds.
image

src/ducks/map-layer-filter/index.js
```diff
@@ -0,0 +1,148 @@
...
+  type: SHOW_ANALYZERS,
+  payload: analyzerFeatureIDs,
+});
+
+export const displayReportsOnMapState = (enabled) => ({
+  type: DISPLAY_REPORTS_ON_MAP,
+  payload: enabled,
+});
+
+export const setMapLayersFilterText = (text) => ({ payload: text, type: SET_TEXT });
```
luixlive, line 65: The map layer filter duck was updated with new action creators: 
- setMapLayersGrouped
- setMapLayersSortBy
- setMapLayersSortDirection

An old action creator was updated to match the new pattern:
- setMapLayersFilterText

And the reducer was updated to properly handle the new actions.

Notice that we have a single filter for all the layers. The filter text affects all the layers, and the sorting options don't specify that they only work for subjects. I wanted to totally separate the filtering, like having a search text for subjects, one for features, one for analyzers and one for events. Same with sorting. But that was not defined or part of the scope, and since it involves a migration of the persisted reducer and potentially lots of code changes, I decided not to do so. It may be something good for future consideration (if we think it makes sense, maybe we actually want the different tabs to share the filters).

src/selectors/subjects.js
```diff
@@ -45,12 +45,26 @@ export const getSubjectGroups = createSelector(
     const hydrateSubjectGroupSubjects = (...groups) => groups.map((group) => {
       const { subgroups, subjects } = group;
 
+      const hydratedSubGroups = hydrateSubjectGroupSubjects(...subgroups);
+      const hydratedSubjects = subjects.map((id) => subjectStore[id]).filter((subject) => !!subject);
+
+      let updatedAt;
+      hydratedSubGroups.forEach((subGroup) => {
```
luixlive, line 52: When selecting subject groups, we recursively "hydrate" them and their subgroups (update their subjects and subgroups to the latest values from the store). While doing so, I decided to calculate a property `updated_at` and set it (recursively too of course) to every subjects group so we can easily sort them by the one that has the subject with the latest update without having to iterate an insane amount of times the subjects of each group to compare the groups with most recent updates in their children. This way we only do it once and now we can easily compare one group vs other when sorting by last update.

tiffanynwong:
Looking so good 😎 

1. For the Features and Analyzer tab, can we remove the high-level group? For example since the whole tab is meant to show features, we don't need this features group.
Screenshot 2025-07-22 at 12 53 38 PM

So instead if the user navigates to the features tab they would see these, without having to expand the "Features" group.
Screenshot 2025-07-22 at 12 54 18 PM

2. Are we not adding the last update timestamp as part of this ticket? The last update sort order isn't as helpful with out it. 
Screenshot 2025-07-22 at 12 55 25 PM
Screenshot 2025-07-22 at 12 56 52 PM

3. It would be nice if the app could remember which groups were expanded so that as I'm navigating between the tabs I don't have to keep re-opening groups. Another option would be to add an "Expand All" button.
Jul-22-2025 13-01-03

These aren't really in the scope of the ticket, but would be nice to fix:
4. The headers are just a tad too close too the top nav. Can we add a bit more padding above the header so it doesn't feel so squished at the top :) 
Screenshot 2025-07-22 at 1 05 54 PM
Screenshot 2025-07-22 at 1 09 10 PM

5. It's weird that the jump to location buttons are not right aligned like they are for subjects. It would look cleaner if it was.
6. The check mark, text and jump to location button aren't center aligned. 
Screenshot 2025-07-22 at 1 12 11 PM

JoshuaVulcan (approved):
There's a bit of ongoing layout work, but the bulk of this is here and looks/feels great. Approved on my end, short of any big regressions before merging!

luixlive:
> Looking so good 😎
> 
> 1. For the Features and Analyzer tab, can we remove the high-level group? For example since the whole tab is meant to show features, we don't need this features group.

Done! For both of them, I worked on a new accessible checkbox that still looks like the old one, but is keyboard navigatable and has screen reader support. We should consider updating the other check marks in these lists too.

> 2. Are we not adding the last update timestamp as part of this ticket? The last update sort order isn't as helpful with out it.

Done. Just notice that I decided to add the timestamp to the groups too. Each group has the timestamp of the most recent updated subject it contains. Why? Because we can also sort by last update when the subjects are grouped, and it can be very confusing to see groups moving around when sorting and changing the direction if they don't have the latest update too.

> 3. It would be nice if the app could remember which groups were expanded so that as I'm navigating between the tabs I don't have to keep re-opening groups. Another option would be to add an "Expand All" button.

Yes, agreed. However this is part of the explanation I added in the PR description, we had to mount and unmount the collapse components because they become unstable if we don't. Adding some kind of state to handle which ones are open / closed while navigating could be achieved but it would take some time. Sounds like work for another ticket.

> These aren't really in the scope of the ticket, but would be nice to fix: 4. The headers are just a tad too close too the top nav. Can we add a bit more padding above the header so it doesn't feel so squished at the top :) 
> 5. It's weird that the jump to location buttons are not right aligned like they are for subjects. It would look cleaner if it was.
> 6. The check mark, text and jump to location button aren't center aligned.

These 3 items are done too ✅ 

@gaboDev ready for review! @JoshuaVulcan I did a few changes so you may want to take a look at that last commit.

**Note:** Consider that even if this PR is approved, we should not merge it until https://github.com/PADAS/das-web-react/pull/1325 is merged since this branch is behind that.

PR #1315, opened 2025-07-11
Measuring ER Web Vitals

JoshuaVulcan:
### What does this PR do?
- Installs, configures, and implements `web-vitals` for EarthRanger. It sends the vitals to Google Analytics, but we can easily extend it to send them to Elastic Cloud or something else as well. The intent is to have some solid fundamental performance metrics to watch for improvement or decline as we release features such as vector tiles in the near future.

### How does it look
Metrics being measured:
* [First Contentful Paint (FCP)](https://web.dev/articles/fcp)
* [Time to First Byte (TTFB)](https://web.dev/articles/ttfb)
* [Cumulative Layout Shift (CLS)](https://web.dev/articles/cls)
* [Largest Contentful Paint (LCP)](https://web.dev/articles/lcp)

### Relevant link(s)
* Tracking tickets: https://allenai.atlassian.net/browse/ERA-11776

- 🤷


chrisj-er (approved):

src/WebVitalsProvider/__tests__/index.test.js
luixlive, line 1: Any reason to define a different testing file structure than we do everywhere else? I would have expected `src/WebVitalsProvider/index.test.js`.

src/WebVitalsProvider/__tests__/index.test.js
```diff
@@ -0,0 +1,130 @@
...
+    });
+
+    return render(
+      <Provider store={testStore}>
+        <WebVitalsProvider />
+      </Provider>
+    );
+  };
+
+  it('should render without crashing', () => {
```
luixlive: nit: I have always felt these tests unnecessary, if the component couldn't render for any reason, this test is not telling why. And the rest of the tests of this file would also crash so I don't think we are getting any benefit from them 🤔

src/WebVitalsProvider/index.js
```diff
@@ -0,0 +1,25 @@
+import { useEffect, useRef } from 'react';
+import { useSelector } from 'react-redux';
+import { initializeWebVitals, createUserAnalyticsData } from '../utils/webVitals';
+
+const WebVitalsProvider = () => { // Initializes web vitals tracking with user context
```
luixlive: Why not a hook? Renderless components were very useful before the hook era, but nowadays we have them, which are louder about the fact they don't render anything and just add some hook logic to a component. Since JSX resemble HTML, I would expect nodes to render something:
```JavaScript
cont Component = () => {
  // This component has no logic

  return <div>
    <WebVitalsProvider /> // It renders something (or at leas this is a real context Provider as the name suggests).
  </div>;
};
```

```JavaScript
cont Component = () => {
  useWebVitals(); // This component has some logic!

  return <div />;
};
```

src/WebVitalsProvider/index.js
```diff
@@ -0,0 +1,25 @@
...
+  const user = useSelector((state) => state.data?.user);
+  const selectedUserProfile = useSelector((state) => state.data?.selectedUserProfile);
+  const isInitialized = useRef(false);
+
+  useEffect(() => {
+    if (user && user.id && !isInitialized.current) {
```
luixlive: nit: most of the times `isInitialized.current` will be true so if we put that first we get a tiny optimization by avoiding the second check (which can be reduced using optional chaining):
`if (!isInitialized.current && user?.id) {`

src/utils/string.js
```diff
@@ -8,11 +8,22 @@ export const hashCode = (string) => {
...
+    hash = ((hash << 5) - hash) + char.charCodeAt(0);
     hash |= 0;
 
     return hash;
   }, 0);
 
   return hash;
 };
+
+export const hashString = (str) => { // Create a simple hash of a string for anonymization. Returns 'unknown' for falsy values, otherwise returns a base36 hash
```
luixlive: I think we should be more specific on why we need this `hashString` if we already have `hashCode` above. Maybe better naming for both of them? Btw they seem to do a very similar type of hashing, can probably this one just be a wrapper of the other one?

src/utils/string.test.js
```diff
@@ -12,4 +12,30 @@ describe('String utils', () => {
...
+      expect(hashString(undefined)).toBe('unknown');
+      expect(hashString('')).toBe('unknown');
+      expect(hashString(0)).toBe('unknown');
+      expect(hashString(false)).toBe('unknown');
+    });
+
+    test('returns consistent hash for same string', () => {
+      const hash1 = hashString('test-user-123');
+      const hash2 = hashString('test-user-123');
+      expect(hash1).toBe(hash2);
```
luixlive, line 28: These hashes are stable right? They return the same hash output if they get the same hash input every single time. Why not to just test the hash value like `expect(hashString('user-123')).toBe('as6s4d5s');` which would be equivalent to all the expects you have in a single one (and even better).

src/utils/webVitals.js
luixlive, line 1: I think these utilities are only used by the `WebVitalsProvider` right? Placing them in our global `src/utils` feels like the wrong scope since they won't be reused. I would suggest to create a `/WebVitalsProvider/utils/index.js`. Stays in the scope of its responsibility and doesn't increase the size of our global utils folder 👍

luixlive (approved):
Nice! A bunch of nits to improve a bit the code, but nothing to block it.

src/WebVitalsProvider/index.js
JoshuaVulcan: You're totally right 🤡 I'm just a bit rusty. Updating now

src/utils/webVitals.js
JoshuaVulcan: I think it's useful to have a generic string hashing function (`hashCode` is a numeric hashing function, so this is indeed a new addition).

PR #1316, opened 2025-07-11
ERA-11525: addings data-testid for better handling of locators

AlanCalvillo:
### What does this PR do?
- read title

### DIFF
--- a/src/SubjectGroupList/Content.js
+++ b/src/SubjectGroupList/Content.js
@@ -48,7 +48,7 @@ const TriggerComponent = memo((props) => {  // eslint-disable-line react/display
   const fullyVisible = track === TRACKING_CONTROL_STATES.FULLY_VISIBLE;
   const partiallyVisible = track === TRACKING_CONTROL_STATES.PARTIALLY_VISIBLE;
 
-  return <div className={listStyles.trigger}>
+  return <div className={listStyles.trigger} data-testid={`collapsible-${itemTitle}`}>
     {listLevel === 0 && <h5>{itemTitle}</h5>}
     {listLevel > 0 && <h6>{itemTitle}</h6>}
     {showTrackingControls && <>


src/SubjectGroupList/Content.js
```diff
@@ -48,7 +48,7 @@ const TriggerComponent = memo((props) => {  // eslint-disable-line react/display
   const fullyVisible = track === TRACKING_CONTROL_STATES.FULLY_VISIBLE;
   const partiallyVisible = track === TRACKING_CONTROL_STATES.PARTIALLY_VISIBLE;
 
-  return <div className={listStyles.trigger}>
+  return <div className={listStyles.trigger} data-testid={`collapsible-${itemTitle}`}>
     {listLevel === 0 && <h5>{itemTitle}</h5>}
```
JoshuaVulcan, line 52: This is rendered recursively, and will fail with repeated subject groups. 

ex/Place the "Rangers" group as a subgroup of both the "Field Staff" group and the "Security Staff" group and you end up with two instances of `data-testid="collapsible-rangers"

luixlive: There's no problem with data-testids being duplicated right? And I guess Alan wants this for automation tests in envs where the data is controlled. 

I'll wait for @AlanCalvillo to decide how important this is 👍

JoshuaVulcan: As long as he doesn't rely on that uniqueness for anything...but given this is for our automated tests, which should be able to run against live sites, and lots of EarthRanger users use subject groups that way, it's not like this is contrived or a corner case.

AlanCalvillo: i guess that's fine. i can always delimit the scope by calling first the parent group. so, in your example i can do 
```
const fieldStaffCollapsible =  page.getByTestId('collapsible-FieldStaff');
const rangerSubGroups = fieldStaffCollapsible.getByTestId('collapsible-Rangers');
```
that way, i can reach to the one i want to interact with

if i want to do an operation to all the collapsible entries, i can do page.getByTestId(/collapsible-.*/).click(); and playwright will only open the ones that are visible, bypassing the click of what is not visible.

i just wanted to have a way to access to the parent of a collapsible element, having that parent its way easier to interact with its children elements.

AlanCalvillo: you can see the usage here:

1.- first i have a way to get all locators that match a name generically.
https://github.com/PADAS/er-web-automation-playwright/blob/1edcb664633324ed748e1474b72782d2e4be93fa/tests/das-ui/pageObjects/mapLayers/List.page.js#L8
2.- second i can get all toggle tracks button (or whatever children locators) the collapsible has
https://github.com/PADAS/er-web-automation-playwright/blob/1edcb664633324ed748e1474b72782d2e4be93fa/tests/das-ui/pageObjects/mapLayers/List.page.js#L43
3.- i can use all the locators matched to make assertions
https://github.com/PADAS/er-web-automation-playwright/blob/1edcb664633324ed748e1474b72782d2e4be93fa/tests/das-ui/tests/subjects/subjects.spec.js#L112

JoshuaVulcan: Yep. It's _currently_ safe, but now these selectors are sketchy if I want to write tests targeting these elements a year in the future...if I don't have this safe handling context in my mind, it could yield unexpected results and annoyance.

It's a smelly smell to have duplicated IDs, on the simple grounds that an ID is supposed to _uniquely_ identify something.

AlanCalvillo: i agree with that, but other options introduce a similar level of sketchiness in the tests 😅

opt 1: adding an iterator to the data-testid adds complexity, especially when trying to map that number to its actual position in the ui.

opt 2: appending the breadcrumb is pretty much what we’re already doing — relying on fixed data maintained by the test.

opt 3: using a generated hash isn’t ideal either, since it could change and break stability.

do you have any ideas in mind?

luixlive (approved):
From my side I think this is ok. I agree on trying to make ids _unique_ as much as possible, but data-testids are tricky. They need to be stable so you can easily reference them from your tests, so you can't use React's utility `const id = useId();`. And some times you have unique strings in your component that could be used as ids, but sometimes you dont. Think of this trivial SearchBar component:
```JavaScript
const SearchBar = ({ label, onChange, value }) => {
  const id = useId();

  return <div className={styles.searchBar} data-testid="search-bar">
    <label htmlFor={id}>{label}</label>
    <input id={id} onChange={onChange} value={value} type="search" />
  </div>;
};
```

You shouldn't enforce the parent to send an id, and you can't use the same id generated by `useId` since you don't know what it's value will be, you need to know it in order to select it from your tests. The solution is to set a constant string like "search-bar" but you don't know how many times the component will be rendered in the screen, so you may have repeated ids, and in data-testids, I have always had the conception that it is fine.

If Alan hasn't found a stable id, I feel like allowing them to be repeated is fine. Happy to continue the discussion, but I'll leave my approval to unblock Alan @JoshuaVulcan

PR #1317, opened 2025-07-12
ERA-11768: Coordinate reference system transformation research

luixlive:
### What does this PR do?
Spike to determine the steps to follow to tackle [ERA-9788](https://allenai.atlassian.net/browse/ERA-9788).

This PR has some nice initial code to get us started with the custom Coordinate Reference Systems (CRS):
- Define a new feature flag to hide the unstable code while the epic is under development: `REACT_APP_CUSTOM_COORDINATE_SYSTEMS_ENABLED`.
- Install the two dependencies we will require:
  - [epsg-index](https://www.npmjs.com/package/epsg-index):`A machine-readable index of all EPSG coordinate systems.`. Part of this PR is the new utility method `getProj4CompatibleCRS` to dynamically import the list of CRS that support [proj4](http://proj4js.org/) and sanitize the data.
  - [proj4](http://proj4js.org/): `JavaScript library to transform coordinates from one coordinate system to another, including datum transformations.`. Exactly what we will need, and this is the official library. The idea is to use it as its API suggest `proj4([fromProjection, ]toProjection[, coordinates])` to transform from the user selected coordinate system to `WSG 84` so we can get a regular latitude / longitude object.
- Update the our `Map` tab under the `Settings` pane of the side bar so it supports internal navigation to show the `Coordinates` settings view.
- Add a "Go to coordinates view" button in the `Display` fieldset of the `Map` tab that opens the `Coordinates` settings view and show the "coordinate system selector" widget, where you can search coordinate systems, add them to the checkable list and select the ones you want to enable in your environment. Note that as part of this spike all the UI elements are extremely rudimentary, no styles, proper semantics, a11y, i18n...
- Add a new Redux duck to handle the Coordinate Reference Systems app settings, which is integrated with the selector widget.
- Read the value of the new Redux duck in the `GpsFormatToggle` component to update the tabs the location picker shows.

What this PR does *not* do (out of the spike):
- Proper UX / UI of all the rendered components.
- Map layer / source for projected CRS that have bounds.
- Actual location data transformation from the selected CRS to latitude and longitude.
- Fallback to DEG in case of locations out of projected CRS with bounds.
- Error states for `GpsFormatToggle` options that can't show a location value.

### How does it look
Kapture 2025-07-11 at 19 57 01

### Relevant link(s)
* [ERA-11768](https://allenai.atlassian.net/browse/ERA-11768)
* [Env](https://era-11768.dev.pamdas.org)

[ERA-9788]: https://allenai.atlassian.net/browse/ERA-9788?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ERA-11768]: https://allenai.atlassian.net/browse/ERA-11768?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/App.scss
```diff
@@ -271,6 +271,7 @@ aside {
 
 .app-container {
   flex: 1;
+  max-height: calc(100vh - 2.75rem);
```
luixlive, line 274: You will see a few layout CSS changes that seem unrelated to this ticket, but it's kind of related,  since we are working in this new coordinate settings view in the side bar. The scroll and width of the side bar is very funky. It's not well designed and some times achieving the desired layouts in it is very difficult. These changes are a step towards a better flex-box, height-defined, side bar. It's not the final fix but at least I can tell the Settings panel of the side bar looks way better.

Just a small example of this, right now, in https://root.dev.pamdas.org/ you can scroll and hide the tabs:
Kapture 2025-07-11 at 20 19 12

I bet this can be frustrating for our users! It doesn't make sense. But thanks to these changes, it doesn't happen anymore. Now the scroll is within the tabs.

src/GpsFormatToggle/index.js
```diff
@@ -45,26 +59,27 @@ const GpsFormatToggle = ({ lat = null, lng = null, name, ref, showGpsString = tr
     <fieldset className={styles.fieldset} ref={fieldsetRef}>
       <legend className={styles.legend}>{t('fieldsetLegend')}</legend>
 
-      {Object.values(GPS_FORMATS).map((itemGpsFormat) =>
+      {/* TODO (CRS): Style label to be in a single line with ellipsis */}
+      {gpsFormatOptions.map((gpsFormatOption) =>
```
luixlive, line 63: Basically, the changes in our GpsFormatToggle (protected by the ff of course) are to read from the new Redux reducer the `selectedCRS` in order to show its tabs instead of hardcoding the `GPS_FORMATS`:
image

src/GpsInput/index.js
```diff
@@ -3,9 +3,9 @@ import { useSelector } from 'react-redux';
 import { useTranslation } from 'react-i18next';
 
 import {
-  calcActualGpsPositionForRawText,
   calcGpsDisplayString,
   GPS_FORMAT_EXAMPLES,
+  normalizeGpsFormatTextToLngLat,
```
luixlive, line 8: Nothing changed here. I just "normalized" the wording in the location util methods, so I had to update them here.

src/SideBar/SettingsPane/MapTab/CoordinateSettingsView/index.js
```diff
@@ -0,0 +1,186 @@
...
+import { getProj4CompatibleCRS, GPS_FORMATS } from '../../../../utils/location';
+import {
+  setSelectedCoordinateReferenceSystems,
+  setStoredCoordinateReferenceSystems,
+} from '../../../../ducks/coordinate-reference-systems';
+
+const MAX_SELECTED_GPS_FORMATS = 5;
+const MAX_FILTERED_PROJECTION_RESULTS = 10;
+
+const GpsFormatOption = ({ identifier, isDeletable, label }) => {
```
luixlive, line 15: The new coordinate settings view, which renders the CRS selector widget:
image

src/SideBar/SettingsPane/MapTab/MainMapSettingsView/DisplayFieldSet/index.js
```diff
@@ -74,6 +78,9 @@ const DisplayFieldSet = () => {
   return <fieldset className={styles.section}>
     <legend className={styles.title}>{t('legend')}</legend>
 
+    {/* TODO (CRS): I18n and a11y */}
+    {customCoordinateSystemsEnabled && <button onClick={() => onShowCoordianteSettingsView()}>Coordinates</button>}
```
luixlive, line 82: The only real change here is the new button to navigate to the coordinate settings view:
image

src/SideBar/SettingsPane/MapTab/MainMapSettingsView/DisplayFieldSet/index.test.js
luixlive, line 1: Since now our `MapTab` doesn't only render the "Main Map Settings View" (as I decided to call it), but may also render the "Coordinate Settings View" through a stateful navigation fashion, we had to re-arrange a bit the folder structure of these components, but nothing really changed here.

src/SideBar/SettingsPane/MapTab/MainMapSettingsView/index.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+
+import DisplayFieldSet from './DisplayFieldSet';
```
luixlive, line 3: This is the (not so) new "Main Map Settings View". It used to be the only thing rendered by our `MapTab`, but since now we can also navigate to the "Coordinate Settings View" and it steals the same area of the tab, we had to define these main map settings as an independent component:
image

src/SideBar/SettingsPane/MapTab/index.js
```diff
@@ -1,23 +1,20 @@
...
+import CoordinateSettingsView from './CoordinateSettingsView';
+import MainMapSettingsView from './MainMapSettingsView';
 
-import * as styles from './styles.module.scss';
+const VIEWS = { COORDINATE_SETTINGS: 'coordinates' };
 
-const MapTab = () => <>
-  <GeneralFieldSet />
+const MapTab = () => {
+  const [currentView, setCurrentView] = useState(null);
```
luixlive, line 9: Now the `MapTab` has a very simple logic to support navigation through state (this is the variable that defines the current view, as its name suggests). By default it is `null`, meaning we are in the default view: `MainMapSettingsView`. But through the `coordinates` button now we can show the `CoordinateSettingsView` in its place. This pattern is defined in a trivial yet scalable way, so if in the future we add more settings that require to steal the entire tab space, we can just add a new property to the `VIEWS` constant above and a new `case` in the `switch` 😄

src/ducks/coordinate-reference-systems/index.js
```diff
@@ -0,0 +1,46 @@
...
+  payload: storedCoordinateReferenceSystems,
+});
+
+// Reducer
+export const INITIAL_STATE = {
+  selectedSystems: Object.values(GPS_FORMATS),
+  storedSystems: [],
+};
+
+const coordinateReferenceSystemsReducer = (state = INITIAL_STATE, action) => {
```
luixlive, line 26: The new Redux duck to handle the CRS state. It's state looks like this:
```JavaScript
{
  selectedSystems: ['DEG', 'DMS', 'DDM', 'UTM', '5367'], //  where 5367 is an EPSG code.
  storedSystems: ['5367'], // EPSG codes that the user has added to the selector widget.
}
```

`storedSystems` don't hold the default GPS formats supported by ER (`DEG`, `DMS`, `DDM`, `UTM`, `MGRS`), those are implicit, and always rendered in the widget. Only custom coordinate systems can be added or deleted.

src/reducers/index.js
```diff
@@ -115,6 +117,7 @@ const rootReducer = combineReducers({
     patrolLeaderSchema: patrolTrackedBySchemaReducer,
   }),
   view: combineReducers({
+    coordinateReferenceSystems: persistReducer(coordinateReferenceSystemsStorageConfig, coordinateReferenceSystemsReducer),
```
luixlive, line 120: Persist the user selection! You may want to take a second look at the way we defined the reducer because once this goes live, the only way to update that reducer is through a migration.

src/utils/location.js
```diff
@@ -1,7 +1,7 @@
 import { bearing } from '@turf/turf';
 import Dms from 'geodesy/dms';
-import Utm, { LatLon as LatLon_Utm } from 'geodesy/utm';
-import Mgrs, { LatLon as Latlon_Utm_Mgrs } from 'geodesy/mgrs';
+import Utm, { LatLon as LatLonUtm } from 'geodesy/utm';
```
luixlive, line 3: I did a full refactor to this file. Since we are going to touch these methods a lot in the following tickets, I took the time to clean them a bit (and protect them more agains invalid user input).

src/utils/location.js
```diff
@@ -22,119 +22,172 @@ export const GPS_FORMAT_EXAMPLES = {
...
 export const calcLocationParamStringForUserLocationCoords = (coords) => `${coords.longitude},${coords.latitude}`;
 
 export const validateLocation = (location) => typeof location?.lng === 'number'
   && typeof location?.lat === 'number'
   && validateLngLat(location.lng, location.lat);
+
+// epsg-index/all.json is 5mb. It's imported dynamically to keep it out of the
+// app bundle and we store it in a singleton to only fetch it once.
+let proj4CompatibleCRSSingleton = null;
+export const getProj4CompatibleCRS = async () => {
```
luixlive, line 171: The new `getProj4CompatibleCRS`. Simple and as performant as it can be. I dynamically import the CRS so the bundle size doesn't grow (and we don't load them if they are not necessary) and also used a singleton pattern to make sure that once they are loaded, they are not re-loaded.

JoshuaVulcan (approved):
This is a great head-start! Nice work 🚀

PR #1319, opened 2025-07-14
[ERA-8351] Using vector tiles for spatial features

JoshuaVulcan:
https://era-8351-alt.dev.pamdas.org/


.env.development
```diff
@@ -1,4 +1,4 @@
-REACT_APP_DAS_HOST='https://root.dev.pamdas.org'
+REACT_APP_DAS_HOST='https://era-8351-alt.dev.pamdas.org'
```
luixlive, line 1: I don't think we want this merged into `develop`

src/EarthRangerMap/index.js
```diff
@@ -56,6 +58,27 @@ const EarthRangerMap = ({ children, controls, onMapLoaded, ...otherProps }) => {
...
   }, [onMapLoaded]);
 
+  /* Keep latest access token in a ref so transformRequest always sees it without
+   needing to recreate the map instance when the token changes. */
+  const tokenRef = useRef();
+  tokenRef.current = token?.access_token;
+
+  /* Ensures Authorization header is added to any spatial features tile requests. 
+   This survives style changes that would otherwise drop per-source transformRequest handlers. */
+  const transformRequest = useCallback((url, resourceType) => {
```
luixlive, line 68: I personally find the code added to this file confusing. This is called `transformRequest`, but I don't understand what the word `transform` stands for. I see that we have a string hardcoded in `resourceType === 'Tile'` that makes me think this has something to do with tiles but the logic inside the condition doesn't seem related to tiles at all, it's just returning an object with a URL (that the implementator already provided from the parameters so I don't get why we need to return it back) and a pretty conventional Authorization header 🤔

src/LocationPicker/index.js
```diff
@@ -117,7 +119,7 @@ const LocationPicker = ({
         title={t('textCopyButtonLabel')}
       />}
 
-      <button
+      {map && <button
```
luixlive: Are we mixing the vector tiles feature with the ticket to asynchronously load app components from the map?

src/App.js
```diff
@@ -185,9 +185,9 @@ export const App = () => {
           {!mapboxSupported && <ErrorMessage className='webgl-error-message'
             message={t('webGlDisabled')} />}
 
-          {!!map && <SidebarScrollProvider>
```
JoshuaVulcan, line 188: Removing the `map` prerequisite for the Sidebar was pretty much the only step necessary to fulfill the requirements for https://allenai.atlassian.net/browse/ERA-11903.

src/LocationPicker/index.js
luixlive: And in that case, this change of conditionally rendering the button until the map is available would produce layout shift in the screen. I would suggest to always show the button and just disable / enable it depending on the map variable. Another suggestion would be to not do modifications here, and instead make sure that `jumpToLocation` handles correctly the scenario of and `undefined` `map` variable so we don't need to go and add conditions in all places where that method is implemented. Anyway, I don't think users should be confused if they click the button while the map is not loaded yet and nothing happens haha.

src/Map/index.js
```diff
@@ -380,12 +377,14 @@ const Map = ({ children, onMapLoad, socket }) => {
...
 
-  const onFeatureSymbolClick = withLocationPickerState(({ geometry, properties }) => {
-    const coordinates = Array.isArray(geometry.coordinates[0]) ? geometry.coordinates[0] : geometry.coordinates;
+  const onFeatureSymbolClick = useCallback((feature) => {
+    const { geometry, properties } = feature;
 
-    showPopup('feature-symbol', { geometry, properties, coordinates });
-    mapInteractionTracker.track('Click Map Feature Symbol Icon', `Feature ID :${properties.id}`);
-  });
+    if (geometry.type === 'Point') {
```
luixlive, line 383: It seems like the old logic supported more than point geometries and now we are removing that. Why is that?

src/Map/index.js
```diff
@@ -584,6 +583,32 @@ const Map = ({ children, onMapLoad, socket }) => {
     }
   }, [i18n.language, map]);
 
+  useEffect(() => {
+    const handleMapStyleImageMissing = async (event) => {
+      const { id } = event;
+      // querying from the root /static/ dir of the host means this is one of our static assets, let's get it
+      // if the map says it's missing.
+      if (id.includes('/static/')) {
+        const src = id.replace(/(\.svg|\.png|\.jpg).*$/, '$1');
```
luixlive, line 592: Could you explain a bit this method? If the map is missing a style image, we are replacing the src file extension with $1, but why?

src/Popup/index.js
```diff
@@ -12,14 +14,16 @@ const Popup = ({ className = '', trackPointer = false, offset, coordinates, anch
   const popupRef = useRef(null);
   const popupContainerRef = useRef(document.createElement('div'));
 
+  const locationIsValid = validateLngLat(coordinates);
```
luixlive, line 17: This variable should go within the `useEffect` since it's only used in that scope. Right now we are 1- redefining it in every render and 2- adding an unnecessary memoization check in the `useEffect` dependency array. The only dependency of this constant is `coordinates` and that is already part of the dependency array of the `useEffect`.

src/ReportListItem/index.js
```diff
@@ -1,7 +1,9 @@
-import React, { memo, useCallback, useRef } from 'react';
+import React, { memo, useCallback, useContext, useRef } from 'react';
 import { useDispatch } from 'react-redux';
+
```
luixlive, line 3: nit: I see a bunch of unexpected new lines around in this PR.

src/SchemaFields/index.test.js
```diff
@@ -57,7 +57,7 @@ describe('DateTimeWidget', () => {
     expect((await screen.findByText('Date Time Widget'))).toBeDefined();
   });
 
-  test('triggers the onChange callback when changing the date', async () => {
+  test.skip('triggers the onChange callback when changing the date', async () => {
```
luixlive, line 60: 😅

src/SideBar/MapLayersTab/FeaturesTab/FeatureListItem.js
```diff
@@ -1,23 +1,48 @@
...
+import { POLYGONS_LAYER_ID, LINES_LAYER_ID } from '../../../SpatialFeaturesLayer';
+
+import { MapContext } from '../../../App';
 
 import * as styles from '../styles.module.scss';
 
 const mapLayerTracker = trackEventFactory(MAP_LAYERS_CATEGORY);
 
 // eslint-disable-next-line react/display-name
 const FeatureListItem = memo((props) => {
```
luixlive, line 23: nit: a well stablished pattern we've been using is to spread the props from the component argument like
```JavaScript
const FeatureListItem = ({ properties, map, geometry, showFeatures, showPopup }) => {
  ...
};

// Get rid of the eslint disabling:
return memo (FeatureListITem);
```

src/SideBar/MapLayersTab/FeaturesTab/index.js
```diff
@@ -30,12 +30,16 @@ const FeaturesTab = () => {
...
   const featureFilterEnabled = mapLayerFilter.text.length > 0;
 
-  const allFeatureIDs = useMemo(() => getUniqueIDsFromFeatures(
-    ...featureLayerList.reduce((accumulator, { featuresByType }) => [
-      ...accumulator,
-      ...featuresByType.reduce((result, { features }) => [...result, ...features], [])
-    ], [])
-  ), [featureLayerList]);
+  const allFeatureIDs = useMemo(() => {
+    return featureLayerList.reduce((accumulator, item, _idx) => {
```
luixlive, line 34: Could you explain why we needed to do this change? I don't get how we ended up needing to tweak this code, pretty sure there's a good reason. Also, we can just remove the unused `_idx` variable so we don't list it in the arguments.

src/SideBar/MapLayersTab/FeaturesTab/selectors.js
```diff
@@ -1,20 +1,34 @@
...
-      features: set.geojson.features.filter(f => f.properties.type_name === name),
-    }));
-    return ({
-      name: set.name,
-      id: set.id,
-      featuresByType,
-    });
-  }),
+  (featureSets) => featureSets
+    .reduce((result, set, _index) => {
```
luixlive, line 8: Same question about the origin of this change. And same thing about the `_index`.

src/SpatialFeaturesLayer/index.js
```diff
@@ -0,0 +1,358 @@
...
+import { useSelector } from 'react-redux';
+import { MapContext } from '../App';
+import { API_URL, DEFAULT_SYMBOL_LAYOUT, DEFAULT_SYMBOL_PAINT } from '../constants';
+
+const SPATIAL_FEATURES_SOURCE = 'spatial-features-source';
+
+const VECTOR_TILE_URL = `${API_URL}spatialfeatures/tiles/{z}/{x}/{y}.pbf`;
+
+export const SYMBOLS_LAYER_ID = 'spatial-features-symbols';
+// const SYMBOLS_LABELS_LAYER_ID = 'spatial-features-point-labels';
```
luixlive, line 11: Can we remove these comments?

src/SpatialFeaturesLayer/index.js
```diff
@@ -0,0 +1,358 @@
+import React, { memo, useContext, useMemo, useCallback, useEffect } from 'react';
+import { useSelector } from 'react-redux';
+import { MapContext } from '../App';
+import { API_URL, DEFAULT_SYMBOL_LAYOUT, DEFAULT_SYMBOL_PAINT } from '../constants';
+
+const SPATIAL_FEATURES_SOURCE = 'spatial-features-source';
```
luixlive, line 6: I think we have a constant in the global `constants` files holding all the source and layer ids so we can reference them in other places in the app if needed.

src/SpatialFeaturesLayer/index.js
```diff
@@ -0,0 +1,358 @@
...
+const VECTOR_TILE_URL = `${API_URL}spatialfeatures/tiles/{z}/{x}/{y}.pbf`;
+
+export const SYMBOLS_LAYER_ID = 'spatial-features-symbols';
+// const SYMBOLS_LABELS_LAYER_ID = 'spatial-features-point-labels';
+export const LINES_LAYER_ID = 'spatial-features-lines';
+// const LINES_LABELS_LAYER_ID = 'spatial-features-line-labels';
+export const POLYGONS_LAYER_ID = 'spatial-features-polygons';
+// const POLYGONS_LABELS_LAYER_ID = 'spatial-features-polygon-labels';
+
+const defaultLinePaintColor = [
```
luixlive: nit: we usually name file constants like these in uppercase.

src/SpatialFeaturesLayer/index.js
```diff
@@ -0,0 +1,358 @@
...
+            ['has', 'outline_color'], ['get', 'outline_color'],
+            ['has', 'border_color'], ['get', 'border_color'],
+            '#ff6600'
+          ]
+        },
+        filter: polygonLayerFilter
+      });
+    }
+
+    /* // Add separate label layers for each geometry type
```
luixlive, line 176: Can we remove these comments too please 🙏

src/SpatialFeaturesLayer/index.js
```diff
@@ -0,0 +1,358 @@
...
+    });
+
+    map.on('mouseenter', SYMBOLS_LAYER_ID, onMouseEnter);
+    map.on('mouseleave', SYMBOLS_LAYER_ID, onMouseLeave);
+
+    return () => {
+      layerIds.forEach(layerId => {
+        if (map.getLayer(layerId)) {
+          map.off('click', layerId, handleFeatureClick);
+          map.removeLayer(layerId);
```
luixlive, line 279: I think we are missing the removal (cleaning) of the source.

src/SpatialFeaturesLayer/index.js
```diff
@@ -0,0 +1,358 @@
...
+
+      map.setPaintProperty(polyLayer.id, 'fill-color', highlightPolygonFillColor);
+    }
+
+  }, [map, mapFeatureHighlightIDs]);
+
+  useEffect(() => {
+    const symbolLayers = [
+      map?.getLayer?.(SYMBOLS_LAYER_ID),
+      // map?.getLayer?.(SYMBOLS_LABELS_LAYER_ID),
```
luixlive, line 326: More comments jeje. And another suggestion: I see many `useEffects` with different responsibilities like updating layer paints and stuff like that. Some of them are very straightforward but some other have certain Mapbox complexity. Could we add a one line comment at the top of each useEffect explaining what it does?

src/ducks/map-ui/index.js
```diff
@@ -195,8 +197,23 @@ export const setShowMapClusterPolygons = (showMapClusterPolygons) => ({
...
+
+export const setMapFeatureHighlightIDs = (featureIDs) => ({
+  type: SET_MAP_FEATURE_HIGHLIGHT_IDS,
+  payload: featureIDs,
+});
+
 // Reducers
 
+const INITIAL_FEATURE_HIGHLIGHT_STATE = [];
+export const mapFeatureHighlightIdReducer = (state = INITIAL_FEATURE_HIGHLIGHT_STATE, action) => {
```
luixlive: Suggestion: I've found easier to maintain, debug, inspect code, etc... To define a single reducer per duck file. The origin of certain state values get obscured by adding several reducers into a single duck file and then injecting them from the `reducers` file. We can just create a `map-feature-highlight-ids` folder with its duck and its test files, or not add a new reducer at all and just find an existing one where the `mapFeatureHighlightIds` array can be added as an extra property and add its action creator there.

Btw, we are missing unit tests here!

luixlive (changes requested):
Nice! I provided a bunch of small suggestions and I know we are missing unit tests, which are important before merging. But these changes area very exciting. 

Do we have an environment where I could test? The code makes sense but I'd like to be able to experience by myself the feel of these layers.

.env.development
JoshuaVulcan: Yep, it's just in here for now to point you towards a good test environment when viewing/validating locally

src/EarthRangerMap/index.js
JoshuaVulcan: `transformRequest` is the configuration name for the mapbox setting we're crafting with this function, so it seemed appropriate IMO.

This is a request interceptor configuration for the Map instance. It is saying "if this map is requesting a Tile-typed asset from our own API path, append our API bearer token to the request for auth purposes". It is a nice, general way to authenticate both this service and future tile services (vector or not) that originate from our own applications.

src/Map/index.js
JoshuaVulcan: This follows the same conventions: most functionality is shared, while the rendering of the popup on click was scoped to point-typed features. Check out any existing env to validate there if you'd like.  The only real change is the responsibility to filter out events by geo-type moved into the handler function, up out of the layer component, which I like better as it lets the layer be a general-use all-types component instead of segregating data by geometry-type, like we used to.

src/Map/index.js
JoshuaVulcan: That just sanitizes the path to make sure it's a good image path. Otherwise all we're doing is catching the `styleimagemissing` event and adding the image. This is because Mapbox expects all images to be either:

1. Part of the base style, or
2. Loaded dynamically before adding the data sources+layers that will use them.

Since our data is dynamic and the overall image set is too large to pre-fetch and load, we just handle it dynamically:

1. Receive data
 2. Warn that image is not loaded to the map
 3. Sanitize URL + load image to the map
 4. Image shows for symbol layer on next render loop

src/SpatialFeaturesLayer/index.js
JoshuaVulcan: The commented out code in here is all the code to show text labels for each different geometry type. I would like to leave it in to possibly use soon... 😇

src/SpatialFeaturesLayer/index.js
JoshuaVulcan: See above: I want to keep the commented-out code here for future use in displaying text labels in the layers.

JoshuaVulcan: But yes! I like the idea of inline comments. Although I'd prefer if the code were legible enough to just read 😓 

Any in particular that are causing confusion?

src/ducks/map-ui/index.js
JoshuaVulcan: Agreed. I'm just following existing conventions to keep the code surface area small 😅 and I didn't find a suitable place in the app state to sidecar-in this property.

JoshuaVulcan: Will add tests + move reducer

JoshuaVulcan:
@luixlive feedback addressed, ready for another pass 🫡

src/SpatialFeaturesLayer/index.js
luixlive: For me, just the fact that there are too many. It's not like they are impossible to understand, but it makes me dizzy to see many small useEffects with similar code. Adding a small comment like `// Refresh the layer pain when ...` at the top of them would make the code faster to grasp!

src/EarthRangerMap/index.js
luixlive: Ok, I guess we could at least move the magic string to a constant like `MAPBOX_TILE_RESOURCE_TYPE`.

src/Map/index.js
luixlive: I guess my question now is, why do we need this code and we didn't before? Don't we have some initial code that loads these images? Why do we need to wait for the map to trigger a `styleimagemissing` in order to try to add the map image?

JoshuaVulcan: Because the image paths used to be transparent as properties within the GeoJSON, whereas now they are opaque and intrinsic to the vector data. So before, we had previous knowledge of the images and could load them. That's no longer the case, and styleimagemissing is the convention for handling those cases.

src/SideBar/MapLayersTab/FeaturesTab/FeatureListItem.js
```diff
@@ -80,3 +81,43 @@ const FeatureListItem = memo((props) => {
...
+let animationFrame = null;
+let t = 0;
+
+const animate = (map, layer_id, feature_id) => {
+  t += 0.1;
+  const pulse = (Math.sin(t) + 1) / 2; // oscillates between 0–1
+
+  const idExpr = ['==', ['get', 'id'], feature_id];
+
+  map.setPaintProperty(layer_id, 'icon-opacity', [
```
luixlive, line 94: Interesting approach to add an animation. Doesn't it overwhelm the map to update the pain property so fast? This code contrasts too much from the rest of the `FeatureListItem` stuff, maybe we could think of a hook to reuse this map layer animation pattern later!

luixlive:
> @luixlive feedback addressed, ready for another pass 🫡

Actually I see many of my comments without response and no changes done 😅 Not sure if you missed them @JoshuaVulcan . I resolved the conversation items that were already addressed.

JoshuaVulcan:
> > @luixlive feedback addressed, ready for another pass 🫡
> 
> Actually I see many of my comments without response and no changes done 😅 Not sure if you missed them @JoshuaVulcan . I resolved the conversation items that were already addressed.

@luixlive criticial/functional feedback has been addressed. Anything not touched is stylistic or preferential, like capital-casing and inline comments on those effects hooks. I'll add them now, but we should discuss at some point the different degrees of feedback and what we consider blocking/non-blocking for feature delivery.

src/SideBar/MapLayersTab/FeaturesTab/FeatureListItem.js
JoshuaVulcan: > maybe we could think of a hook to reuse this map layer animation pattern later!

Totally! Later being the key word 😆

src/LocationPicker/index.js
JoshuaVulcan: Done!

JoshuaVulcan:
@luixlive did a bit more fit and finish work from your feedback and a bit of stuff I saw. Care to give it another pass?

luixlive:
> @luixlive criticial/functional feedback has been addressed. Anything not touched is stylistic or preferential, like capital-casing and inline comments on those effects hooks. I'll add them now, but we should discuss at some point the different degrees of feedback and what we consider blocking/non-blocking for feature delivery.

Oh totally, I 100% agree that not all feedback is a PR blocker. But I think I was expecting a reply or something because I wasn't sure if you didn't seem them, forgot to send a commit, etc 😅

src/LocationJumpButton/index.js
```diff
@@ -72,6 +75,7 @@ const LocationJumpButton = ({
     : <MarkerIcon />;
 
   return isValidLocation && <button
+    disabled={!map}
```
luixlive, line 78: I just see a possible issue here, if the implementator provides a `disabled` prop with some custom logic, this would be overriden. Maybe we want to read the `disabled` prop explicitly and use it here like:
```Javascript
const LocationJumpButton = ({ disabled, ... }) => {
  ...
  return isValidLocation && <button disabled={disabled || !map} ... />;
};
```

src/ducks/map-ui/index.js
```diff
@@ -31,6 +31,7 @@ const SET_BOUNCE_EVENT_ID = 'SET_BOUNCE_EVENT_ID';
 
 const SET_MAP_DATA_ZOOM_SIMPLIFICATION = 'SET_MAP_DATA_ZOOM_SIMPLIFICATION';
 
+
```
luixlive, line 34: 😅

luixlive (approved):

PR #1320, opened 2025-07-14
ERA-11774: Event creation dialog' "jump to" dropdown doesn't show the list of event categories

gaboDev:
### What does this PR do?
- Adds default behavior of jump-to-event select in`AddReportTab` to avoid hiding the menu behind the modal.

### How does it look
- image

### Relevant link(s)
* Tracking tickets: [ERA-11774](https://allenai.atlassian.net/browse/ERA-11774)

[ERA-11774]: https://allenai.atlassian.net/browse/ERA-11774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/AddItemButton/AddItemModal/AddReportTab/index.js
```diff
@@ -84,6 +84,8 @@ const AddReportTab = ({ onHideModal }) => {
         getOptionLabel={(option) => option.display}
         isSearchable
         onChange={onQuickJumpSelectChange}
+        menuPortalTarget={null}
```
luixlive: nit: use double quotes for JSX strings `menuPlacement="bottom"`

luixlive (changes requested):
@gaboDev , this is not fixing the issue, just patching it in the only place that was reported. Now that we noticed the problem of the new defaults of our custom `Select` component, we should go and verify that it works properly in all the places where it is used. You just repaired it for one place, but for example, I noticed that we are using it in `src/MessagingSelect/index.js` too:
image

And as you can see, it's rendering badly. The options menu is rendering behind the `MessagesModal`. And that's the only one I verified! We use that select in many places, the issue may be in a few of them. I think we should fix this at the `src/Select/index.js` level, so they all work well again. And what we should do is find a different way to fix the original issue of the ticket where you added those changes. Maybe we can send the `menuPortalTarget={document.body}` just from there.

gaboDev:
Ok, so I have re-done the original fix added to [ERA-11311](https://allenai.atlassian.net/browse/ERA-11311).

[ERA-11774](https://allenai.atlassian.net/browse/ERA-11774) was a side effects of it, so by re-doing this bug is not happening anymore.

cc
@luixlive @AlanCalvillo 

[ERA-11311]: https://allenai.atlassian.net/browse/ERA-11311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ERA-11774]: https://allenai.atlassian.net/browse/ERA-11774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

JoshuaVulcan (approved):

luixlive (approved):
This seems like the right approach ✅  Probably you (@gaboDev ) and @AlanCalvillo could still go and double check the places where we render these selects and validate their proper rendering so Jes don't come back with another one 😅 
- `AddReportTab` (reported by Jes)
- `MessagingSelect` (reported by me)
- `PrioritySelect` - you find these in any event, where you choose the priority.
- `ReportedBySelect` - in events too, but also in event filters, patrols plan section and patrol filters.
- Custom `SelectWidget` of v1 schema fields
- Language selector in the side bar

Since we just got back to the original code, they should be all working well, but it doesn't hurt to go and validate to avoid more regresssions.

PR #1323, opened 2025-07-15
Fix dev unit tests after Coordinate Settings spike

luixlive:


src/GpsFormatToggle/index.js
```diff
@@ -22,13 +22,7 @@ const GpsFormatToggle = ({ lat = null, lng = null, name, ref, showGpsString = tr
 
   const gpsFormat = useSelector((state) => state.view.userPreferences.gpsFormat);
   const selectedCRS = useSelector((state) => state.view.coordinateReferenceSystems.selectedSystems);
-  const storedCRS = useSelector(
```
luixlive, line 25: This wasn't really causing tests to fail but it is a bad implemented selector. Every single time the selector is called it returns a new reference (after the reduce operation) causing lots of renders. We needed to memoize the hook by either:
- Using reselect
- Removing the reduce operation from the selector and do it at the component level

I decided to go by the second option since this doesn't seem like responsibility of Redux selectors. We are just fetching a single store value and the transformation can be easily done and kept at the component level.

src/hooks/index.js
```diff
@@ -19,7 +19,7 @@ export const useFeatureFlag = (flagName) => {
     throw new Error('no feature flag with that name exists');
   }
 
-  return featureFlagOverrides.hasOwnProperty(flagName)
+  return featureFlagOverrides?.hasOwnProperty(flagName)
```
luixlive, line 22: A small protection for unit tests, shouldn't cause issues when running the app.

gaboDev (approved):

PR #1328, opened 2025-07-22
Custom Google Analytics for dev

andersondario-er:
### What does this PR do?
- Adds a new pipeline logic for develop branch to inject the value `REACT_APP_GA4_TRACKING_ID` from `.env.development` as build-arg.

### How does it look
- Dockerfile will receive that value as build-arg before the yarn build command. 

### How was it tested ?
1. I changed temporarily the trigger rules to target the PR's branch.
2. Then, I run locally the image created by the job: https://github.com/PADAS/das-web-react/actions/runs/16449778473/job/46491631607#step:8:19978
```sh
docker run --platform linux/amd64 -it --rm europe-west3-docker.pkg.dev/padas-app/er-mt/das-web-react:custom-google-analytics-for-dev-326@sha256:09d00ae7c2a4157e7884802025ed50298c4e069a24d975a126e121eec6af2e77 sh

grep -r "REACT_APP_GA4_TRACKING_ID" /usr/share/nginx/html/ > out.js
docker cp <container_name>:/out.js .
```
3. And was able to verify the code in VSCode Search since it was a bit tricky with sh and grep. It does contain the expected value. 
Screenshot 2025-07-22 at 13 26 27

4. Same result for legacy build. For that I used a test branch, and this is the workflow there: https://github.com/PADAS/das-web-react/actions/runs/16469553118/job/46555100570#step:14:105. I got the same result repeating the steps 2 and 3 with the legacy image. 

### Relevant link(s)
* Tracking tickets: [DASOPS-1770](https://allenai.atlassian.net/browse/DASOPS-1770)

[DASOPS-1770]: https://allenai.atlassian.net/browse/DASOPS-1770?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,7 +23,8 @@ jobs:
 
       - name: Read env file
         id: env-vars
-        if: github.ref_name == 'develop'
+        # if: github.ref_name == 'develop'
+        if: github.ref_name == 'custom-google-analytics-for-dev'
         run: |
           tracking_id=$(grep REACT_APP_GA4_TRACKING_ID .env.development | cut -d '=' -f2)
           echo "build-args=REACT_APP_GA4_TRACKING_ID=$tracking_id" >> $GITHUB_OUTPUT
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,10 +4,13 @@ on:
     branches:
       - develop
       - 'release-**'
+      - custom-google-analytics-for-dev
 
 jobs:
   config:
     runs-on: ubuntu-latest
+    outputs:
+      build-args: ${{ steps.env-vars.outputs.build-args }}
     steps:
       - uses: actions/checkout@v4
       - name: Setup Node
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -17,17 +19,26 @@ jobs:
       - run: |
           npm pkg set "buildbranch"="${{ github.head_ref || github.ref_name }}"
           npm pkg set "buildnum"="${{ github.run_number }}"
+
+      - name: Read env file
+        id: env-vars
+        if: github.ref_name == 'develop'
```
JoshuaVulcan, line 25: This will only work for `develop.pamdas.org` as written, right?

We want this to work for all dev environments, ST and MT -- so develop + anything at *.dev.pamdas.org. 
Since our deployment process uses ERA-* branches for those feature environments, perhaps we can key off that here as well?

andersondario-er: @JoshuaVulcan oh, it will not. `develop.pamdas.org` is located at ST/legacy, I didn't consider ST logic and feature branches. Let me add it, one sec...

andersondario-er:
@JoshuaVulcan now Legacy and MT Preview envs are covered too

JoshuaVulcan (approved):

PR #1335, opened 2025-07-28
ERA-11767: Map View - New Map Layer to show current coordinate reference system

luixlive:
### What does this PR do?
Call new hook `useCrsBoundingBoxLayer` from the `Map` component to render the bounds of the currently selected GPS format when it is a Coordinate Reference System with a defined BBOX.

### How does it look
Marker menu closed:
image

Marker menu open:
image

### Relevant link(s)
* [ERA-11767](https://allenai.atlassian.net/browse/ERA-11767)
* [Env](https://era-11767.dev.pamdas.org/)

[ERA-11767]: https://allenai.atlassian.net/browse/ERA-11767?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/GpsFormatToggle/index.js
```diff
@@ -83,6 +82,7 @@ const GpsFormatToggle = ({ lat = null, lng = null, name = null, ref, showGpsStri
           <label
             className={`${styles.label} ${gpsFormat === gpsFormatOption ? styles.active : ''}`}
             htmlFor={`${gpsFormatOption}-radio`}
+            title={storedCRSMappedByCode[gpsFormatOption]?.name || gpsFormatOption}
```
luixlive, line 85: Adds a title to the GPS format options so users can read the full name of the truncated CRS by hovering the mouse:
image

src/GpsFormatToggle/styles.module.scss
```diff
@@ -21,7 +21,8 @@
   }
 
   .radio {
-    width: 100%;
```
luixlive, line 24: Styles to properly show the long CRS names without breaking the layout, truncating the words and adding ellipsis.

src/GpsFormatTogglePopup/index.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+import GpsFormatToggle from '../GpsFormatToggle';
+
+const GpsFormatTogglePopup = () => <GpsFormatToggle />;
```
luixlive, line 5: Trivial component for the rendering of the GPS format toggle as a Map popup. Right now it doesn't do much but avoids popup-related props to bleed into the GPS format toggle that spreads it in the div and provides its own `styles.module.scss` which is used in the popup. This may be useful later for further customization of the popup.
image

src/Map/index.js
```diff
@@ -109,6 +110,8 @@ const Map = ({ children, onMapLoad, socket }) => {
   const location = useLocation();
   const navigate = useNavigate();
 
+  useCrsBoundingBoxLayer();
```
luixlive, line 113: Add the hook to inject the CRS BBOX source, layer and control marker to the map. Historically, layers have been injected as renderless components, but here I follow the recent pattern we've been adding to our application in later layers through a hook that properly handles the injection of resources (sources, layers, markers, popups) to the map and clean the resources when they are not necessary anymore.

src/Map/layers/useCrsBoundingBoxLayer/index.js
```diff
@@ -0,0 +1,155 @@
...
+  containerElement.appendChild(nameElement);
+
+  const caretElement = document.createElement('div');
+  injectStylesToElement(caretElement, CRS_BBOX_CONTROL_MARKER_STYLES.CARET(isControlMenuOpen));
+  containerElement.appendChild(caretElement);
+
+  return containerElement;
+};
+
+const useCrsBoundingBoxLayer = () => {
```
luixlive, line 66: This hook reads the currently selected GPS format from the store, and if it is a CRS with a BBOX that covers over 85% of the max span of both latitude and longitude, we provide the BBOX as the data of a map source and a layer to render the square. We also add the marker control with the CRS name and a click handler to show and hide the GPS toggle popup:
image

src/SideBar/SettingsPane/MapTab/CoordinateSystemSettingsView/GpsFormatSelector/index.js
```diff
@@ -28,6 +30,12 @@ const CrsGpsFormatOption = ({ epsgCode, name }) => {
 
   const onCheckboxChange = () => {
     if (isChecked) {
+      if (gpsFormat === epsgCode) {
```
luixlive, line 33: Fixes a bug when unchecking or removing a CRS from our stored systems in the Coordinate System Settings view that is currently selected as the GPS format. Now if the user removes the selected GPS format from the options, we automatically switch to DEG which is always available.

src/utils/location/index.js
```diff
@@ -229,9 +229,17 @@ export const getProj4CompatibleCRS = async () => {
         // and that do not require grid shift files so they can be transformed
         // by the proj4 library.
         if (coordinateReferenceSystem.proj4 && !needsGridShiftFile) {
+          let bbox = coordinateReferenceSystem.bbox;
+          if (Array.isArray(bbox) && bbox.length === 4) {
```
luixlive, line 233: Seems like `epsg-index` and `turf` disagree on the ordering of the parts of a bbox. Latitudes and longitudes from our coordinate reference systems come inverted from what Turf expects so here we reorder them.

ctrier:
Looks great! One small thing: I wonder if it’s possible to provide a bit more contrast between the grey of the deselected options and the white background of the selected option? It’s kind of hard to see and maybe a slightly darker grey would make it more clear
Screenshot 2025-08-05 at 1 00 36 PM

tiffanynwong:
> Looks great! One small thing: I wonder if it’s possible to provide a bit more contrast between the grey of the deselected options and the white background of the selected option? It’s kind of hard to see and maybe a slightly darker grey would make it more clear Screenshot 2025-08-05 at 1 00 36 PM

I have a future design for location search where I added a blue underline under the selected tab. This may help and is consistent with the tab selection we have in settings. 
Screenshot 2025-08-05 at 1 22 37 PM

This is a small thing, but if I'm too zoomed into the map it's hard to tell that anything has changed when I change the coordinate system. Maybe that's ok but it might be nice if we could zoom the user out to the extent of the blue boundary to confirm they've made the change. Think this is a nice to have but not needed if it's too complicated, since these areas do change in size.

luixlive:
> This is a small thing, but if I'm too zoomed into the map it's hard to tell that anything has changed when I change the coordinate system. Maybe that's ok but it might be nice if we could zoom the user out to the extent of the blue boundary to confirm they've made the change. Think this is a nice to have but not needed if it's too complicated, since these areas do change in size.

Actually it's not hard, but it's something I'm hesitant about. I feel like as a user, I could get annoyed if I'm at one point of interest of the map and by changing the coordinate system the map moves away. The idea of changing the system is just to update the format of the locations, I feel like moving the map is a bit invasive and could affect the user experience if they were doing something in the map but just wanted to switch the GPS format 🤔 

Remember users can change the GPS format from many places, for example from an event detail view from the location widget. It may be unexpected to automatically move the map if all they wanted was to change the format in which the location is displayed.

Anyway it's doable so if we decide we want it, I can add that functionality @tiffanynwong .

tiffanynwong:
Really good points! Ok let's leave it as is :)

luixlive:
GPS format toggle updates:
image

Just waiting for a dev approval @JoshuaVulcan @gaboDev

src/Map/layers/useCrsBoundingBoxLayer/index.js
```diff
@@ -0,0 +1,162 @@
...
+    cursor: 'pointer',
+    display: 'flex',
+    flexDirection: 'row',
+    fontSize: '0.875rem',
+    padding: '0.25rem',
+  },
+};
+
+// Axis thresholds are 85%.
+const LATITUDE_SPAN_THRESHOLD = 180 * 0.85;
```
gaboDev: Shouldn't the 0.85 threshold be on its own const?

src/Map/layers/useCrsBoundingBoxLayer/index.js
```diff
@@ -0,0 +1,162 @@
...
+const LONGITUDE_SPAN_THRESHOLD = 360 * 0.85;
+
+const createCrsBboxControlMakerElement = (crsName, isControlMenuOpen, onClick) => {
+  const containerElement = document.createElement('div');
+  containerElement.onclick = onClick;
+  injectStylesToElement(containerElement, CRS_BBOX_CONTROL_MARKER_STYLES.CONTAINER);
+  // Hover effect.
+  containerElement.onmouseenter = () => {
+    containerElement.style.backgroundColor = '#005fc0';
```
gaboDev: Also, is this color value used anywhere else? wouldn't be worth it to take it to its own color const like `BRIGHT_BLUE`?

gaboDev (approved):
Just left really nit comments, LGTM 🚀

src/Map/layers/useCrsBoundingBoxLayer/index.js
luixlive: I like the idea, more explicit 👍

src/Map/layers/useCrsBoundingBoxLayer/index.js
luixlive: That was my intention! I forgot to do it haha.

Giving some context, in recent designs I've been using the new SASS version functions (thanks for the Create React App eject 🙏 ), specifically for hover effects I use the [color.adjust(color, $lightness: -5%);](https://sass-lang.com/documentation/modules/color/#adjust). So I thought it would be a good idea to see what the corresponding hex for `BRIGHT_BLUE` after that color adjustment was, which is this one, and set it just for sake of consistency. I will just wrap this string in its own constant for clarity.

src/utils/location/index.js
JoshuaVulcan: The LngLat/LatLng battle is as old as time 😩

JoshuaVulcan (approved):
Looks good, works well, nice work 👍

PR #1340, opened 2025-07-31
ERA-11828: Polygon Edit Drawer crashes the site

gaboDev:
### What does this PR do?
- Adds security check when `event` object turns to null in a `ReportGeometryDrawer` useEffect.

### Relevant link(s)
* Tracking tickets: [ERA-11828](https://allenai.atlassian.net/browse/ERA-11828)
* [Hosted demo environments](https://era-11828.dev.pamdas.org)

[ERA-11828]: https://allenai.atlassian.net/browse/ERA-11828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ReportGeometryDrawer/index.js
+++ b/src/ReportGeometryDrawer/index.js
@@ -141,7 +141,7 @@ const ReportGeometryDrawer = () => {
   }, [isDrawing, isGeometryAValidPolygon, onCancel, onUndo]);
 
   useEffect(() => {
-    if (event?.geometry) {
+    if (event?.geometry && isDrawing) {
       const eventPolygon = event.geometry.type === 'FeatureCollection'
         ? event.geometry.features[0]
         : event.geometry;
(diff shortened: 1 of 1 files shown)


src/ReportGeometryDrawer/index.js
```diff
@@ -152,7 +152,7 @@ const ReportGeometryDrawer = () => {
       setTimeout(() => dispatchReportGeometry(reset()));
       setIsDrawing(false);
     }
-  }, [event.geometry, map]);
+  }, [event?.geometry, map, isDrawing]);
```
gaboDev, line 155: This `event` object is coming from this:
`const event = useSelector((state) => state.view.mapLocationSelection.event); `

There are renders where `mapLocationSelection.event` is set to null, I see that `<DetailsSection />` it is setting this event object to null when being unmounted, you can see it [here](https://github.com/PADAS/das-web-react/blob/d6e477056992fb2c5fca8ec87f77f8faacb3d206/src/ReportManager/DetailsSection/index.js#L132), anyways after it the event is properly set so We just needed to add the security check.

JoshuaVulcan (approved):

PR #1342, opened 2025-08-01
ERA-11828: Polygon Edit Drawer crashes the site

gaboDev:
### What does this PR do?
- Adds security check when `event` object turns to null in a `ReportGeometryDrawer` useEffect.
- Solves [ERA-11289](https://allenai.atlassian.net/browse/ERA-11289)

### Relevant link(s)
* Tracking tickets: [ERA-11828](https://allenai.atlassian.net/browse/ERA-11828)
* [Hosted demo environments](https://era-11828.dev.pamdas.org)

[ERA-11828]: https://allenai.atlassian.net/browse/ERA-11828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ERA-11289]: https://allenai.atlassian.net/browse/ERA-11289?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/MapLocationSelectionOverview/styles.module.scss
+++ b/src/MapLocationSelectionOverview/styles.module.scss
@@ -17,7 +17,7 @@
   @media (min-width: $md-layout-width-min) {
     left: 1rem;
     top: 2rem;
-    width: 25rem;
+    width: auto;
   }
 }
 
@@ -62,10 +62,6 @@
   border-bottom: 1px solid #DDD;
   margin: 0.5rem 0 0.5rem -0.5rem;
   width: calc(100% + 1rem);
-
-  @media (min-width: $md-layout-width-min) {
-    width: 25rem;
-  }
 }
 
 .measurements {
--- a/src/ReportGeometryDrawer/index.js
+++ b/src/ReportGeometryDrawer/index.js
@@ -7,7 +7,7 @@ import { LAYER_IDS } from '../MapDrawingTools/MapLayers';
 import { MapContext } from '../App';
 import { MapDrawingToolsContext } from '../MapDrawingTools/ContextProvider';
 import reportGeometryReducer, { reset, setGeometryPoints, undo } from '../ducks/report-geometry';
-import { setIsPickingLocation } from '../ducks/map-ui';
+import { setIsPickingLocation, setMapLocationSelectionEvent } from '../ducks/map-ui';
 import { useMapEventBinding } from '../hooks';
 import { validateEventPolygonPoints } from '../utils/geometry';
 
@@ -81,6 +81,7 @@ const ReportGeometryDrawer = () => {
       } else {
         setMapDrawingData(null);
         dispatch(setIsPickingLocation(false));
+        dispatch(setMapLocationSelectionEvent(null));
       }
     }
   }, [dispatch, event?.geometry, points, setMapDrawingData, showInformationModal]);
@@ -107,6 +108,7 @@ const ReportGeometryDrawer = () => {
 
   const onSaveGeometry = useCallback(() => {
     dispatch(setIsPickingLocation(false));
+    dispatch(setMapLocationSelectionEvent(null));
   }, [dispatch]);
 
   const onHideCancellationConfirmationModal = useCallback(() => setShowCancellationConfirmationModal(false), []);
--- a/src/ReportGeometryDrawer/index.js
+++ b/src/ReportGeometryDrawer/index.js
@@ -152,7 +152,7 @@ const ReportGeometryDrawer = () => {
       setTimeout(() => dispatchReportGeometry(reset()));
       setIsDrawing(false);
     }
-  }, [map]);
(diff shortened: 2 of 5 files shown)


src/MapLocationSelectionOverview/styles.module.scss
```diff
@@ -17,7 +17,7 @@
   @media (min-width: $md-layout-width-min) {
```
gaboDev, line 17: All these styles changes are the fixes for [ERA-11828](https://allenai.atlassian.net/browse/ERA-11828)

image

src/ReportGeometryDrawer/index.js
```diff
@@ -141,18 +141,18 @@ const ReportGeometryDrawer = () => {
...
 
-      map.fitBounds(bbox(eventPolygon), { padding: VERTICAL_POLYGON_PADDING });
+      map.fitBounds(bbox(eventPolygon), { padding: VERTICAL_POLYGON_PADDING });
 
       dispatchReportGeometry(setGeometryPoints(eventPolygon.geometry.coordinates[0].slice(0, -1)));
       setTimeout(() => dispatchReportGeometry(reset()));
       setIsDrawing(false);
     }
-  }, [event?.geometry, map, isDrawing]);
+  }, [event?.geometry, map]);
```
gaboDev, line 157: I reverted this changes, only left the security check for the `event.geometry` in the deps array

src/ReportManager/DetailsSection/index.js
```diff
@@ -129,8 +133,12 @@ const DetailsSection = ({
   useEffect(() => {
     dispatch(setMapLocationSelectionEvent(reportForm));
 
-    return () => dispatch(setMapLocationSelectionEvent(null));
-  }, [dispatch, reportForm]);
+    return () => {
+      if (!isDrawingEventGeometry){
```
gaboDev: The issue was related with setting to null the `mapLocationSelectionEvent` in the store, causing multiple re-renders/crashes on `ReportGeometryDrawer`, now this object will be set to null only when there is no map drawing occurring

luixlive:
@gaboDev could you provide more context on the root cause of the bug? I don't really understand it, which makes hard to get why some of the code changes you did. I feel like you are updating old code that had been resilient for a while, why did it start failing? The previous PR injected a bug, and I don't want to happen again so we should be very clear on where the bug is and what the right fix is.

gaboDev:
> @gaboDev could you provide more context on the root cause of the bug? I don't really understand it, which makes hard to get why some of the code changes you did. I feel like you are updating old code that had been resilient for a while, why did it start failing? The previous PR injected a bug, and I don't want to happen again so we should be very clear on where the bug is and what the right fix is.

I was not able to trace the origin of why before this was not causing issues, but now every time I create a new event with a polygon I see this issue happening.

The issue was related with a `useEffect` in `DetailsSection` where the `setMapLocationSelectionEvent` was being dispatched, this effect had as dep an event object coming directly from the event store, which was causing multiple re-renders while setting `setMapLocationSelectionEvent` to a proper object in the effect logic and then to null in the clean-up callback of the effect.

The original bug was originated from trying to access the property `geometry` of this object mentioned `event` in `ReportGeometryDrawer` when the event was `null`.

The first approached I followed was to conditionally set to null `setMapLocationSelectionEvent` based on if the user was drawing in the map, but this provoke side effects.

What I did as second approach is to remove completely this effect, I feel like this assignation is not really an effect, I moved that logic to the proper places where `setMapLocationSelectionEvent` needed to be set which is only when pressing the edit button of the `GeometryPreview` and saving/canceling the edition of the polygon in the drawing screen.


cc.
@luixlive @JoshuaVulcan

luixlive (approved):
The code changes make sense ✅  

I'd just like to mention that I think the previous approach was better. The `useEffect` was attached to the lifecycle of the `DetailsSection` component, which is driven by our deep linking mechanism of events `/events/:eventId`. Meaning that whenever we had an event open, we used to set it as the `mapLocationSelectionEvent`, and as soon as we leave the event route (unmount the `DetailsSection`) we cleared the store variable. The  dependency array of the `useEffect` only had `dispatch`, whose reference never changes, and `reportForm`, which makes perfect sense because we want to update the `mapLocationSelectionEvent` whenever the selected event changes. That was pretty straightforward. 

The new approach seems to fix the error, but is harder to maintain because now we need to set the `mapLocationSelectionEvent` from more places and we need to clear it in more places too. If tomorrow we add a new mechanism to open the drawer or to close it, we will need to remember to dispatch the action or things could go bad.

We need to go forward with it, so I'm approving since we couldn't find the root cause of the issue, but just something to think about.

PR #1344, opened 2025-08-08
ERA-11803: Split message API calls

luixlive:
### What does this PR do?
Add logic to our `fetchAllMessages` action creator to split message requests in chunks of 25 subject ids at the same time to avoid server errors with enormous requests.

Also, this PR fixes the UI of our `GPSFormatToggle` in the subjects map popup.

### Relevant link(s)
* [ERA-11803](https://allenai.atlassian.net/browse/ERA-11803)
* [Env](https://era-11803.dev.pamdas.org/)

In order to test this in the environment, I needed to create 25+ subjects so we can see the requests being broken. Since that would take me much time, and I know @AlanCalvillo has scripts to generate subjects in bulk, I didn't do that.

Instead, what I recommend to the dev reviewers (@gaboDev @JoshuaVulcan ) is to checkout to this branch locally (pull the changes) and run your local server against https://era-11803.dev.pamdas.org/ . Then, go to `src/ducks/messaging.js` and modify the constant `MAXIMUM_SUBJECT_IDS_PER_REQUEST` to have a value of 3. That way, you can easily navigate the map to Madagascar and see all the message requests of the subjects I created being broken in several network calls in chunks of 2 subjects:
image

[ERA-11803]: https://allenai.atlassian.net/browse/ERA-11803?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/GpsFormatToggle/styles.module.scss
+++ b/src/GpsFormatToggle/styles.module.scss
@@ -35,7 +35,7 @@
       display: block;
       height: 100%;
       overflow: hidden;
-      padding: 0.25rem;
+      padding: 0.25rem 0;
       text-align: center;
       text-overflow: ellipsis;
       white-space: nowrap;
--- a/src/ducks/messaging.js
+++ b/src/ducks/messaging.js
@@ -7,8 +7,9 @@ import { API_URL } from '../constants';
 
 import { objectToParamString, recursivePaginatedQuery } from '../utils/query';
 
-
 import { messageIsValidForDisplay } from '../utils/messaging';
+
+const MAXIMUM_SUBJECT_IDS_PER_REQUEST = 25;
 
 const FETCH_MESSAGES_SUCCESS = 'FETCH_MESSAGES_SUCCESS';
 const UPDATE_UNREAD_MESSAGES_COUNT = 'UPDATE_UNREAD_MESSAGES_COUNT';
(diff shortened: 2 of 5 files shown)


src/GpsFormatToggle/styles.module.scss
```diff
@@ -35,7 +35,7 @@
       display: block;
       height: 100%;
       overflow: hidden;
-      padding: 0.25rem;
```
luixlive, line 38: After the recent changes of the `GPSFormatToggle` component, the widget renders the GPS options with ellipsis in the subjects popup due to the limit of space:
image

In the two .scss files modified in this PR I did some tweaks to keep the nice UI we have everywhere, while optimizing the space in the subjects popup:
image

src/ducks/messaging.js
```diff
@@ -7,9 +7,10 @@ import { API_URL } from '../constants';
 
 import { objectToParamString, recursivePaginatedQuery } from '../utils/query';
 
-
```
luixlive, line 10: And here is the magic to break the message requests. I added a couple comments, so I don't think I need to explain more here. The code is mostly self-explanatory and the comments add a bit of context 😄

gaboDev (approved):
Smooth ✅

JoshuaVulcan:
Just ran this locally against a couple affected sites and confirmed this to be working as expected.

PR #1345, opened 2025-08-11
ERA-11801: Choices/choice lists - Add type safety to ER Web form controls to handle string AND array values

gaboDev:
### What does this PR do?
- Adds parseValue to `CheckboxesWidget` component to support string values

### Relevant link(s)
* Tracking tickets: [ERA-11801](https://allenai.atlassian.net/browse/ERA-11801)
* [Hosted demo environments](https://era-11801.dev.pamdas.org)

[ERA-11801]: https://allenai.atlassian.net/browse/ERA-11801?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/SchemaFields/index.js
+++ b/src/SchemaFields/index.js
@@ -20,6 +20,7 @@ import DateTimePicker, { EMPTY_DATE_TIME_VALUE } from '../DateTimePicker';
 import Select from '../Select';
 
 import * as styles from './styles.module.scss';
+import { isArray } from 'lodash-es';
 
 const eventReportTracker = trackEventFactory(EVENT_REPORT_CATEGORY);
 
@@ -388,6 +389,22 @@ const selectValue = (value, selected, all) => {
 
 const deselectValue = (value, selected) => selected.filter((v) => v !== value);
 
+const parseValue = (value) => {
+  if (isArray(value)){
+    return value.map((val) => isPlainObject(val) ? val.value : val);
+  }
+
+  if (isPlainObject(value)){
+    return [value.value];
+  }
+
+  if (typeof value === 'string' && value.length > 0){
+    return  [value];
+  }
+
+  return [];
+};
+
 export const CheckboxesWidget = ({
   autofocus,
   disabled,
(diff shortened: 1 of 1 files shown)


src/SchemaFields/index.js
```diff
@@ -388,6 +388,22 @@ const selectValue = (value, selected, all) => {
 
 const deselectValue = (value, selected) => selected.filter((v) => v !== value);
 
+const parseValue = (value) => {
```
luixlive: This file is big and it has many components. If we are going to add file-level utilities for some of these components, I would just recommend to add explicit names so it's easy to understand what the scope of these methods are: `parseCheckboxesWidgetValue`.

src/SchemaFields/index.js
```diff
@@ -388,6 +388,22 @@ const selectValue = (value, selected, all) => {
...
+  if (Array.isArray(value)){
+    return value.map((val) => isPlainObject(val) ? val.value : val);
```
luixlive, line 393: Hmm are we 100% sure that `rjsf` expects an array of strings? Because I see that now if the value is an array of objects, you transform it to an array of strings, we weren't doing that before. If we can make sure that this works by testing some real scenarios, it would be great.

luixlive (dismissed):
Exactly what I had in mind! Well done ✅ 

Now we are always using an array of string, so we need to be 100% sure that data from old events in our prod sites work well with this new forced parsing. Because I'm pretty sure some of them store the values as an array of objects that now we are enforcing to be an array of strings. @AlanCalvillo

src/SchemaFields/index.js
gaboDev: We were doing it before, I just moved that map under the correct condition 👁️

src/SchemaFields/index.js
gaboDev: Sure!

JoshuaVulcan (approved):

PR #1346, opened 2025-08-12
Hotfix: GPS format toggle popups

luixlive:
### What does this PR do?
Fix the GPS format toggle UI in map popups by adding a new `size` prop to the component it renders itself in a small size that fits better in smaller layouts.

### How does it look
Small size, optimal for map popups:
image
image
image

Still looking the same for normal size:
image

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -404,6 +404,14 @@ div.mapboxgl-popup-content {
     margin-top: 0.5rem;
     margin-bottom: 0.5rem;
   }
+
+  .gps-format-toggle {
+    fieldset {
+      label {
+        font-size: 0.6875rem;
+      }
+    }
+  }
 }
 
 div.mapboxgl-popup-tip {
--- a/src/GpsFormatToggle/index.js
+++ b/src/GpsFormatToggle/index.js
@@ -14,15 +14,13 @@ import * as styles from './styles.module.scss';
 
 const gpsFormatTracker = trackEventFactory(GPS_FORMAT_CATEGORY);
 
-const SIZES = { NORMAL: 'normal', SMALL: 'small' };
-
 const GpsFormatToggle = ({
+  className = '',
   lat = null,
   lng = null,
   name = null,
   ref,
   showGpsString = true,
-  size = SIZES.NORMAL,
   ...otherProps
 }) => {
   const customCoordinateSystemsEnabled = useFeatureFlag(FEATURE_FLAG_LABELS.CUSTOM_COORDINATE_SYSTEMS_ENABLED);
@@ -68,7 +66,7 @@ const GpsFormatToggle = ({
     });
   }, []);
 
-  return <div {...otherProps}>
+  return <div className={`gps-format-toggle ${className}`} {...otherProps}>
     <fieldset className={styles.fieldset} ref={fieldsetRef} role="radiogroup">
       <legend className={styles.legend}>{t('fieldsetLegend')}</legend>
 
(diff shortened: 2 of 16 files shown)


src/GpsFormatToggle/index.js
```diff
@@ -14,7 +14,17 @@ import * as styles from './styles.module.scss';
 
 const gpsFormatTracker = trackEventFactory(GPS_FORMAT_CATEGORY);
 
-const GpsFormatToggle = ({ lat = null, lng = null, name = null, ref, showGpsString = true, ...otherProps }) => {
+const SIZES = { NORMAL: 'normal', SMALL: 'small' };
```
gaboDev: it would be idea to export this variable and change its name to something like `GPS_FORMAT_TOGGLE_SIZES` and use it instead of regular strings

gaboDev (approved):

JoshuaVulcan (commented):
This should be the default font-size for this control inside of any map pop-up, right?

With that in mind, seems like the PR would be a couple lines of CSS instead of touching 12 files and a new prop.  Could you help me understand why you chose this approach?

src/GpsFormatToggle/index.js
luixlive: Good idea 👍

luixlive:
> This should be the default font-size for this control inside of any map pop-up, right?
> 
> With that in mind, seems like the PR would be a couple lines of CSS instead of touching 12 files and a new prop. Could you help me understand why you chose this approach?

Well, in my experience with this app global styles have made what would sound like simple changes a real pain before. I understand the point, this PR would be smaller, but in my opinion it's better to keep the CSS of each component scoped in its file or modified directly via props (like `className` or in this case `size`). It feels more specific, easier to debug and more flexible if we want different styles in different popups. It's nice to have a popup class that wraps all map popups to add generic styles like background color, borders, shadows, etc... Basically anything related to what would be our "app theme". But going deep into the specifics of a component feels a bit nasty for me.

But that's just my opinion, and it's not a big deal for me. I'm ok closing this PR and sending that alternative solution @JoshuaVulcan

src/App.scss
```diff
@@ -404,6 +404,14 @@ div.mapboxgl-popup-content {
     margin-top: 0.5rem;
     margin-bottom: 0.5rem;
   }
+
+  .gps-format-toggle {
+    fieldset {
+      label {
+        font-size: 0.6875rem;
```
luixlive, line 411: Global style so all `div.mapboxgl-popup-content` that contain a `gps-format-toggle` render smaller labels.

src/GpsFormatToggle/index.js
```diff
@@ -58,7 +66,7 @@ const GpsFormatToggle = ({ lat = null, lng = null, name = null, ref, showGpsStri
     });
   }, []);
 
-  return <div {...otherProps}>
+  return <div className={`gps-format-toggle ${className}`} {...otherProps}>
```
luixlive, line 69: Added a regular string as a class name so the global style can be easily applied.

src/GpsFormatToggle/styles.module.scss
```diff
@@ -35,7 +35,7 @@
       display: block;
       height: 100%;
       overflow: hidden;
-      padding: 0.25rem;
+      padding: 0.25rem 0;
```
luixlive: Extra style tweaks for the `GpsFormatToggle` to look nice in several sizes.

src/SubjectPopup/styles.module.scss
```diff
@@ -70,40 +70,7 @@ $background_image_url: '../common/images/icons/';
 }
 
 .gpsFormatToggle {
-  background-color: $very-light-gray;
```
luixlive: Old unnecessary and a bit weird styles. Our `SubjectPopup` looks good and more consistent with other popups without this:
image

luixlive:
Updated following @JoshuaVulcan 's suggestion on adding a global style for all our map popups 👍 cc: @gaboDev

JoshuaVulcan (approved):
Approved!

PR #1350, opened 2025-08-14
ERA-11770: Update location picker to read the projection from redux 

luixlive:
### What does this PR do?
`LocationPicker`, `GpsInput` and `GpsFormatToggle` are now aware of the coordinate reference systems and show coordinates in the right formats, error states, tooltips, etc.

Lot's of files were minimally touched, specially because after the last couple weeks I have improved my understanding about the differences between coordinate formats, coordinate systems, EPSG codes, projections, etc... And I have decided to do a few variable and function naming tweaks to better reflect what each method and store variable does. There are still a few improvements that could be done, but some of them affect so much the app that I guess we will have to live with them, like our constant that appear all over the app: `gpsFormat`s. GPS is a satellite system to resolve locations of devices, but it has nothing to do with how coordinates are shown. Still, there are other few tweaks I could do but I will do them in the following PRs since this is big enough.

### How does it look
`LocationPicker` showing coordinates in a CRS when the location falls within the bounds:
image

`LocationPicker` fallbacking to degrees and showing a warning tooltip and an error state in the GPS input when the location falls outside the bounds of the CRS:
image

image

### Relevant link(s)
* [ERA-11770](https://allenai.atlassian.net/browse/ERA-11770)
* [Env](https://era-11770.dev.pamdas.org/)

### Additional info
There are many things that were changed for the location picker to show the new coordinate reference systems but I'll mention them in code comments.

Initially I wanted to disable the options from the `GpsFormatToggle` radiogroup that couldn't show the location, but then I got philosophical and decided it is better to let the user change the coordinates representation (aka GPS format) freely. If the purpose of our widget was only to show a location in a given format, it would make sense to prohibit the user from choosing a system that cannot display it. But our widget capabilities go beyond showing the location, it also changes the coordinates representation across the app. And all the places that read the representation are syncrhonized, so even if we disable the options from the widget, the user can easily change the representation from another place (like the map control at the top right), so it would be weird to prohibit the change from some places but not from others. And we definitely don't want to prohibit the change globally just because in a popup the user is seeing a location that can't be shown in a specific system. What I did instead is leave the options always enabled, but follow the designs to add warnings, error states and discouraging the user from switching to a representation that can't show the current location:
image

[ERA-11770]: https://allenai.atlassian.net/browse/ERA-11770?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/App.scss
```diff
@@ -843,4 +843,16 @@ div.mapboxgl-popup-tip {
   }
 }
 
+.sr-only {
```
luixlive, line 846: Global class so instead of adding this definition in every `styles.module.scss` that requires a screen reader only element, we just add `className="sr-only"` 😄

src/CursorGpsDisplay/MenuPopover/index.test.js
```diff
@@ -33,7 +33,7 @@ describe('CursorGpsDisplay - MenuPopover', () => {
       data: {},
       view: {
         coordinateReferenceSystems: {
-          selectedSystems: Object.values(GPS_FORMATS),
+          selectedCoordinateRepresentations: Object.values(GPS_FORMATS),
```
luixlive, line 36: The main reason why this PR touches so many files is that I decided to do some naming updates to a few variables and methods. After the last few tickets, I have improved my understanding of certain concepts, and the differences between coordinate reference system, projection, coordinate format, GPS, etc... And I have noticed that we could do a few tweaks to have a better defined ubiquitous language. So I decided to update at least the variables that I added myself in the first PRs of this Coordinate Systems epic.

The first change in naming we have here: `selectedSystems` -> `selectedCoordinateRepresentations`. You will see this one many times. This variable holds an array with the currently selected CRS or GPS formats that the user can toggle between. Thea array can hold values like: `DEG`, `UTM`, or a CRS EPSG code like `5369`. So we need name that could properly encompass what those things are. Initially I thought "systems" is a good word, but now I realise that "DEG", "DMS" or "UTM" are not really systems, but coordinate formats. So we could hold coordinate formats and coordinate systems here, and at the end, what we do with them is just a way to represent the coordinates accross the app in those formats. So I decided `coordinateRepresentations` is a better way to call this variable.

src/CursorGpsDisplay/index.js
```diff
@@ -6,7 +6,7 @@ import { useTranslation } from 'react-i18next';
 
 import { ReactComponent as SearchIcon } from '../common/images/icons/search-icon.svg';
 
-import { transformLngLatToLocationType, validateLocation } from '../utils/location';
+import { stringifyCoordinates, validateLocation } from '../utils/location';
```
luixlive, line 9: Another change you will see often, I decided to rename the two utility methods we use to transform from any coordinate representation string to a `lngLat` object and viceversa. I decided to use very standard names that scream their intention:
- `stringifyCoordinates(lngLat, representation = GPS_FORMATS.DEG)` returns a readable string in the desired coordinates representation.
- `parseCoordinates(coordinatesString, representation = GPS_FORMATS.DEG)` returns a lngLat object.

src/DroppedMarkerPopup/index.js
```diff
@@ -14,8 +14,7 @@ const DroppedMarkerPopup = ({ data: { location }, id }) => {
 
   return <>
     <GpsFormatToggle
-      lat={location.lat}
```
luixlive, line 17: Now the GPS Format Toggle component follows the standard we have in the rest of the app of receiving a `lngLat` object instead of separate `lng` and `lat` values.

src/GpsFormatToggle/index.js
```diff
@@ -16,39 +24,58 @@ const gpsFormatTracker = trackEventFactory(GPS_FORMAT_CATEGORY);
 
 const GpsFormatToggle = ({
   className = '',
-  lat = null,
-  lng = null,
+  lngLat = null,
   name = null,
   ref,
-  showGpsString = true,
+  showCoordinates = true,
```
luixlive, line 30: I haven't gone deep on this matter, but I want to remove as much as possible the acronym "GPS" from our variable names. GPS is a satellite-navigation system, but it's not related at all to EarthRanger data. EarthRanger doesn't know or care how we received the coordinates we store in the system. They could come from a GPS device, or they could have been directly written, it's all the same for us. Calling our coordinates, `lngLat` objects, coordinate formats, etc... GPS, is simply wrong, and it could cause confusion. I didn't change it everywhere since we have it in looots of places, but at least I did in some of the components I touched, like here. This prop is a boolean to decide if we want to show a coordinates string, GPS is simply not the right word.

src/GpsFormatToggle/index.js
```diff
@@ -16,39 +24,58 @@ const gpsFormatTracker = trackEventFactory(GPS_FORMAT_CATEGORY);
...
+  // add a fallback in case the implementator didn't provide one.
   const nameFallback = useId();
 
-  const gpsString = (showGpsString && lat !== null && lng !== null)
-    ? transformLngLatToLocationType({ latitude: lat, longitude: lng }, gpsFormat)
-    : null;
-
-  const gpsFormatOptions = customCoordinateSystemsEnabled ? selectedCRS : Object.values(GPS_FORMATS);
+  const gpsFormatOptions = customCoordinateSystemsEnabled
+    ? selectedCoordinateRepresentations.sort((optionA, optionB) => {
```
luixlive, line 56: Sort our `gpsFormatOptions` alphabetically as promised (again GPS is wrong, abut this widget sets the Redux gpsFormat variable that is so widely used that I didn't want to remove it, so I left the naming here intact).
image

src/GpsFormatToggle/index.js
```diff
@@ -16,39 +24,58 @@ const gpsFormatTracker = trackEventFactory(GPS_FORMAT_CATEGORY);
...
+      // Sort coordinate representation options alphabetically. If they are a
+      // CRS, we use the name property, otherwise we simply use the GPS format
+      // string.
+      const optionAName = storedCoordinateReferenceSystemsMappedByCode[optionA]?.name || optionA;
+      const optionBName = storedCoordinateReferenceSystemsMappedByCode[optionB]?.name || optionB;
+      return optionAName > optionBName ? 1 : -1;
+    })
+    : Object.values(GPS_FORMATS);
+
+  const { areCoordinatesOutsideCrsBbox, coordinatesString } = useMemo(() => {
```
luixlive, line 66: Calculate the coordinates string, fallbacking to DEG if the coordinates fall outside the representation BBOX.

src/GpsFormatToggle/index.js
```diff
@@ -88,24 +124,56 @@ const GpsFormatToggle = ({
           />
 
           <label
-            className={`${styles.label} ${gpsFormat === gpsFormatOption ? styles.active : ''}`}
+            className={`${styles.label} ${gpsFormat === gpsFormatOption ? styles.active : ''} ${isLngLatOutsideCrsBbox ? styles.invalid : ''}`}
```
luixlive, line 125: Style the labels in red if the `lngLat` provided is outside of their BBOX:
image

Notice something important, you can provide a `lngLat` object and setting `showCoordinates` as false. It may sound useless, but it is not. Maybe you don't want to show the coordinates string, but still want to format the invalid labels, like we do in our location picker!
image

There, this component (GpsFormatToggle) doesn't show the coordinates because the picker already has an input rendering them, but we still want to show the invalid representations painting their labels red.

src/GpsFormatToggle/index.js
```diff
@@ -88,24 +124,56 @@ const GpsFormatToggle = ({
...
-        className={styles.textCopyButton}
-        text={gpsString}
-        title={t('textCopyButtonLabel')}
-      />
+    {coordinatesString && <div className={styles.coordinatesStringWrapper}>
+      <span aria-describedby={coordinatesOutsideBboxTooltipId} className={styles.coordinatesString}>
+        {coordinatesString}
+      </span>
+
+      {areCoordinatesOutsideCrsBbox
```
luixlive, line 140: Render the warning icon button tooltip if the user selects a representation where the current `lngLat` is not supported (outside of the BBOX):
image

This is totally accessible since we render a screen reader only paragraph used as the accessible description of the coordinates string.

src/GpsFormatTogglePopup/styles.module.scss
```diff
@@ -10,5 +10,13 @@
     background: white;
     padding: 0.5rem;
     width: auto;
+
+    [class*=gps-format-toggle] {
+      fieldset {
+        label {
+          font-size: 0.75rem;
```
luixlive, line 18: Recently, we decided to add a global style to decrease the font size of the labels in the GPS format toggles that are within a map popup. However, in this particular map popup we want to keep the original size:
image

src/GpsInput/index.js
```diff
@@ -25,101 +27,115 @@ const GpsInput = ({
...
-  // The input value is handled locally to accept any input, we just trigger onChange when the input value is valid.
-  const [inputValue, setInputValue] = useState(value
-    ? transformLngLatToLocationType(value, gpsFormat)
+  // The input value is handled locally to accept any input, but we just
+  // trigger onChange when the value is valid.
+  const [inputValue, setInputValue] = useState(() => value
+    ? stringifyCoordinates(value, coordinatesRepresentation)
     : '');
-  const [isValid, setIsValid] = useState(true);
+  const [isInputValueValid, setIsInputValueValid] = useState(true);
```
luixlive, line 42: Now we have two ways to set the input as invalid: if the value itself is not a valid coordinates string or if the current coordinates from the value fall outside the BBOX of the coordinates representation (CRS). So we rename this variable state to properly reflect the fact that it only tracks if the value is valid, not if the whole input is valid.

src/GpsInput/index.js
```diff
@@ -25,101 +27,115 @@ const GpsInput = ({
...
         ref={(element) => {
           if (inputRef) {
             inputRef.current = element;
           }
           innerInputRef.current = element;
         }}
         type="text"
-        value={inputValue}
+        // If the value is outside of the BBOX of the CRS, show N/A.
+        value={inputValue === OUTSIDE_BBOX ? t('outsideBboxInputValue') : inputValue}
```
luixlive, line 118: Here an example of the GPS Input component in error state due to the coordinates being outside the supported BBOX by the current CRS selected as the coordinates representation:
image

src/GpsInput/index.js
```diff
@@ -25,101 +27,115 @@ const GpsInput = ({
...
       id={descriptionId}
     >
-      {isValid
-        ? t('inputDescription', { gpsFormat: GPS_FORMAT_EXAMPLES[gpsFormat] })
-        : t('errorMessage')}
-    </p>
+      {t('inputDescription', { gpsFormat: GPS_FORMAT_EXAMPLES[coordinatesRepresentation] })}
+    </p>}
+
+    {!isInputValid && <p aria-live="assertive" className={`${styles.description} ${styles.error}`} id={errorMessageId}>
```
luixlive, line 131: Here the new error state that supports the old invalid location error message and the new outside BBOX error message, properly linked to the input for good accessibility.

src/LocationPicker/index.js
```diff
@@ -1,11 +1,15 @@
-import React, { memo, useId, useImperativeHandle, useRef, useState } from 'react';
+import React, { memo, useId, useImperativeHandle, useMemo, useRef, useState } from 'react';
```
luixlive, line 1: And finally, the changes to the `LocationPicker` itself to properly show coordinates in a CRS coordinates representation:
image

Or to show an error state with a tooltip if the current value falls outside of the supported BBOX:
image

image

src/LocationPicker/index.js
```diff
@@ -43,10 +47,23 @@ const LocationPicker = ({
 
   const inputDescriptionId = useId();
   const menuPopoverId = useId();
+  const valueOutsideBboxTooltipId = useId();
 
   const [isMenuPopoverOpen, setIsMenuPopoverOpen] = useState(false);
 
-  const displayValue = value ? transformLngLatToLocationType(value, gpsFormat) : '';
+  const { isValueOutsideCrsBbox, displayValue } = useMemo(() => {
```
luixlive, line 54: Very similar to what we did in the `GpsFormatToggle`. Same thing with the warning icon button tooltip.

src/Map/layers/useCrsBoundingBoxLayer/index.js
```diff
@@ -137,7 +137,7 @@ const useCrsBoundingBoxLayer = () => {
           // the current size of the marker and open the menu.
           const markerElementRect = event.currentTarget.getBoundingClientRect();
           dispatch(showPopup('gps-format-toggle', {
-            coordinates: [selectedCrs.bbox[0], selectedCrs.bbox[1]],
+            coordinates: [selectedCrs.bbox[0], selectedCrs.bbox[3]],
```
luixlive, line 140: Small fix after flipping the latitudes from the BBOX to match Turf format in our `epsg-index` coordinate reference system objects.

src/SideBar/SettingsPane/MapTab/CoordinateSystemSettingsView/GpsFormatSelector/index.js
```diff
@@ -6,7 +6,7 @@ import { ReactComponent as TrashCanIcon } from '../../../../../common/images/ico
 
 import { GPS_FORMAT_EXAMPLES, GPS_FORMATS } from '../../../../../utils/location';
 import {
-  setSelectedCoordinateReferenceSystems,
```
luixlive, line 9: Seems like a bunch of changes but this file only has naming updates.

src/__test-helpers/fixtures/location.js
```diff
@@ -0,0 +1,55 @@
+export const epsg2154 = {
```
luixlive, line 1: New fixtures! I noticed I had repeated myself like 1000 times in the tests adding mock EPSG objects, so I decided to move them to a fixture file. Now they are easier to manage and to import in whatever test we may need them.

You may ask: why not to simply import them from `epsg-index` library, but no, because we do some sanitization in our utility method `getProj4CompatibleCRS`, like pruning unnecessary properties and reordering the BBOX. So its easier to just keep a few of them as fixtures instead of  sing our function to load all of them and then iterate to find the one we want to  se in our test.

src/ducks/coordinate-reference-systems/index.js
```diff
@@ -1,39 +1,42 @@
 import { GPS_FORMATS } from '../../utils/location';
 
-const MAX_SELECTED_COORDINATE_REFERENCE_SYSTEMS = 5;
```
luixlive, line 3: More naming changes.

src/selectors/location/index.js
```diff
@@ -0,0 +1,25 @@
+import { createSelector } from 'reselect';
+
+const selectGpsFormat = (state) => state.view.userPreferences.gpsFormat;
+const selectStoredCoordinateReferenceSystems = (state) => state.view.coordinateReferenceSystems.storedSystems;
+
+export const selectCoordinatesRepresentation = createSelector(
```
luixlive, line 6: Two new memoized Redux selectors:
- `selectCoordinatesRepresentation`: the idea is to have a simple way to replace the places were we used to get the `gpsFormat`, since now our coordinate representations can be coordinate formats (GPS format) or coordinate reference systems. This selector provides a simple replacement and our utilities `parseCoordinates` and `stringifyCoordinates` already know how to handle these representations. 
```JavaScript
// Before
const gpsFormat = useSelector((state) => state.view.userPreferences.gpsFormat);
const coordinatesString = transformLngLatToLocationType(lngLat, gpsFormat);

// After
const coordinatesRepresentation = useSelector(selectCoordinatesRepresentation);
const coordinatesString = stringifyCoordinates(lngLat, coordinatesRepresentation);
```
- `selectStoredCoordinateReferenceSystemsMappedByCode`: returns an object with the stored CRS mapped by their code. The state variable `storedSystems` is an array, because it's useful to remember the order in which they need to be rendered in the Coordinate Settings page. However, in some places of the code we want to read the CRS objects of one of the stored systems, several times. Instead of iterating again and again the array to find the desired CRS, it's easier to use this selector and then simply read the desired stored CRS with its code:
```JavaScript
const storedCoordinateReferenceSystemsMappedByCode =
  useSelector(selectStoredCoordinateReferenceSystemsMappedByCode);
const proj4 = storedCoordinateReferenceSystemsMappedByCode['5326'].proj4;
```

src/utils/location/index.js
```diff
@@ -32,12 +33,12 @@ const isValidLatitude = (latitude) => !Number.isNaN(latitude) && Math.abs(latitu
 const isValidLongitude = (longitude, unwrapped = false) => !Number.isNaN(longitude)
   && (unwrapped || Math.abs(longitude) <= 180);
 
-const degToLngLat = (degValue) => {
```
luixlive, line 35: 75% of the changes here are naming updates.

src/utils/location/index.js
```diff
@@ -119,77 +120,104 @@ const crsToLngLat = (crsValue, crsProjection) => {
...
-          // location must be transformed to a coordinate reference system
-          // format.
-          const locationTypeCoordinates = proj4('WGS84', locationType.proj4, [numericLongitude, numericLatitude]);
-          const x = locationTypeCoordinates[0].toFixed(LNG_LAT_DECIMAL_PRECISION);
-          const y = locationTypeCoordinates[1].toFixed(LNG_LAT_DECIMAL_PRECISION);
+        if (representation?.proj4) {
+          // If the coordinates representation is an object with a property
+          // proj4, the location must be transformed to a coordinate reference
+          // system format.
+          if (representation.bbox) {
```
luixlive, line 177: Here is the important part to review of this file, how we now check if the `lngLat` object that comes as an argument of `stringifyCoordinates` falls within the BBOX of a CRS representation. If it doesn't we return a constant `OUTSIDE_BBOX`. The process is a bit tricky since we need to adjust the longitude for unwrapped BBOX.

src/utils/location/index.js
```diff
@@ -236,8 +264,15 @@ export const getProj4CompatibleCRS = async () => {
...
             // If the coordinate reference system has a bbox, reorder its parts
             // to the format expected by Turf utilities.
-            const [minLat, minLng, maxLat, maxLng] = bbox;
-            bbox = [minLng, minLat, maxLng, maxLat];
+            const [maxLatitude, minLongitude, minLatitude, maxLongitude] = bbox;
+            // If the maximum longitude is less than the minimum longitude, the
+            // BBOX passes through the antimeridian. In those cases Turf
+            // expects the locations to be unwrapped so we adjust the maximum
+            // longitude.
+            const maxLongitudeWithAntimeridianCrossingFixed = maxLongitude < minLongitude
```
luixlive, line 272: Antimeridian crossing fix for BBOX with locations that go through the antimeridian. `epsg-index` returns something like:
minLng: 175.23
maxLng: -164.94

It may be ackward that the max is less than the min, but it makes sense if you consider that the BBOX goes through the antimeridian, the minLng needs to be at the left of the maxLng, and when crossing the antimeridian, that means that the min is higher than the max. However Turf doesn't understand this. Turf expects to receive "unwrapped" longitudes, this means, that they are not wrapped by -180 / 180. maxLng must always be higher than minLng, so we need to unwrap them by adding 360 (total longitude span) to the max:
minLng: 175.23
maxLng: ~-164.94~ 195.06 

Now maxLng is over 180, meaning that it went accross the antimeridian. When unwrapped, the line of a longitude 195.06  is exactly the same than -164.94.

src/App.scss
JoshuaVulcan: 🥹 I never thought I'd see the day you reach for a global CSS class 🥹

JoshuaVulcan:
Overall, this looks awesome! Clean code, nice design, great execution.  There's just a couple of UX bits I wanted to address:

The ellipsis/tab size limitation is too much. I only have two systems selected here, and it somehow still manages to cut off most of one of them:
Screenshot 2025-08-19 at 1 47 28 PM

Ideally, those would take up the available space within a sensible max for legibility.

The other thing: seems like should make location fields uneditable when you have an invalid coordinate reference system selected as the representation of that place. Otherwise you can edit the "N/A" and do all sorts of weird things to your location fields, including delete them.
invalid-formatted-field-still-editable

luixlive:
> The ellipsis/tab size limitation is too much. I only have two systems selected here, and it somehow still manages to cut off most of one of them
> 
> Ideally, those would take up the available space within a sensible max for legibility.

Good catch. Yep, I'll work a bit more on those styles, I don't know why are we cutting the width so much.

> The other thing: seems like should make location fields uneditable when you have an invalid coordinate reference system selected as the representation of that place. Otherwise you can edit the "N/A" and do all sorts of weird things to your location fields, including delete them.

Ok, I can do that. However before that I want to say that I kind of did this consciously. The thing here is, from my perspective and with all the thought I have put in this task, I feel like we don't want to prohibit the user doing something anything. Like making the field uneditable (disabled). I think of a scenario where an event was originally configured in a wrong location outside the allowed BBOX of a CRS, and then the user wants to update with the right coordinates in that same CRS projection. With the flow you suggest, we would make impossible for the user to copy / paste (or simply write) the coordinates since the original value was outside the BBOX, thus disabling the field. I decided to set a `N/A` with the input in an error state and the LocationPicker fallbacking to degrees so the user still sees the current value and understands that in the currently selected CRS, it's not possible to _show_ those coordinate, but if they want, they still can _edit_ them.

In contrast to other epics I've worked on, this one really requires you to think on the UX a user would expect, and there are many possible scenarios. Having these discussions is beneficial so it's not only my decision. Again I feel like it's better to never disable these fields, just to provide all the feedback for the user to understand why the coordinates may be invalid in a certain representation, but still allow them to edit those coordinates.

image

_User mind: ok, these coordinates can't be shown in this CRS, but I can still clear them and write whatever I want in this format_

Let me know what you think @JoshuaVulcan

luixlive:
Done @JoshuaVulcan !

Just 2 representations with small names:
image

5 representations with small names:
image

JoshuaVulcan (approved):

PR #1356, opened 2025-08-18
Hotfix: avoid overriding the open event from the event store when fetching map events

luixlive:


src/ducks/events.js
```diff
@@ -467,20 +479,24 @@ export const fetchMapEvents = (map, parameters) => async (dispatch, getState) =>
...
-const fetchMapEventsComplete = results => (dispatch) => {
+const fetchMapEventsComplete = results => (dispatch, getState) => {
+  const state = getState();
+
   dispatch({
     type: FETCH_MAP_EVENTS_COMPLETE,
     payload: results,
   });
-  dispatch(updateEventStore(...results));
+  dispatch(updateEventStore(...removeOpenEventIfAlreadyLoaded(results, state.data.eventStore)));
```
luixlive: This call to `updateEventStore` and the next one (line 499) were causing the issue. Basically when loading map events, on every page load and after completion. We were updating the event store, overriding the old events with the news that were loaded from the map bbox. Events that were in the store but not in the bbox were left untouched in the store. My solution was simply to remove the open event from the "fetchMapEvents" response if it was already loaded, since we want to leave the handling of that event to the sidebar.

src/ducks/events.js
```diff
@@ -127,7 +137,7 @@ const fetchNamedFeedActionCreator = (name) => {
     })
       .then((response) => {
         if (typeof response !== 'undefined') { /* response === undefined for canceled requests. it's not an error, but it's a no-op for state management */
-          dispatch(updateEventStore(...response.data.data.results));
+          dispatch(updateEventStore(...removeOpenEventIfAlreadyLoaded(response.data.data.results, state.data.eventStore)));
```
luixlive: I also added the utility method to clear the event here and in line 175. These aren't part of the original issue, they are called when fetching the event feed. It's nice to add the check there too.

src/ducks/events.js
```diff
@@ -70,6 +71,15 @@ const shouldAppendLocationToRequest = (state) => {
   return userIsGeoPermissionRestricted(currentUser) && !!state?.view?.userLocation?.coords;
 };
 
+const removeOpenEventIfAlreadyLoaded = (events, eventStore) => {
```
luixlive: New utility to calculate if we currently have an open event in the side bar and if we do, remove that event from an array of events only if it is part of the event store already. If it's not, of course we want to keep it in the array since it's better to have something than nothing. The idea of this utility is to add it just in the places were we fetch events in bulk just to make sure to not override the data of the open event in the store.

PR #1357, opened 2025-08-18
Hotfix: avoid overriding the open event from the event store when fetching map events [Release]

luixlive:


src/SideBar/SettingsPane/MapTab/MainMapSettingsView/GeneralFieldSet/index.test.js
```diff
@@ -96,7 +96,7 @@ describe('SideBar - SettingsPane - MapTab - MainMapSettingsView - GeneralFieldSe
     expect(toggleMapDataSimplificationOnZoom).toHaveBeenCalledTimes(1);
   });
 
-  test.skip('opens the coordinate system settings view when the user clicks the coordinate systems button', async () => {
```
luixlive, line 99: No longer necessary since the `testUtils` file mocks the flag as true.

src/ducks/events.js
```diff
@@ -70,6 +71,15 @@ const shouldAppendLocationToRequest = (state) => {
   return userIsGeoPermissionRestricted(currentUser) && !!state?.view?.userLocation?.coords;
 };
 
+const removeOpenEventIfAlreadyLoaded = (events, eventStore) => {
```
JoshuaVulcan: nit: `openEventIfAlreadyLoaded` is vague. The notion is to convey a "locked" state for an event by ID based on its focus in the view/path right? How would you convey that? `excludeStateDependentEventIDs` or something?

JoshuaVulcan (dismissed):

src/ducks/events.js
luixlive: I think the vague portion may be just the "already loaded". The "remove open event if..." sounds good and explicit to me. Since the check we are doing is `eventStore[openEventId]` to decide if we need to remove the open event or not, I guess I can update it to `excludeOpenEventIfAlreadyInEventStore`?

I also like `excludeStateDependentEventIDs`, but I feel like it's not giving enough information. Lots of places in the app are state dependent of the event store, and they are not affected by it being updated like in this specific scenario we want to fix.

Will send that new suggestion for now, if you like it feel free to merge. Otherwise, I'm sure we'll get to a name we both like 🤠

JoshuaVulcan: It's true, naming is the WORST. call it whatever you want

PR #1358, opened 2025-08-18
Hotfix: Correct the selector that calculates subject groups last update in map layers

luixlive:


src/selectors/subjects.js
```diff
@@ -51,14 +51,14 @@ export const getSubjectGroups = createSelector(
...
-          (!lastPositionTime || new Date(subGroup.lastPositionTime) > new Date (lastPositionTime))) {
+          (!lastPositionTime || new Date(subGroup.lastPositionTime) > new Date(lastPositionTime))) {
           lastPositionTime = subGroup.lastPositionTime;
         }
       });
       hydratedSubjects.forEach((subject) => {
         const subjectLastPositionTime = subject?.last_position?.properties?.coordinateProperties?.time;
         if (subjectLastPositionTime &&
-          (!lastPositionTime || new Date(subject.updated_at) > new Date (lastPositionTime))) {
+          (!lastPositionTime || new Date(subjectLastPositionTime) > new Date(lastPositionTime))) {
```
luixlive, line 61: This is the fix. When I initially implemented this selector, I thought the property we wanted to check fore sorting was `updated_at`. But that is wrong, that property tells you when was the last time the subject was modified. What we really want is the time of the last position of the user. Some times the integrations also update the subject, that's why it works some times. But some times we just append a new observation to the assigned source, updating the calculated value `last_position`. I "fixed" my original issue, but I just forgot one place where I had to do the change from `updated_at` to `last_position...`.

PR #1359, opened 2025-08-18
Hotfix: Correct the selector that calculates subject groups last update in map layers [Release]

luixlive:


JoshuaVulcan (approved):
Easy to miss 😓

PR #1363, opened 2025-08-21
ERA-11773: Apply selected coordinate transform to all location fields and strings (fallback to DEG)

luixlive:
### What does this PR do?
This PR implements the new utility method `stringifyCoordinates` in the combination with the Redux selector `selectCoordinatesRepresentation` in all the places where we show coordinates to the user so they are rendered in the selected "representation".

In order to achieve this, and reuse code, I created two new utilities:
- `useStringifyCoordinates`: A utility hook that receives a `coordinates` argument (lngLat object) and returns an object with two properties:
  1- `coordinatesString`: the coordinates string in the currently selected representation. If the representation is a coordinate reference system and the coordinates fall outside of its supported BBOX, the coordinates string is returned in `DEG` format.
  2- `outsideRepresentationBbox`: a boolean notifying if the coordinates fall outside the representation supported BBOX, so the implementator is aware that the string is actually in `DEG`.

- `IconTooltip`: A shared component to render accessible icon tooltips. It works on hover, focus and tap (for mobile devices).

### How does it look
GPS Cursor Coordinates:
image

Subject Historical Data:
image

Mouse Map Marker:
Kapture 2025-08-21 at 16 10 13

### Relevant link(s)
* [ERA-11773](https://allenai.atlassian.net/browse/ERA-11773)
* [Env](https://era-11773.dev.pamdas.org/)

I'd suggest to start by understanding the utilities in isolation:
- `src/hooks/useStringifyCoordinates/index.js`
- `src/IconTooltip/index.js`

Then just navigate the rest of the files watching their implementation. Some of them still directly use `stringifyCoordinates`, since the hook can only be implemented directly on the body of a component, but even then the implementation is pretty straightforward.

Don't get scared about the number of files, most of them are translation updates!

[ERA-11773]: https://allenai.atlassian.net/browse/ERA-11773?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tiffanynwong:
Looks so good!!! 
Just one small thing, can we change the title to say "Coordinate Systems" instead of "Coordinates." That way it matches the setting name in the previous view. 
Screenshot 2025-08-21 at 3 11 20 PM

ctrier:
Looks great! Everything seems to be working as-expected for me

src/CursorGpsDisplay/index.js
```diff
@@ -18,19 +17,21 @@ const CursorGpsDisplay = () => {
...
 
-  const gpsFormat = useSelector((state) => state.view.userPreferences.gpsFormat);
-
   const buttonRef = useRef();
 
   const [cursorCoordinates, setCursorCoordinates] = useState(null);
   const [isMenuPopoverOpen, setIsMenuPopoverOpen] = useState(false);
 
-  const isValidLocation = validateLocation(cursorCoordinates);
+  const {
```
luixlive, line 25: Implement the new `useStringifyCoordinates` in our `CursorGpsDisplay` component. Since this widget shows the current coordinates of the cursor over the map, it doesn't make sense to add the tooltip icon here since it would be impossible to hover it while it shows the coordinates of the button. So what it does instead is just show a `N/A` if the cursor is currently over a point in the map outside of the CRS BBOX:
Kapture 2025-08-22 at 14 04 05

_The CursorGpsDisplay is located at the top right of the map._

src/GpsFormatToggle/index.js
```diff
@@ -52,6 +50,8 @@ const GpsFormatToggle = ({
   // add a fallback in case the implementator didn't provide one.
   const nameFallback = useId();
 
+  const { coordinatesString, outsideRepresentationBbox } = useStringifyCoordinates(lngLat);
```
luixlive, line 53: There are no tangible change sin this component, other than implementing the new `useStringifyCoordinates` hook and the `IconTooltip` component instead of repeating code.

src/IconTooltip/index.js
```diff
@@ -0,0 +1,42 @@
...
+
+import { ReactComponent as TriangleExclamationIcon } from '../common/images/icons/triangle-exclamation.svg';
+
+import * as styles from './styles.module.scss';
+
+export const TYPES = { WARNING: 'WARNING' };
+
+const ICONS = { [TYPES.WARNING]: <TriangleExclamationIcon /> };
+
+const IconTooltip = ({
```
luixlive, line 13: First I thought: let's create a reusable `CoordinatesOutsideBboxIconTooltip`. So we don't have to implement the warning icon with the same text everywhere. But then I realized that we could abstract it even further so it's even more reusable in the future. I decided to create a generic `IconTooltip` that could be useful in plenty of scenarios later, and all we need to do is add more themes so we can easily configure the icon through the `type` prop. Very simple to implement and customize, it's semantic, accessible and keyboard navigatable!
Kapture 2025-08-22 at 14 09 48

src/LocationPicker/index.js
```diff
@@ -51,19 +49,10 @@ const LocationPicker = ({
...
-      const displayValue = stringifyCoordinates(value, coordinatesRepresentation);
-      if (displayValue === OUTSIDE_BBOX) {
-        return { displayValue: stringifyCoordinates(value), isValueOutsideCrsBbox: true };
-      }
-      return { displayValue, isValueOutsideCrsBbox: false };
-    }
-
-    return { displayValue: '', isValueOutsideCrsBbox: false };
-  }, [coordinatesRepresentation, value]);
+  const {
```
luixlive, line 52: Just as with `GpsFormatToggle`, there aren't real changes here, other than refactoring the code to reuse the new utilities `useStringifyCoordinates` and the `IconTooltip`.

src/MouseMarkerPopup/index.js
```diff
@@ -12,9 +10,12 @@ import * as styles from './styles.module.scss';
 const MouseMarkerPopup = ({ location = null, ...rest }) => {
   const { t } = useTranslation('map-popups', { keyPrefix: 'mouseMarkerPopup' });
 
-  const gpsFormat = useSelector((state) => state.view.userPreferences.gpsFormat);
+  const { coordinatesString, outsideRepresentationBbox } = useStringifyCoordinates({
```
luixlive, line 13: Second place where we implement our new `useStringifyCoordinates` to properly show the coordinates (or a "N/A") of the `MouseMarkerPopup` while the user sets a marker in the map:
Kapture 2025-08-22 at 14 13 46

src/ReportFormSummary/V2SchemaFormSummary/index.js
```diff
@@ -2,16 +2,17 @@ import React, { useMemo } from 'react';
...
-import { FORM_ELEMENT_TYPES, ROOT_CANVAS_ID } from '../../utils/v2-event-schemas/constants';
+import { selectCoordinatesRepresentation } from '../../selectors/location';
 
 import * as styles from './styles.module.scss';
 
 const FieldSummary = ({ field, formData, id }) => {
   const { i18n, t } = useTranslation('details-view', { keyPrefix: 'reportFormSummary.v2SchemaFormSummary' });
 
-  const gpsFormat = useSelector((state) => state.view.userPreferences.gpsFormat);
+  const coordinatesRepresentation = useSelector(selectCoordinatesRepresentation);
```
luixlive, line 15: You will see a couple files like this one where we call the method `getHumanizedFieldValue` to render readable strings of v2 event fields. That method used to receive the `gpsFormat` but now in order to properly render coordinates we need the new `coordinatesRepresentation`. It's just a swap.

src/SideBar/SettingsPane/MapTab/CoordinateSystemSettingsView/CoordinateRepresentationsSelector/index.js
```diff
@@ -5,6 +5,7 @@ import { useTranslation } from 'react-i18next';
 import { ReactComponent as TrashCanIcon } from '../../../../../common/images/icons/trash-can.svg';
 
 import { GPS_FORMAT_EXAMPLES, GPS_FORMATS } from '../../../../../utils/location';
+import { selectCoordinatesRepresentation } from '../../../../../selectors/location';
```
luixlive, line 8: This file seems to have a big amount of changes, but it's just the continuation of what I started in the previous PR, a re-wording to better reflect what each variable and method holds. I'm replacing strings like "GpsFormatOption" to "CoordinatesRepresentationOption". No logical or visual changes, just code readability improvements.

src/SubjectHistoricalDataModal/index.js
```diff
@@ -49,9 +51,14 @@ const getPageItemNumbers = (pagesCount, activePage) => {
 const ObservationRow = ({ observation, observationProperties, subjectIsStatic }) => {
   const { t } = useTranslation('subjects', { keyPrefix: 'subjectHistoricalDataModal.observationRow' });
 
-  const gpsFormat = useSelector((state) => state.view.userPreferences.gpsFormat);
+  const coordinatesRepresentation = useSelector(selectCoordinatesRepresentation);
 
-  const locationString = !subjectIsStatic && stringifyCoordinates(observation.location, gpsFormat);
+  const observationOutsideBboxTooltipId = useId();
+
+  const {
```
luixlive, line 58: Third and last place where we needed to implement our new utilities to render coordinates, the `SubjectHistoricalDataModal`. Here an example of a subject with a source that has two observations, one inside the BBOX of the CRS and the other outside. The additional data table shows the coordinates in the CRS or in DEG with a warning tooltip:
image

src/hooks/useStringifyCoordinates/index.js
```diff
@@ -0,0 +1,19 @@
+import { useMemo } from 'react';
+import { useSelector } from 'react-redux';
+
+import { OUTSIDE_BBOX, stringifyCoordinates } from '../../utils/location';
+import { selectCoordinatesRepresentation } from '../../selectors/location';
+
+const useStringifyCoordinates = (coordinates) => {
```
luixlive, line 7: The other protagonist of this PR! A very nice and simple to implement hook that is fed a set of coordinates and returns them as a string in the current coordiantes representation. It also provides a flag so the implementator is aware if the coordinates are outside the supported BBOX of the representation. In that case, the coordinates are returned in DEG format.
```JavaScript
const { coordinatesString, outsideRepresentationBbox } = useStringifyCoordinates(coordinates);

return <span>
  {coordinatesString}
  {outsideRepresentationBbox && 'Coordinates are shown in degrees because they are outside the BBOX.'}
</span>
```

src/utils/location/index.js
```diff
@@ -276,7 +276,7 @@ export const getProj4CompatibleCRS = async () => {
           }
 
           acc.push({
-            area: coordinateReferenceSystem.area,
+            area: coordinateReferenceSystem.area || '',
```
luixlive, line 279: @JoshuaVulcan found an issue with some CRS that don't have an `area` property breaking the app, so just added this empty string as a fallback to avoid runtime exceptions.

src/utils/v2-event-schemas/getHumanizedFieldValue/index.js
```diff
@@ -1,15 +1,15 @@
...
 import { shouldUse12HourFormat } from '../../datetime';
-import { stringifyCoordinates } from '../../location';
 
 const getChoiceListOptionLabel = (value, field) => field.details.options
   .find((option) => option.const === value)?.title;
 
 // Utility to calculate a human readable version of the field values. For example, render a date-time like
 // 2020/01/01 12:00 PM instead of 2020-01-01T12:00:00Z.
-const getHumanizedFieldValue = (field, value, defaultHumanizedValue, language, gpsFormat, t) => {
+const getHumanizedFieldValue = (field, value, defaultHumanizedValue, language, coordinatesRepresentation, t) => {
```
luixlive, line 12: This is a method used in many places since the integration with event types v2. It's useful to show event fields summaries in readable strings. Since we support the location field, and it's summary consists in coordinates, we needed to use the new `stringifyCoordinates` utility and provide the `coordinatesRepresentation` to it, instead of the `gpsFormat`. Of course it's no possible to use the hook `useStringifyCoordinates` here because this is not a component.

gaboDev (approved):

PR #1367, opened 2025-08-27
add guard hook to ensure mapboxGL rendering is supported in current client context

JoshuaVulcan:


public/locales/en-US/errors.json
```diff
@@ -0,0 +1,3 @@
+{
+  "webGlDisabled": "Hardware acceleration is disabled or unsupported in your browser. This is required to use EarthRanger. Please turn on hardware acceleration in your browser settings, or contact your administrator for technical support."
```
luixlive, line 2: Maybe in order to avoid support time and tickets, we could work on a more complete set of instructions for the users to try before asking for support: update the browser, enable hardware acceleration (maybe provide a link or more specific instructions on how to do this in different browsers), restart the browser or computer...

src/App.js
```diff
@@ -52,6 +57,8 @@ export const App = () => {
   );
   const trackSettings = useSelector((state) => state.view.trackSettings);
 
+  const canRenderMap = useDetectMapboxSupported();
```
luixlive: nit: honor the name of the hook by calling the variable: `isMapboxSupported`.

src/App.js
```diff
@@ -157,7 +164,9 @@ export const App = () => {
         <Nav map={map} />
 
         <div className={`app-container ${sidebarOpen ? 'sidebar-open' : 'sidebar-closed'}`}>
-          <Map map={map} onMapLoad={onMapHasLoaded} socket={socket} />
+          {canRenderMap && <Map map={map} onMapLoad={onMapHasLoaded} socket={socket} />}
+          {!canRenderMap && <ErrorMessage className='webgl-error-message'
```
luixlive: Kind of agree with Copilot here, why not to use a SCSS imported class like `style.webglErrorMessage` as we usually do?

src/hooks/useDetectMapboxSupported/index.js
```diff
@@ -0,0 +1,7 @@
+import mapboxgl from 'mapbox-gl';
+
+const useDetectMapboxSupported = () => {
```
luixlive: Hmm seems a bit of overengineering to wrap this single line in a hook. Maybe we could have just placed it directly in `App`like:
```JavaScript
const isMapboxSupported = mapboxgl.supported();
```

luixlive (approved):
Great solution. Added a few suggestions, but don't want to block this fix given the current impact. Maybe we can create a proper ticket to work on that as technical debt.

src/App.js
JoshuaVulcan: App.scss doesn't have a style module, and it seemed like overkill to add one for this one quick hotfix

src/hooks/useDetectMapboxSupported/index.js
JoshuaVulcan: On it 🫡

PR #1368, opened 2025-08-27
[ERA-11923] Capturing important support data

JoshuaVulcan:
Adds analytics collection for affected sites so we can maintain some tracking and info for supporting client-side rendering issues related to hardware acceleration support changes in Chrome v139.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,4 +1,4 @@
-import React, { createContext, useCallback, useContext, useEffect, useState } from 'react';
+import React, { createContext, useCallback, useContext, useEffect, useMemo, useState } from 'react';
 import axios from 'axios';
 import mapboxgl from 'mapbox-gl';
 import { loadProgressBar } from 'axios-progress-bar';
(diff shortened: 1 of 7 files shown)


src/App.js
```diff
@@ -40,8 +44,6 @@ import WithSocketContext, { SocketContext } from './withSocketConnection';
 import 'axios-progress-bar/dist/nprogress.css';
 import './App.scss';
 
-const mapboxSupported = !!mapboxgl.supported();
```
luixlive, line 43: Why not to keep it outside so we don't have to memoize it?

src/App.js
```diff
@@ -148,6 +155,17 @@ export const App = () => {
     }
   }, [showGeoPermWarningMessage]);
 
+  useEffect(() => {
+    if (!mapboxSupported && process.env.NODE_ENV === 'production') {
+      const userData = createUserAnalyticsData(user, selectedUserProfile, serverVersion);
+
+      ReactGA4.event('MapboxGL not supported', {
+        event_category: 'Client Hardware Issue',
```
luixlive, line 162: Shouldn't this string be a constant somewhere?

luixlive (approved):

PR #1369, opened 2025-08-28
ERA-11901: Add link to Ecoscope from hamburger menu

luixlive:
### What does this PR do?
Full refactor of our `GlobalMenuDrawer` to make it semantically correct, accessible and keyboard navigatable. And of course, adds the new Ecoscope link.

Extra: Updates the dependencies of the project and `yarn` to version 4.

### How does it look
image

### Relevant link(s)
* [ERA-11901](https://allenai.atlassian.net/browse/ERA-11901)
* [Env](https://era-11901.dev.pamdas.org)

**Note:** I know this ticket has 1 story points so 33 files changed doesn't make sense, but most of them are caused by the upgrades of our dependencies. yarn.lock, package.json, a few new yarn config files, Dockerfile updates to work with the new package manager version, and some tests I had to fix after `react-date-picker` updated some of the accessible roles of its components. It won't take you long to go through those changes.

[ERA-11901]: https://allenai.atlassian.net/browse/ERA-11901?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.yarn/install-state.gz
luixlive, line 1: The first three files are part of the `yarn` 4 upgrade.

Dockerfile.dev
luixlive, line 1: Then a bunch of updates to our `Dockerfiles`. 
- Upgrading to the latest version of Nginx (removes a bunch of vulnerabilities).
- Fix our Dockerfiles to not run unnecessary steps (like `RUN apk add --no-cache git alpine-sdk python3`).
- Update the `yarn` installation after the upgrade since `--ignore-scripts` is no longer supported.
- Remove an unnecessary volume from our `docker-compose` file.

package.json
```diff
@@ -5,29 +5,29 @@
   "buildbranch": "generated_by_build",
   "buildnum": "also_generated_by_build",
   "dependencies": {
-    "@babel/core": "^7.28.0",
+    "@babel/core": "^7.28.3",
```
luixlive, line 8: Lots of upgrades, most of them patches.

package.json
```diff
@@ -36,112 +36,110 @@
...
+    "web-vitals": "^5.1.0",
+    "webpack": "^5.101.3",
     "webpack-dev-server": "^5.2.2",
     "webpack-manifest-plugin": "^5.0.1",
     "workbox-webpack-plugin": "^7.3.0"
   },
   "scripts": {
     "start": "node react-scripts/start.js",
     "build-sw": "node ./src/sw-build.js",
-    "clean-cra-sw": "rm -f build/precache-manifest.*.js && rm -f build/service-worker.js",
```
luixlive, line 117: I removed two unnecessary legacy scripts:
- `clean-cra-sw`: designed to remove the service worker output files from the `build` react script. `yarn` 4 runs these scripts with `zsh`, which return an error when trying to remove a file that doesn't exist with `rm`. The previous version of `yarn` used `sh` which just silently ignored the fact that the file didn't exist. The thing is that those files are not being generated anymore, probably since the ejection, so we can just remove this script. Our custom service worker is still there after `yarn build-sw` execution.
- `postinstall`: A `sh` script from a file `setUpLocalGitConfig.sh`. Basically checks that git is in the env and it tries to read a `.gitconfig` file from the parent. There are three reasons why I decided to remove it:
  1- `yarn` 4 doesn't support `--ignore-scripts`. In order to run the installation without the postinstall we need to create a custom `.yarnrc` file for docker only, which we can do but feels unnecessary.
  2- I don't think we should fail the installation process just because a dev doesn't have git installed yet. Git is a software for collaboration, but isn't related to installing and building the app really.
  3- That script assumes that the `.gitconfig` file is in the parent folder of where this project is stored in the local environment, which won't be true for the big majority of the cases. For example, in my case it's not true.

src/App.scss
```diff
@@ -113,13 +113,6 @@ body {
   }
 }
 
-aside {
```
luixlive, line 116: I moved this styles to a more scoped file (`MapLayersTab`) instead of having a global for a very specific component.

src/GlobalMenuDrawer/index.js
```diff
@@ -1,4 +1,4 @@
-import React, { lazy, useCallback, useMemo } from 'react';
+import React, { lazy, useCallback, useEffect, useMemo, useRef } from 'react';
```
luixlive, line 1: This is the file you really want to take your time to review. The diff is very noisy here, I'd recommend you to pull the changes and review the file in your text editor. It's really simple really, and our `GlobalMenuDrawer` is now semantic, accessible and easy to navigate with the keyboard (I even added a focus trap).

src/Nav/NavHomeMenu/styles.module.scss
```diff
@@ -1,7 +1,7 @@
 @import '../../common/styles/layout';
 @import '../../common/styles/vars/colors';
 
-nav {
+header {
```
luixlive, line 4: In the next files I did a few other semantic updates to some of the most important (semantically speaking) components of our app. We were wrongly using a `<nav>` for our header. We don't have any navigation elements in it, so I updated it from a `<nav>` to a `<header>`:
image

We were also using an `<aside>` for our real navigation component haha. `<aside>` is usually a component that is always visible and that is complimentary info of the main content of the app. That is not our case, it is a navigation component because there's where our different links are, so I updated it too:
image

src/PatrolDetailView/PlanSection/index.test.js
```diff
@@ -183,8 +183,7 @@ describe('PatrolDetailView - PlanSection', () => {
     const startDatePicker = await screen.findByTestId('patrolDetailView-planSection-startDatePicker');
     const startDatePickerOpenCalendarButton = await within(startDatePicker).findByLabelText('Open calendar');
     await userEvent.click(startDatePickerOpenCalendarButton);
-    const options = await screen.findAllByRole('option');
-    await userEvent.click(options[25]);
+    await userEvent.click(screen.getByRole('gridcell', { name: 'Choose Friday, January 28th, 2022' }));
```
luixlive, line 186: You will see a few test updates like this one. `react-date-picker` decided to do an accessibility change. Instead of giving the `option` role to each day in the calendar, they decided to switch to `gridcell` since the team thought it was a more fitting role. I had to update the tests where we were relying on their role to get them.

JoshuaVulcan (approved):

PR #1371, opened 2025-08-29
ERA-11882: Subjects/Subject groups are displayed incorrectly when the last location string is too long.

luixlive:
### What does this PR do?
Set humanize-duration largest option to 2 so we don't have long time ago texts.

### How does it look
**Before:**
image

**After:**
image

### Relevant link(s)
* [ERA-11882](https://allenai.atlassian.net/browse/ERA-11882)

Here is the official NPM docs for `humanize-duration`: https://www.npmjs.com/package/humanize-duration .
In that page, search `largest` option: "Integer representing the maximum number of units to use.". Pretty straightforward.

Since it's a trivial / safe change, and I saw the fix working directly in the affected by running it locally, I didn't create an environment.

[ERA-11882]: https://allenai.atlassian.net/browse/ERA-11882?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gaboDev (approved):
LGTM

Also, it would be worth it to add flexibility to the map layer title with CSS.

Not sure of that would have side effects in other parts of the app so I'm approving since the main goal was accomplished

luixlive:
> Also, it would be worth it to add flexibility to the map layer title with CSS.

Happy to work on anything that improves our list. But could you explain a bit more? What do you mean with _adding flexibility to the map layer title_ @gaboDev ?

JoshuaVulcan (approved):

gaboDev:
> > Also, it would be worth it to add flexibility to the map layer title with CSS.
> 
> Happy to work on anything that improves our list. But could you explain a bit more? What do you mean with _adding flexibility to the map layer title_ @gaboDev ?

When there is no enough space the title tends to break chaotically instead showing elipsis dots or break gracefully 

@luixlive

luixlive:
Done @gaboDev 
image

I added ellipsis and if the user hover they will see the tooltip with the complete name (I added it as the title of the wrapper div).

PR #1374, opened 2025-09-04
ERA-11795: Search for location by name

luixlive:
### What does this PR do?
Adds support to search a location by text in the GpsInput component.

### How does it look
image

### Relevant link(s)
* [ERA-11795](https://allenai.atlassian.net/browse/ERA-11795)
* [Env](https://era-11795.dev.pamdas.org)

I'd suggest to review this files in order:
1- `src/utils/location/index.js`
2- `src/GpsInput/index.js` and its styles.
3-`src/GpsFormatToggle/index.js` and its styles.

The rest of the files will be way easier to understand once you fully got the changes done in those two.

As suggested in the ticket, I'm using [Mapbox Forward Geocoding endpoint](https://docs.mapbox.com/api/search/geocoding/#forward-geocoding-with-search-text-input) to fetch the locations results, and I created a utility to fetch the locations asynchronously and sanitize the response.

From there, I just added the logic to run the search in our existing `GpsInput` component, making sure the input remains totally accessible as a [combobox](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/combobox_role).

The changes in `GpsFormatToggle` are a bit tricky since previously, all the `radiogroup` of that component was controlled through our `gpsFormat` store variable, but now the text search option isn't. That one will be controlled by the parent (which is `GpsInput`) and the code conciliates these two sources (store and parent props) to control the `radiogroup` as a whole, keeping it accessible and semantically correct.

[ERA-11795]: https://allenai.atlassian.net/browse/ERA-11795?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/CursorGpsDisplay/MenuPopover/index.js
+++ b/src/CursorGpsDisplay/MenuPopover/index.js
@@ -24,21 +24,30 @@ const MenuPopover = ({ buttonRef, className, onClose, ref, ...otherProps }) => {
 
   const [gpsInputValue, setGpsInputValue] = useState(null);
 
-  const onJumpToCoordinates = useCallback(() => {
+  const onJumpToCoordinates = useCallback((coordinates = gpsInputValue) => {
     // If the GPS input value is defined, jump to the coordinates, show a dropped marker popup in the map and close the
     // menu.
-    if (gpsInputValue) {
-      jumpToLocation([gpsInputValue.longitude, gpsInputValue.latitude]);
+    if (coordinates) {
+      jumpToLocation([coordinates.longitude, coordinates.latitude]);
 
       setTimeout(() => dispatch(showPopup('dropped-marker', {
-        coordinates: [gpsInputValue.longitude, gpsInputValue.latitude],
-        location: { lat: gpsInputValue.latitude, lng: gpsInputValue.longitude },
+        coordinates: [coordinates.longitude, coordinates.latitude],
+        location: { lat: coordinates.latitude, lng: coordinates.longitude },
         popupAttrsOverride: { offset: [0, 0] },
       })), 50);
 
       onClose();
     }
   }, [dispatch, gpsInputValue, jumpToLocation, onClose]);
+
+  const onGpsInputPlaceSelected = (event, place) => {
+    // Either if the place was selected by clicking or pressing enter, we stop
+    // the propagation of the event so the GPS input keyboard events and map
+    // clicking events don't get triggered.
+    event.stopPropagation();
+
+    onJumpToCoordinates(place.coordinates);
+  };
 
   const onGpsInputKeyDown = (event) => {
     switch (event.key) {
@@ -116,6 +125,7 @@ const MenuPopover = ({ buttonRef, className, onClose, ref, ...otherProps }) => {
       inputRef={gpsInputRef}
       onChange={setGpsInputValue}
       onKeyDown={onGpsInputKeyDown}
+      onPlaceSelected={onGpsInputPlaceSelected}
       ref={gpsInputWrapperRef}
       renderButton={() => <button
         aria-label={t('gpsInputButtonLabel')}
--- a/src/CursorGpsDisplay/MenuPopover/index.js
+++ b/src/CursorGpsDisplay/MenuPopover/index.js
@@ -128,6 +128,7 @@ const MenuPopover = ({ buttonRef, className, onClose, ref, ...otherProps }) => {
       >
         <SearchIcon className={styles.searchIcon} />
       </button>}
+      showTextSearchOption
       title={t('gpsDisplayTooltip')}
       value={gpsInputValue}
     />
--- a/src/CursorGpsDisplay/MenuPopover/index.js
+++ b/src/CursorGpsDisplay/MenuPopover/index.js
@@ -93,13 +93,13 @@ const MenuPopover = ({ buttonRef, className, onClose, ref, ...otherProps }) => {
   }, [gpsInputValue]);
 
   useEffect(() => {
-    const onMouseDown = (event) => !buttonRef.current.contains(event.target)
-    && !gpsInputWrapperRef.current.contains(event.target)
-    && onClose();
+    const onPointerDown = (event) => !buttonRef.current.contains(event.target)
+      && !gpsInputWrapperRef.current.contains(event.target)
+      && onClose();
 
-    document.addEventListener('mousedown', onMouseDown);
+    document.addEventListener('pointerdown', onPointerDown);
 
-    return () => document.removeEventListener('mousedown', onMouseDown);
+    return () => document.removeEventListener('pointerdown', onPointerDown);
   }, [buttonRef, onClose]);
 
   return <Popover
--- a/src/GpsFormatToggle/index.js
+++ b/src/GpsFormatToggle/index.js
@@ -2,6 +2,8 @@ import React, { memo, useEffect, useId, useImperativeHandle, useRef } from 'reac
 import { useDispatch, useSelector } from 'react-redux';
 import { useTranslation } from 'react-i18next';
 
+import { ReactComponent as SearchIcon } from '../common/images/icons/search-icon.svg';
+
 import { FEATURE_FLAG_LABELS } from '../constants';
 import { GPS_FORMAT_CATEGORY, trackEventFactory } from '../utils/analytics';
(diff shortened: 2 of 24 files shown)


luixlive:
@JoshuaVulcan if I remember correctly you said it wasn't really necessary to add analytics, so I haven't added them. But if you think it would be great to add GA event so we are aware of how much this functionality is used (and the Mapbox endpoint is called) please let me know. If yes, just let me know what category / action / label you would like to see in GA.

src/CursorGpsDisplay/MenuPopover/index.js
```diff
@@ -93,13 +93,13 @@ const MenuPopover = ({ buttonRef, className, onClose, ref, ...otherProps }) => {
...
   useEffect(() => {
-    const onMouseDown = (event) => !buttonRef.current.contains(event.target)
-    && !gpsInputWrapperRef.current.contains(event.target)
-    && onClose();
+    const onPointerDown = (event) => !buttonRef.current.contains(event.target)
+      && !gpsInputWrapperRef.current.contains(event.target)
+      && onClose();
 
-    document.addEventListener('mousedown', onMouseDown);
+    document.addEventListener('pointerdown', onPointerDown);
```
luixlive, line 109: I noticed a trivial issue in a couple places in our app where we were depending on `mousedown` events that don't get fired in mobile devices (because there is no mouse involved). They were really simple to fix by changing them to be [Pointer events](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events).

src/CursorGpsDisplay/MenuPopover/index.js
```diff
@@ -128,6 +128,7 @@ const MenuPopover = ({ buttonRef, className, onClose, ref, ...otherProps }) => {
       >
         <SearchIcon className={styles.searchIcon} />
       </button>}
+      showTextSearchOption
```
luixlive, line 141: Here we set our new flag so our `CursorGpsDisplay`'s `MenuPopover` shows the text search option in the `GpsInput`:
image

src/GpsFormatToggle/index.js
```diff
@@ -22,10 +24,13 @@ const gpsFormatTracker = trackEventFactory(GPS_FORMAT_CATEGORY);
 
 const GpsFormatToggle = ({
   className = '',
+  isTextSearchOptionChecked = false,
```
luixlive, line 27: Now our `GpsFormatToggle` component supports the flag `showTextSearchOption` to render an extra option at the beginning of the `radiogroup` with a search icon. The selection of that option is entirely controlled by the parent, opposed to the selection of a GPS format that is controlled by our redux store in `state.view.userPreferences.gpsFormat`.

If the callback and state variable are properly handled by the parent, the `radiogroup` works perfectly even though we have this hybrid control of the component between the parent for the text search option and Redux for the GPS format options.

Note that if the `showTextSearchOption` is set to `false` (which is its default) the component works exactly as before, so all the other places where we use this component (like the `LocationPicker` or many map popups) still work as before.

image

src/GpsInput/index.js
```diff
@@ -15,13 +19,16 @@ import GpsFormatToggle from '../GpsFormatToggle';
...
+const FETCH_PLACES_FROM_SEARCH_TEXT_DEBOUNCE_TIME = 200;
+
 const GpsInput = ({
   gpsFormatToggleRef = null,
   id = null,
   inputRef = null,
   onChange,
   ref,
   renderButton = null,
+  showTextSearchOption = false,
```
luixlive, line 32: This component too receives the new prop `showTextSearchOption`, and actually this is the parent of `GpsFormatToggle` that handles the state of the text search option (you will se a new state variable below that handles that).

This input used to be very simple, it just captured coordinates from user input and transformed them to `LonLat` objects using our location utilities, and then, when displaying the values, we would transform them back from `LonLat` to the current coordinates representation (a coordinate reference system of a coordinates format like degrees). Now it's not that simple anymore, because it also supports text input for the text search. You will see lots of conditionals regarding functionality and rendering of the input depending on the new variable `isTextSearchOptionChecked`, with the intention of making the input semantically correct and fully accessible.

image

src/GpsInput/index.js
```diff
@@ -33,99 +40,220 @@ const GpsInput = ({
...
   const [inputValue, setInputValue] = useState(() => value
     ? stringifyCoordinates(value, coordinatesRepresentation)
     : '');
   const [isInputValueValid, setIsInputValueValid] = useState(true);
+  const [isTextSearchOptionChecked, setIsTextSearchOptionChecked] = useState(false);
+  const [placesFromSearchText, setPlacesFromSearchText] = useState(null);
+  const [selectedPlaceIndex, setSelectedPlaceIndex] = useState(null);
+
+  // Debounced method to fetch forward geocoding from the search text.
+  const fetchPlacesFromSearchText = useRef(debounce(async (searchText) => {
```
luixlive, line 58: We debounce the forward geocoding calls to 200ms. When we receive the results we store them in a state variable to display them in a `listbox` and reset the active and selected options.

src/GpsInput/index.js
```diff
@@ -33,99 +40,220 @@ const GpsInput = ({
...
+      placeFormatted: place.place_formatted,
+    })));
+  }, FETCH_PLACES_FROM_SEARCH_TEXT_DEBOUNCE_TIME)).current;
+
+  // The input is set in an error state if the text search option is not
+  // checked and the input value not a valid set of coordinates or they are
+  // outside of the coordinates representation BBOX.
+  const isInputValid = isTextSearchOptionChecked || (isInputValueValid && inputValue !== OUTSIDE_BBOX);
+
+  let description;
```
luixlive, line 76: Calculate the description, label and placeholder of the input depending on if the text search option is checked and if there are results, or if its not checked, depending on if the representation is a coordinates reference system or a coordinates format.

src/GpsInput/index.js
```diff
@@ -33,99 +40,220 @@ const GpsInput = ({
...
+        } catch (error) {
+          setIsInputValueValid(false);
+        }
+      }
+    }
+  };
 
-        const isNewInputValueValid = validateLngLat(lngLat.longitude, lngLat.latitude);
-        setIsInputValueValid(isNewInputValueValid);
+  const onPlaceOptionSelected = (index) => {
```
luixlive: When the user selects a place from the list of results (via keyboard or by clicking it) we store the selection in a state variable, update the input, and notify the parent about the new coordinates.

src/GpsInput/index.js
```diff
@@ -33,99 +40,220 @@ const GpsInput = ({
...
       lngLat={value}
       onKeyDown={(event) => event.key === 'Enter' && innerInputRef.current.focus()}
       ref={gpsFormatToggleRef}
+      setIsTextSearchOptionChecked={setIsTextSearchOptionChecked}
       showCoordinates={false}
+      showTextSearchOption={showTextSearchOption}
     />
 
     <div className={styles.inputWrapper}>
       <input
```
luixlive, line 224: This input can work now as a regular `search` input or as a `combobox` depending on if we have the text search option checked or not, so you will find many attributes (a bunch of them aria attributes) being set conditionally.

src/GpsInput/index.js
```diff
@@ -136,6 +264,36 @@ const GpsInput = ({
         })
         : t('invalidLocationErrorMessage')}
     </p>}
+
+    {isTextSearchOptionChecked && placesFromSearchText?.length > 0 && <ul
```
luixlive, line 272: If the text search option is checked and there are places retrieved already from a search, show the results list as a `listbox` with all the places shown as `options`, associated to the `combobox`. It fully accessible and you can navigate it with the up and down arrows and select an item by clicking or pressing enter:
image

src/utils/location/index.js
```diff
@@ -289,3 +294,16 @@ export const getProj4CompatibleCRS = async () => {
 
   return proj4CompatibleCRSSingleton;
 };
+
+export const fetchForwardGeocoding = async (searchText) => {
```
luixlive, line 298: And our new utility that calls the Mapbox Forward Geocoding endpoint and process the response to return an array of "place" objects.

gaboDev (approved):
LGTM!

PR #1376, opened 2025-09-10
ERA-11939: New Coordinates component is not interactable on read-only events.

luixlive:
### What does this PR do?
Update to the components we render in the Details section of events so they render as read-only when the event is read-only too. We used to simply disable the cursor events and add some weak CSS to change the visualization.

### How does it look
image

### Relevant link(s)
* [ERA-11939](https://allenai.atlassian.net/browse/ERA-11939)
* [Env](https://era-11939.dev.pamdas.org)

The bug fix is in `src/ReportManager/DetailsSection/index.js`, changing the way we make the fields read only.
What will take you more time is the whole refactoring of our `AreaPicker` component. It may be a good idea to pull the changes and analyze it carefully before starting the review. After that, you will go through the PR pretty fast.

[ERA-11939]: https://allenai.atlassian.net/browse/ERA-11939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/LocationPicker/MenuPopover/index.js
```diff
@@ -149,9 +149,10 @@ const MenuPopover = ({
   }, [isPickingLocation, onBlur, onClose, setLocationButtonRef, target]);
 
   return <Popover
+      aria-label={t('dialogLabel')}
       className={`${className} ${styles.menuPopover}`}
       ref={ref}
-      role="presentation"
+      role="dialog"
```
luixlive, line 155: A11y fix. Presentational nodes are ignored by screen readers. The proper role here is "dialog".

src/LocationPicker/MenuPopover/styles.module.scss
```diff
@@ -28,12 +28,11 @@
         color: colors.$off-black;
         display: flex;
         font-weight: 500;
-        outline: none;
         padding: 0.375rem;
         padding-right: 0.5rem;
     
         &:focus-visible {
-          border: 2px solid colors.$bright-blue;
+          outline: 2px solid colors.$bright-blue;
```
luixlive, line 35: Outline is better as a focus adornment since it doesn't cause layout shifting.

src/LocationPicker/index.js
```diff
@@ -68,6 +68,7 @@ const LocationPicker = ({
       <button
         aria-controls={menuPopoverId}
         aria-expanded={isMenuPopoverOpen}
+        aria-haspopup="dialog"
```
luixlive, line 71: Another a11y improvement over our location picker dialog. The button that opens / closes it should have `aria-haspopup="dialog"`.

src/LocationPicker/styles.module.scss
```diff
@@ -4,6 +4,7 @@
 
 .locationPicker {
   align-items: center;
+  background: white;
```
luixlive, line 7: We were setting the white background in the input and not in the group div (which is the one that receives the `className` prop. That way, it was unnecessarily complicated to change the background of the widget, like we do in the Details section when an event is read only. After this change, it is trivial to do so:
image

src/PatrolDetailView/PlanSection/styles.module.scss
```diff
@@ -74,10 +74,6 @@
   &.timePickerLabel {
     margin-left: 0.5rem;
   }
-
-  [class*=locationSelectionLabel] {
```
luixlive, line 78: Old unused styles.

src/ReportManager/DetailsSection/AreaPicker/MenuPopover/index.js
```diff
@@ -0,0 +1,241 @@
+import React, { useContext, useEffect, useRef } from 'react';
```
luixlive, line 1: Interesting how the origin of this ticket was a hacky implementation of our new `LocationPicker` as a read only field, but I ended up re-doing the `AreaPicker` component. `LocationPicker` is actually a pretty good component with good semantic and a11y practices that supports being a read-only field, the thing is that we weren't implementing it correctly. The way we were disabling all the fields in the report detail view when an event is readonly was very poor. Now I'm passing correctly the `readOnly` prop and adding some styles to make the field look grayed out (which actually isn't the default behavior of native components).

However, doing that in the `AreaPicker` was impossible because that component is older than the `LocationPicer` and didn't have all the good practiced to properly support a prop `readOnly`. I basically reworked the entire component to behave the same, while making it fully accessible, keyboard navigatable and even fixed some layout issues.

image

src/ReportManager/DetailsSection/AreaSelectorInput/GeometryPreview/index.js
luixlive, line 1: The diff says the entire component was deleted and created because I changed the name to match the pattern of our latest components: `AreaPicker`.

src/ReportManager/DetailsSection/index.js
```diff
@@ -191,14 +191,21 @@ const DetailsSection = ({
             {t('locationLabel')}
 
             {geometryType === VALID_EVENT_GEOMETRY_TYPES.POLYGON
-              ? <AreaSelectorInput
+              ? <AreaPicker
```
luixlive, line 194: This file contains the real fix of the bug: using `readOnly` props and injecting styles to grey out the background of the fields.

src/ReportManager/ReportDetailView/index.js
```diff
@@ -723,10 +723,8 @@ const ReportDetailView = ({
   const shouldRenderHistorySection = reportForm?.updates;
   const shouldRenderLinksSection = !!linkedReports.length || !!linkedPatrols.length;
 
-  const isReadOnly = eventSchema?.schema?.readonly;
-
   return <div
-    className={`${styles.reportDetailView} ${className || ''} ${isReadOnly ? styles.readonly : ''}`}
```
luixlive, line 729: The old solution (which was causing the issue) was to inject a `readonly` class name to the entire `ReportDetailView` and disabling pointer event, focus state, etc... of everything below a field label. A bit nasty, and prone to errors.

gaboDev (approved):

PR #1378, opened 2025-09-12
ERA-11899: Add beta previews to ER Web

luixlive:
### What does this PR do?
Adds a new experimental features mechanism for the user to show activation checkboxes in the General Settings view for each available feature flag. If one or many of the feature flags in our global constant object `DEVELOPMENT_FEATURE_FLAGS` is listed in the new query parameter `ef`, the user will see an "Experimental Features" field set with checkboxes for the listed features where they can override the system's default flag.

For example, if we have:
```JavaScript
// .env
REACT_APP_FEATURE_1=true
REACT_APP_FEATURE_2=false
REACT_APP_FEATURE_3=false

// src/constants/index.js
export const FEATURE_FLAG_LABELS = {
  FEATURE_1: 'FEATURE_1',
  FEATURE_2: 'FEATURE_2',
  FEATURE_3: 'FEATURE_3',
};

export const DEVELOPMENT_FEATURE_FLAGS = {
  [FEATURE_FLAG_LABELS.FEATURE_1]: process.env.REACT_APP_FEATURE_1 === 'true',
  [FEATURE_FLAG_LABELS.FEATURE_2]: process.env.REACT_APP_FEATURE_1 === 'true',
  [FEATURE_FLAG_LABELS.FEATURE_3]: process.env.REACT_APP_FEATURE_1 === 'true',
};
```

Then the user will be able to override the defaults by opening ER with the query parameter `...pamdas.org/?ef=FEATURE_1,FEATURE_2,FEATURE_3` or list just one or a few. The checkbox will have the system default value the first time it renders, meaning that in the scenario before, the checkbox for `FEATURE_1` would be checked initially, but the checkboxes for `FEATURE_2` and `FEATURE_3` would not.

Once the user interacts with a checkbox, the selected value will be stored in a local storage persisted reducer and hold the value the user gave it forever (unless the user goes and deletes the local storage value manually). The mechanism is built in such a way that the reducer doesn't require to be updated every time we add new feature flags, nor require migrations. Once a feature flag no longer exists in `DEVELOPMENT_FEATURE_FLAGS`, the checkbox will never render again, no matter what, and the key of that feature flag will eventually be removed from the persisted reducer.

### How does it look
No experimental features shown if there is no query parameter:
image

Experimental feature is shown if the query parameter is present:
image

Once modified, that specific experimental feature will continue to show up even without the query parameter so the user can still turn it on and off (until the feature flag is removed from the system, aka no longer a property in `DEVELOPMENT_FEATURE_FLAGS`):
image

### Relevant link(s)
* [ERA-11899](https://allenai.atlassian.net/browse/ERA-11899)
* [Env](https://era-11899.dev.pamdas.org/)

[ERA-11899]: https://allenai.atlassian.net/browse/ERA-11899?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/AddItemButton/AddItemModal/AddPatrolTab/index.test.js
```diff
@@ -20,7 +20,7 @@ describe('AddItemButton - AddItemModal - AddPatrolTab', () => {
     useNavigateMock = jest.fn(() => navigate);
     useNavigate.mockImplementation(useNavigateMock);
 
-    store = { data: { patrolTypes }, view: { featureFlagOverrides: {} } };
```
luixlive, line 23: The old feature flag overrides reducer is gone, so you will see lots of files just cleaning the test mocks.

src/GlobalMenuDrawer/BetaToggles/index.js
luixlive, line 1: Some time ago, the "beta toggles" were part of the `GlobalMenuDrawer`, but it's been a while since they were moved to the General settings pane. I totally deleted the old component that was within the `GlobalMenuDrawer` folder.

src/SideBar/SettingsPane/GeneralTab/AppRefreshFieldSet/index.js
```diff
@@ -0,0 +1,122 @@
...
+import { MAP_POSITION_STORAGE_KEY } from '../../../../ducks/map-position';
+import { PATROL_FILTER_STORAGE_KEY } from '../../../../ducks/patrol-filter';
+import { SETTINGS_CATEGORY, trackEventFactory } from '../../../../utils/analytics';
+import { useOptionalPersistence } from '../../../../reducers/storage-config';
+
+import * as styles from '../../styles.module.scss';
+
+const settingsTracker = trackEventFactory(SETTINGS_CATEGORY);
+
+const AppRefreshFieldSet = () => {
```
luixlive, line 15: I upgraded our General tab UI to match the latest designs (and the Map tab). Next you will see many "new files" that aren't adding anything new really, but I'm just dividing our code better, naming components, and adding improved UI / UX.

This file is for the "App Refresh" field set of the General Settings:
image

src/SideBar/SettingsPane/GeneralTab/AppRefreshFieldSet/index.test.js
```diff
@@ -0,0 +1,86 @@
+import React from 'react';
+import userEvent from '@testing-library/user-event';
+
+import { render, screen } from '../../../../test-utils';
+
+import AppRefreshFieldSet from './';
+
+describe('SideBar - SettingsPane - GeneralTab - AppRefreshFieldSet', () => {
```
luixlive, line 8: And I added / improved the unit tests of all these settings!

src/SideBar/SettingsPane/GeneralTab/ExperimentalFeaturesFieldSet/index.js
```diff
@@ -0,0 +1,97 @@
...
+import { setExperimentalFeatures } from '../../../../ducks/experimental-features';
+import { SETTINGS_CATEGORY, trackEventFactory } from '../../../../utils/analytics';
+
+import * as styles from '../../styles.module.scss';
+
+const settingsTracker = trackEventFactory(SETTINGS_CATEGORY);
+
+export const EXPERIMENTA_FEATURES_QUERY_PARAMETER = 'ef';
+
+const ExperimentalFeaturesFieldSet = () => {
```
luixlive, line 16: The field set for the experimental features:
image

This is actually the file with the work related to the ticket. Basically a new logic to show / hide experimental feature settings through a new query paramater `ef` (experimental features). Hopefully the code + comments are clear enough.

src/SideBar/SettingsPane/GeneralTab/LanguageSelect/index.js
```diff
@@ -0,0 +1,38 @@
...
+import * as styles from '../../styles.module.scss';
+
+const settingsTracker = trackEventFactory(SETTINGS_CATEGORY);
+
+const LANGUAGE_OPTIONS = Object.entries(SUPPORTED_LANGUAGES).reduce(
+  (accumulator, [value, label]) => [...accumulator, { label, value }],
+  []
+);
+
+const LanguageSelect = () => {
```
luixlive, line 17: Language select field:
image

src/SideBar/SettingsPane/GeneralTab/SoundFieldSet/index.js
```diff
@@ -0,0 +1,103 @@
...
+  setPlaySoundForNewEvents,
+  setPlaySoundForNewInReachMessages,
+  setPlaySoundForRadioStateChangeToRed,
+} from '../../../../ducks/user-preferences';
+
+import * as styles from '../../styles.module.scss';
+
+const settingsTracker = trackEventFactory(SETTINGS_CATEGORY);
+
+const SoundFieldSet = () => {
```
luixlive, line 16: And the new sound field set:
image

src/SideBar/SettingsPane/MapTab/MainMapSettingsView/DisplayFieldSet/index.js
```diff
@@ -81,12 +81,12 @@ const DisplayFieldSet = () => {
         <input
           checked={showTrackTimepoints}
           className={styles.checkbox}
-          id="show-track-timepoints-checkbox"
+          id="map-display-show-track-timepoints-checkbox"
```
luixlive, line 84: I added a prefix to the setting checkbox ids with the name of the tab and the field set to make sure each id is unique.

src/SideBar/SettingsPane/MapTab/MainMapSettingsView/styles.module.scss
luixlive, line 1: I used to have the styling of the checkboxes at the `MapTab` level, but now that they are also reused by the `GeneralTab` subcomponents, I lifted the SCSS module up.

src/SideBar/SettingsPane/index.js
```diff
@@ -1,132 +1,39 @@
...
 const ALERTS_URL = `${DAS_HOST}/alerts`;
 
-const TAB_KEYS = {
-  GENERAL: 'general',
-  MAP: 'map',
-  ALERTS: 'alerts',
-};
+const TAB_KEYS = { ALERTS: 'alerts', GENERAL: 'general', MAP: 'map' };
 
 const SettingsPane = () => {
```
luixlive, line 18: The settings pane was simplified to be just responsible of rendering the tabs and the tab panels, but it no longer implements any child logic:
image

src/SoundNotificationsPlayer/RadioStateChangeToAlarmNotificationPlayer/index.js
```diff
@@ -35,8 +35,12 @@ const AlarmRadioStateToastMessage = ({ onClickJumpToLocation, subject }) => {
 const RadioStateChangeToAlarmNotificationPlayer = ({ onPlayNotificationSound }) => {
   const jumpToLocation = useJumpToLocation();
 
-  const subjectsWithAlarmRadioState = useSelector((state) => Object.values(state.data.subjectStore)
-    .filter((subject) => subject?.last_position_status?.radio_state === 'alarm'));
+  const subjectStore = useSelector((state) => state.data.subjectStore);
```
luixlive, line 38: Since I was already touching the tests of these files, I decided to clean a redux selector warning by implementing the selector correctly (avoid renders by returning new objects unnecessarily from the selector).

src/ducks/experimental-features/index.js
```diff
@@ -0,0 +1,28 @@
...
+// Action creators
+export const setExperimentalFeatures = (experimentalFeatures) => ({
+  payload: experimentalFeatures,
+  type: SET_EXPERIMENTAL_FEATURES,
+});
+
+// Reducer
+export const INITIAL_STATE = {};
+
+const experimentalFeaturesReducer = (state = INITIAL_STATE, action) => {
```
luixlive, line 15: New reducer to hold the experimental features the user has already enabled and modified. It is persisted, but it won't require migrations since it fixes itself by making sure only valid development feature flags are stored every time it gets modified (see `SET_EXPERIMENTAL_FEATURES` action handling below). It's state is a simple object with the feature keys and a boolean that tells if the checkbox is checked or not:
```JavaScript
{
  CUSTOM_COORDINATE_SYSTEMS_ENABLED: true,
}
```

src/ducks/user-preferences/index.js
```diff
@@ -0,0 +1,62 @@
...
+  autoStartPatrols: false,
+  enable3D: true,
+  gpsFormat: Object.values(GPS_FORMATS)[0],
+  hideModals: false,
+  playSoundForNewEvents: false,
+  playSoundForNewInReachMessages: false,
+  playSoundForRadioStateChangeToRed: false,
+};
+
+const userPreferencesReducer = (state = INITIAL_STATE, action) => {
```
luixlive, line 43: This reducer isn't new, but I moved it to a folder so I can add unit tests. All I did is added the new state properties `playSoundForNewEvents`, `playSoundForNewInReachMessages` and `playSoundForRadioStateChangeToRed`, and some action creators to set them cleanly.

src/SoundNotificationsPlayer/index.js
```diff
@@ -18,18 +13,16 @@ export const SOUND_DEBOUNCE_TIME = 5000;
...
   const [play] = useSound(ding);
 
-  const canPlayNotificationSound = useRef(true);
-
-  const isNewEventNotificationSoundEnabled = useSelector(
-    (state) => !!state.view.featureFlagOverrides?.[ENABLE_NEW_REPORT_NOTIFICATION_SOUND]?.value
-  );
-  const isNewInReachMessageNotificationSoundEnabled = useSelector(
-    (state) => !!state.view.featureFlagOverrides?.[ENABLE_NEW_IN_REACH_MESSAGE_NOTIFICATION_SOUND]?.value
+  const playSoundForNewEvents = useSelector((state) => state.view.userPreferences.playSoundForNewEvents);
```
luixlive, line 16: I didn't do any logical change to this component. I just updated the source from where we get if the features are enabled or not. We used to read from the now deleted `featureFlagOverrides`, because they were still considered experimental features. But now we hold the enabled / disabled values of these features under the `userPreferences` state property.

src/hooks/index.js
```diff
@@ -11,17 +11,15 @@ import { DEVELOPMENT_FEATURE_FLAGS } from '../constants';
 export const useSystemConfigFlag = (flag) => useSelector((state) => !!state?.view?.systemConfig?.[flag]);
 
 export const useFeatureFlag = (flagName) => {
-  const featureFlagOverrides = useSelector(state =>
-    state.view.featureFlagOverrides
-  );
+  const experimentalFeatures = useSelector((state) => state.view.experimentalFeatures) || {};
```
luixlive, line 14: Small update in our `useFeatureFlag` hook  to read the override values of the feature flags from the new `experimentalFeatures` substore .

src/utils/analytics.js
```diff
@@ -36,6 +36,7 @@ export const GPS_FORMAT_CATEGORY = 'GPS Format';
 export const ALERTS_CATEGORY = 'Alerts';
 export const FEED_CATEGORY = 'Feed';
 export const BETA_PREVIEW_CATEGORY = 'Beta Preview';
+export const SETTINGS_CATEGORY = 'Settings';
```
luixlive, line 39: We didn't have any analytics for all the settings I updated in this PR. But I thought it could be useful to have a log of how many users enable / disable the fields, change language, etc... So I added this new category and a tracking event for each checkbox `onChange`. 

I just have a question here for @JoshuaVulcan . All the changes here are for the "General Settings Tab", but in the other tab that I already re-worked "Map Settings Tab", we are using an analytics category: `export const MAP_INTERACTION_CATEGORY = 'Map Interaction';`. Which in my opinion, isn't appropriate. I would suggest to use this new one too for the map settings (and basically all settings), but I didn't want to change it without asking because I know this will mean now "breaking" with the previous contract and all the stored data. I feel like it's ok because anyway we didn't have lots of tracking events for these settings as I already mentioned. I just added them, so we can basically start from scratch capturing all the setting udpate events. What do you think?

gaboDev (approved):

PR #1381, opened 2025-09-19
ERA-11913: Default "App Refresh" settings to enabled

gaboDev:
### What does this PR do?
- Adds `defaultPersistenceConfig` param to extend `useOptionalPersistence` in order to set app refreshes settings to true by default.

### Relevant link(s)
* Tracking tickets: [ERA-11913](https://allenai.atlassian.net/browse/ERA-11913)
* [Hosted demo environments](https://era-11913.dev.pamdas.org)

[ERA-11913]: https://allenai.atlassian.net/browse/ERA-11913?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/reducers/storage-config.js
```diff
@@ -44,10 +44,10 @@ export const generateOptionalStorageConfig = (key, INITIAL_STATE) => {
   return storageConfig;
 };
 
-export const useOptionalPersistence = (key) => {
+export const useOptionalPersistence = (key, defaultPersistenceConfig = { restore: false }) => {
```
luixlive: My only suggestion is to stick to the existing convention with the new parameter. If you see below, this hook returns a `setRestorable` function to change the `restore` value (the parameter is even called that way). Why not to do the same with the default? Instead of expecting an object like `{ restore: Boolean }` we can just expect the boolean like `setRestorable`. And I would even suggest to match the naming:
```JavaScript
export const useOptionalPersistence = (key, defaultRestorable = false) => {
  ...
  const [value, setValue] = useLocalStorage(namespace, { restore: defaultRestorable });
  ...
};
```

That  way `useOptionalPersistence(key, defaultRestorable)` and `setRestorable(restorable)` wording and functionality match.

luixlive (commented):
Simple 💯  Just a small suggestion, but also note that the changes related to the new beta toggle mechanism caused conflicts here (which is expected). Solving conflicts + the suggestion should be easy 👍

JoshuaVulcan:
@aclainap tagging you for visibility: we'll want to be really strong on comms for this, since it updates a default app setting. 
Something like "EarthRanger now saves your working state, even when you refresh the page! You can disable it by blah blah blah"

luixlive (approved):
@JoshuaVulcan it's simple, I don't have any more feedback so I guess we can just merge it but I'd wait for your review so it's not only me (now the author is no longer here).

PR #1387, opened 2025-10-10
[ERA-8348] Observation/Track Vector Layer

JoshuaVulcan:


public/locales/en-US/tracks.json
```diff
@@ -5,11 +5,19 @@
     "trackLegendItemsName": "patrols"
   },
   "subjectTrackLegend": {
-    "description": "{{pointCount}} points over {{trackTime}}",
+    "description": "{{trackTime}}",
```
luixlive: Wait what? Hahaha this doesn't have a translation text, is just printing a variable. We could just remove the translation entirely:
```Javascript
// Before
t('description', { trackTime })

// After
trackTime
```

src/constants/index.js
```diff
@@ -383,12 +389,12 @@ export const TIME_OF_DAY_PERIODS = [
   {
     rangeString: '06:01 - 09:00',
     rangeMinutesMin: 361,
-    rangeMinutesMax: 54,
+    rangeMinutesMax: 540,
```
luixlive, line 399: Why are we modifying these values?

src/ducks/subjects.js
```diff
@@ -52,11 +52,20 @@ const cancelableMapSubjectsFetch = () => {
       cancelToken.cancel();
       cancelToken = CancelToken.source();
 
+      // When not in timeslider mode, only fetch subjects updated in the last
+      // hour — stale subjects are already rendered from the vector tile layer.
+      // The 1-hour window matches selectFreshSubjectIds in selectors/subjects.
```
luixlive: I'd remove the `The 1-hour window` because we may change that value later in the constant and this comment would be outdated.

Also, defining an object just optionally apend a property and then spread it seems weird. Why not to instead simply calculate the `updated_since` parameter?
```Javascript
let updated_since = params?.updated_since;
if (!updated_since && !timeSliderActive) {
  // The parameters didn't provide an updated_since value and the time slider is inactive
  // Set the fresh subject window.
  updated_since = new Date(Date.now() - FRESH_SUBJECT_WINDOW_MS).toISOString();
}
```

src/Map/layers/useCrsBoundingBoxLayer/index.js
```diff
@@ -116,8 +117,8 @@ const useCrsBoundingBoxLayer = () => {
       }
 
       return () => {
-        map.removeLayer(LAYER_IDS.COORDINATE_REFERENCE_SYSTEM_BBOX);
-        map.removeSource(SOURCE_IDS.COORDINATE_REFERENCE_SYSTEM_BBOX);
+        safeRemoveMapLayer(map, LAYER_IDS.COORDINATE_REFERENCE_SYSTEM_BBOX);
```
luixlive, line 120: The job is already done, but I think we could have simply added an optional chaining operator right? `map?.removeLayer();`. Jeje.

Of course that wouldn't help for any errors thrown by `map`, but we shouldn't hit any if properly handling the sources and layer. Which brings me back to the conversation of how unreliable the `useMapLayer` and `useMapSource` hooks are. I have the guess that we needed to add `safeRemoveMapLayer` just to avoid issues with those hooks poor map management. The name `safeRemoveMapLayer` implies that we are actually making sure to properly remove the layer, but we are not. If the map is defined and an error is thrown, we just ignore it, post a log, and go on without really cleaning the map layer or source.

src/Map/index.js
```diff
@@ -134,7 +137,7 @@ const Map = ({ children, onMapLoad, socket }) => {
   const patrolTrackState = useSelector(state => state.view.patrolTrackState);
   const popup = useSelector(state => state.view.popup);
   const showReportHeatmap = useSelector(state => state.view.showReportHeatmap);
-  const showTrackTimepoints = useSelector(state => state.view.showTrackTimepoints);
+  // const showTrackTimepoints = useSelector(state => state.view.showTrackTimepoints);
```
luixlive: Why to keep this code commented? Should we remove it? I see more commented lines below.

src/Map/index.js
```diff
@@ -553,6 +556,16 @@ const Map = ({ children, onMapLoad, socket }) => {
     }
   }, [fetchMapData, map, timeSliderState.active]);
 
+  useEffect(() => {
+    if (!map || !subjectsEnabled) return;
+    overlayCancelToken.current.cancel();
```
luixlive, line 558: We are cancelling before a new fetch and after the hook is unmounted. Do we need that redundancy?

src/Map/index.js
```diff
@@ -690,7 +703,13 @@ const Map = ({ children, onMapLoad, socket }) => {
         bounceEventIDs={bounceEventIDs}
       />}
 
-      {subjectsEnabled && <SubjectsLayer mapImages={mapImages} onSubjectClick={onSelectSubject} />}
+      {subjectsEnabled && !timeSliderActive && <SubjectTileLayer onSubjectClick={onSelectSubject} />}
```
luixlive: Maybe I'm just not understanding correctly, but if subjects are enabled and time slider is off, we will render two subject layers. The old one and the tiles one. Shouldn't it be one or the other?

src/MessageList/MessageListItem.js
```diff
@@ -57,7 +57,7 @@ const MessageListItem = (props) => {
 
           const subjectTrackHidden = !uniq([...subjectTrackState.visible, ...subjectTrackState.pinned]).includes(subject.id);
           if (subjectTrackHidden) {
-            await fetchTracksIfNecessary([subject.id]);
+            // await fetchTracksIfNecessary([subject.id]);
```
luixlive: 👀

src/RealtimeOverlayLayer/index.js
```diff
@@ -0,0 +1,184 @@
...
+            'case',
+            ['==', ['get', 'subject_subtype'], 'ropeless_buoy_gearset'], 0.5,
+            1,
+          ],
+        },
+      });
+    }
+
+    return () => {
+      if (!map) return;
```
luixlive: I see defensive code here that isn't necessary. `safeRemoveMapLayer` and `safeRemoveMapSource` do the same validations.

src/selectors/tracks/index.js
```diff
@@ -12,14 +12,14 @@ import {
 const selectEventFilterLowerDateRange = (state) => state.data.eventFilter.filter.date_range.lower;
 const selectHeatmapSubjectIDs = (state) => state.view.heatmapSubjectIDs;
 const selectIsTimeOfDayColoringActive = (state) => state.view.trackSettings.isTimeOfDayColoringActive;
-const selectSubjectTrackState = (state) => state.view.subjectTrackState;
+export const selectSubjectTrackState = (state) => state.view.subjectTrackState;
```
luixlive: nit: These are trivial selectors, why to export them? They don't save lines:
```Javascript
useSelector((state) => state.view.subjectTrackState);
useSelector(selectSubjectTrackState);
```

src/selectors/tracks/index.js
```diff
@@ -28,6 +28,17 @@ const selectTrackLengthInDays = createSelector(
...
 
+/**
+ * Vector tile API only accepts two range values: "45" (last 45 days) or "all".
+ * Client keeps fine-grained track length; this maps to the param when building the tile URL.
+ * Use the smallest range that covers the requested length: <= 45 days → "45", > 45 → "all".
+ * The client filters segments to the track time envelope (start_time) so 21 days still shows 21 days.
+ */
+export const selectVectorTileRangeParam = createSelector(
+  [selectTrackLengthInDays],
+  (trackLengthInDays) => (trackLengthInDays <= 45 ? '45' : 'all')
```
luixlive: I think this falls under the statement of "memoization costs being higher than the actual processing". I don't think we need a selector for a trivial condition 🤔 
```Javascript
const trackLengthInDays = useSelector(selectTrackLengthInDays);

const vectorTileRangeParam = trackLengthInDays <= 45 ? '45' : 'all';
```

src/SideBar/SettingsPane/GeneralTab/AppRefreshFieldSet/index.js
```diff
@@ -20,6 +20,12 @@ const AppRefreshFieldSet = () => {
   const { restorable: mapPositionRestorable, setRestorable: setMapPositionIsRestorable } = useOptionalPersistence(MAP_POSITION_STORAGE_KEY, true);
   const { restorable: patrolFilterRestorable, setRestorable: setPatrolFilterIsRestorable } = useOptionalPersistence(PATROL_FILTER_STORAGE_KEY, true);
 
+  useEffect(() => {
+    if (mapPositionRestorable === undefined) {
+      setMapPositionIsRestorable(true);
```
luixlive: I don't understand why we need this `useEffect`, and I think it confused Copilot too.

src/SubjectTileLayer/index.js
```diff
@@ -0,0 +1,219 @@
...
+  const rangeParam = useSelector(selectVectorTileRangeParam);
+  const showInactiveRadios = useSelector((state) => state.view.showInactiveRadios);
+  const subjectStore = useSelector((state) => state.data.subjectStore);
+
+  // Ref so the click handler always reads the latest store without
+  // re-creating the memoised callback on every store update.
+  const subjectStoreRef = useRef(subjectStore);
+  subjectStoreRef.current = subjectStore;
+
+  /* ── source & layers ──────────────────────────────────────────────── */
```
luixlive: nit: not a big fan of these new "------- comments -----------" haha. The rest of the files semas clean 👍

src/SubjectTrackLegend/index.js
```diff
@@ -17,68 +24,154 @@ import * as styles from './styles.module.scss';
...
+ * Fetches track GeoJSON for the given subjects and date range, counts points, then discards the response.
+ * Returns total point count; does not store GeoJSON (no memory bloat).
+ */
+const fetchTrackPointCount = async (subjectIds, from, until, signal) => {
+  if (!subjectIds?.length || !from) return 0;
+  const since = from.toISOString();
+  const untilParam = until ? until.toISOString() : new Date().toISOString();
+  const params = { since, until: untilParam };
+  const responses = await Promise.all(
+    subjectIds.map((id) => axios.get(TRACKS_API_URL(id), { params, signal }))
```
luixlive: Why are we doing axios calls here? We should use a thunk action creator and sanitize / store the data in a reducer.

src/SubjectTrackLegend/index.js
```diff
@@ -17,68 +24,154 @@ import * as styles from './styles.module.scss';
...
+  );
+  let total = 0;
+  for (const res of responses) {
+    const fc = res?.data?.data;
+    if (fc) total += countTrackPointsInFeatureCollection(fc);
+  }
+  return total;
+};
+
+const TrackDetailsTooltipContent = memo(function TrackDetailsTooltipContent({
```
luixlive: Moving this to a subfolder would keep this file cleaner and easier to read! Also, weird to see a `function` here. This repository favors arrow functions always. AI agent?

src/utils/datetime.js
```diff
@@ -162,3 +162,18 @@ export const getTimeInTimezone = (date, timeZone) => new Intl.DateTimeFormat('en
   hourCycle: 'h23',
 }).format(date);
 
+/**
```
luixlive: Why the JSdocs here? Haha

I see variables named `d`, `h`, `m`. From the context it's obvious what they are, but readability should be preferred over conciseness.

src/utils/map.js
```diff
@@ -213,32 +213,51 @@ export const calculatePopoverPlacement = async (map, popoverLocation) => {
...
-
-  const minLon = Math.min(lon1, lon2);
-  const maxLon = Math.max(lon1, lon2);
-  const minLat = Math.min(lat1, lat2);
-  const maxLat = Math.max(lat1, lat2);
-  return [minLon, minLat, maxLon, maxLat];
 };
 
-
+/**
```
luixlive: Another giant JSdoc block I guess was added by an agent 🤔 

I see many numbers and even Hex colors hardcoded here. We don't have to move everything to constants of course, some are very intuitive, but honestly this code block is a bit hard to read.

.gitignore
```diff
@@ -1,18 +1,98 @@
...
-!.yarn/plugins
-!.yarn/releases
-!.yarn/sdks
-!.yarn/versions
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+
+yarn
+.yarn/install-state.gz
```
luixlive, line 19: This was already handled. Solving conflicts should bring a total update to our gitignore.

luixlive (changes requested):
Ok, this is the first pass and I didn't hold on any comments. With all the new sources, layers, filters, and all the old code that was touched, I'd say we really need thorough manual testing here. But the branch env is down: https://era-8348.dev.pamdas.org/ . Is there an env where I can test these changes?

About my review, I see a lot of commented code, Copilot already signaled it, and also lots of new code without a single test. I think the a way to summarize my review is that I notice inconsistencies in the code, between the new code and our existing practices, but also comparing the new code against itself. I guess I've been a bit defensive about moving to a total vibe-code approach and just let the AI do stuff based on only prompting. But I guess we are moving towards that. If we are going to do it, I think we need to establish practices, and definitely go with smaller PRs. This PR has big undocumented code blocks, without tests, and with obscure variable names like `x`; and some other blocks with a JSDoc twice the size of the function itself. It feels random, and it's hard to review.

.gitignore
JoshuaVulcan: Done 🏁

public/locales/en-US/tracks.json
JoshuaVulcan: This is still in-progress 😆  yes it would be silly to just have the variable there.

JoshuaVulcan:
@luixlive thanks for the review! I'll get going on the first pass of changes, and in the meantime the env should be building correctly now.

luixlive:
I was able to log to the environment now! It's looking great overall, I just have a couple feedback items, all of them very small:

**1-** The changes in `SubjectTrackLegend` feel a bit like a degradation, and the new icon button doesn't follow our UI / UX.
Before
image

After
image

You can see that before, we had the amount of points information visible and available at all times. Now it is hidden behind a click to that info icon button. And the button itself looks very different from our nice looking and accessible buttons at the right. Those have accessible hover and focus state, and can be navigated with the keyboard. The new one doesn't.
image

**2-** Small UI issue, but we should make the numeric inputs of the new controls wider since some numbers don't fit:
image

Also, now that we have a bigger form, I feel like "Track Length" and "Segmentation Settings" texts should be bold or bigger because they get lost with the labels.

**3-** Playing with the new segmentation controls, I see that we can break the track lines, is that expected? I'm not entirely sure if I understand what the segment settings should do, but I would expect it to filter in / out certain observations, but keeping the lines connected.
image

**4-** Not really a feedback point, but worth mentioning. I don't know how to be sure, or tell if I properly tested some of the most important things of this PR: realtime overlay, fresh subject retrieval window, even the segment controls (easy to find and play with them, but I'm not sure if I'm understanding what they should do in order to properly validate it). A little bit of guidance on how to be aware of the functionality of those items would help.

PR #1389, opened 2025-10-15
[ERA-12023] using `buffer` instead of `circle` for proximity analyzer display...

JoshuaVulcan:
…now that we support non-point-typed proximity analyzer geometries.

### What does this PR do?
- See title

### How does it look
- N/A

### Relevant link(s)
* https://allenai.atlassian.net/browse/ERA-12023
* Demo env building

- This is, by design, the smallest possible change to achieve the desired outcome.

- We recently added back-end support for proximity analyzers with all geometry types. This is the front-end compliment to that.


.github/copilot-instructions.md
```diff
@@ -0,0 +1,108 @@
+# EarthRanger Web - AI Agent Instructions
```
luixlive, line 1: Amazing! This will be very helpful as context for agents before prompting. I would be glad to take a deeper look, see if I can add anything, refine wording, reduce characters (to not consume too much agent context), etc... As a separate task. But love it so far 💯

src/utils/analyzers.js
```diff
@@ -65,12 +65,10 @@ export const getAnalyzerAdminPoint = (geometry) => {
   return centerPt.geometry.coordinates;
 };
 
-// use turf.circle to construct a GEOJson Feature of type polygon
-// increase/decrease steps will affect the render fps
-export const createGeoJSONCircle = (geometry, radius, options) => {
-  if (!options) options = { steps: 32, units: 'kilometers' };
-  const poly_circle = circle(centroid(geometry), radius / 1000, options);
-  return poly_circle;
+export const createGeoJSONBuffer = (geojson, radius = 500, opts = {}) => {
```
luixlive: nit: IMO, this method is an overkill for 2 reasons:
1- It could be a one liner:
```Javascript
export const createGeoJSONBuffer = (geojson, radius = 500, opts = {}) =>
  buffer(geojson, radius / 1000, { steps: 32, ...opts, units: 'kilometers' });
```
Which means we are not really saving a lot of space.

2- It's only being used in one single place, which means we are not really reusing this method.

It makes no harm, but feels like more code to maintain. I would just keep it simple and just have the code directly in `src/ducks/analyzers.js`:
```Javascript
const proximityBufferPolygon = buffer(
  feature,
  analyzer.threshold_dist_meters / 1000,
  { steps: 32, units: 'kilometers' }
};
```

That feels simple and keeps the logic in one place so I don't have to go and open another file when debugging or updating something in this file.

luixlive (dismissed):
Thanks to that agent instructions file, now we could ask them to re-do our `README.md` which honestly is not very useful right now 😅  I'll be glad to work on refining even more the context file (if I find something, because it looks so good already) and build the readme in a separate ticket. Either this sprint or the next, when I have a bit of spare time.

luixlive (approved):
Awesome!

PR #1392, opened 2025-10-21
ERA-12053: Handle schema rendering errors (event details)

luixlive:
### What does this PR do?
Fix the event schema loading spinner logic and add an error message if the schema request fails.

### How does it look
image

### Relevant link(s)
* [ERA-12053](ERA-12053)
* [Env](https://era-12053.dev.pamdas.org)

### Notes
Since there weren't any designs for this task, I added a simple error message. Notice that the second part of the error message is a string that comes from the backend response, so it won't be translated with our i18n files.

[ERA-12053]: https://allenai.atlassian.net/browse/ERA-12053?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ReportManager/DetailsSection/index.js
+++ b/src/ReportManager/DetailsSection/index.js
@@ -302,11 +302,13 @@ const DetailsSection = ({
       role="alert"
       aria-live="polite"
     >
-      <strong>{t('errorLoadingSchema')}</strong>
+      <p>
+        <strong>{t('errorLoadingSchema')}</strong>
 
-      {eventSchema?.response?.data?.status?.detail && <span>
-        {eventSchema.response.data.status.detail}
-      </span>}
+        {eventSchema?.response?.data?.status?.detail && <span className={styles.errorDetail}>
+          {eventSchema.response.data.status.detail}
+        </span>}
+      </p>
     </div>}
   </div>;
 };
--- a/src/ReportManager/DetailsSection/index.js
+++ b/src/ReportManager/DetailsSection/index.js
@@ -299,13 +299,13 @@ const DetailsSection = ({
 
     {eventSchema instanceof Error && <div
       aria-live="polite"
-      className={styles.errorMessage}
+      className={styles.errorMessageWrapper}
       role="alert"
     >
-      <p>
+      <p className={styles.errorMessage}>
         <strong>{t('errorLoadingSchema')}</strong>
 
-        {eventSchema?.response?.data?.status?.detail && <span className={styles.errorDetail}>
+        {eventSchema?.response?.data?.status?.detail && <span>
           {eventSchema.response.data.status.detail}
         </span>}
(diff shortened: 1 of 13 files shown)


src/ducks/event-schemas/index.js
```diff
@@ -126,7 +126,7 @@ const eventSchemasReducer = (state, action) => {
       loading: false,
       [action.payload.eventTypeValue]: {
         ...state[action.payload.eventTypeValue],
-        [action.payload.eventId]: action.payload.error,
+        [action.payload.eventId || 'base']: action.payload.error,
```
luixlive, line 129: This was the cause of the infinite loader. When calling the fetch schema action creator, there are two possibilites:
1- We are fetching a schema of an existing event, so we pass the event id to get the cached version of that schema.
2- We are fetching a schema for a new event that the user wants to create, so there's no event id.

In the second scenario, when fetching the schema either receiving the schema successfully or an error, we store the data under the `state.data.eventSchemas[eventTypeValue][eventId || 'base']`. In the case of a new event, we store it under "base" because there is no event id. However we forgot to add the "base" key when the request fails, and the error was being saved under `state.data.eventSchemas[eventTypeValue].undefined`, so one of the constraints to show the loader was still passing.

src/ReportManager/DetailsSection/index.js
```diff
@@ -290,13 +289,27 @@ const DetailsSection = ({
...
-    {!eventSchema && !reportForm.is_collection && loadingSchema && <div className={styles.loaderWrapper}>
+    {!eventSchema && !reportForm.is_collection && loadingEventSchemas && <div className={styles.loaderWrapper}>
       <MoonLoader
         color={LOADER_COLOR}
         data-testid="reportManager-detailsSection-loader"
         size={LOADER_SIZE}
       />
     </div>}
+
+    {eventSchema instanceof Error && <div
```
luixlive, line 300: New error message shown in case the schema fetch request fails.

src/ReportManager/ReportDetailView/index.js
```diff
@@ -171,7 +170,6 @@ const ReportDetailView = ({
   const isActive = isReportActive(originalReport);
   const isCollection = !!reportForm?.is_collection;
   const isCollectionChild = eventBelongsToCollection(reportForm);
-  const isLoadingSchemas = (!!reportForm && !eventSchema) || !!loadingEventSchemas;
```
luixlive, line 174: I simplified the logic to show the loader or not. For starters, it's no longer calculated in the parent (`ReportDetailView`) but in directly in `DetailsSection`. So I removed that logic from here.

JoshuaVulcan (approved):

PR #1394, opened 2025-10-22
[ERA-12142] Vector tile caching

JoshuaVulcan:
### What does this PR do?
- Due to the necessity of the Vary: Authorization headers of our tile service responses, we must rely on custom code to ensure disk-level caching for our vector tile data.

### How does it look
- N/A

### Relevant link(s)
* https://allenai.atlassian.net/browse/ERA-12142
* env building

- sw-custom

### DIFF
--- a/.github/workflows/develop-workflow.yml
+++ b/.github/workflows/develop-workflow.yml
@@ -49,7 +49,7 @@ jobs:
         run: ci-helpers/makeversion.py
 
       - name: Install dependencies
-        run: yarn --ignore-scripts
+        run: yarn install
 
       - name: Run tests
         run: yarn test-ci
--- a/react-app-config/webpack.config.js
+++ b/react-app-config/webpack.config.js
@@ -670,7 +670,7 @@ module.exports = function (webpackEnv) {
           // Bump up the default maximum size (2mb) that's precached,
           // to make lazy-loading failure scenarios less likely.
           // See https://github.com/cra-template/pwa/issues/13#issuecomment-722667270
-          maximumFileSizeToCacheInBytes: 5 * 1024 * 1024,
+          maximumFileSizeToCacheInBytes: 10 * 1024 * 1024,
         }),
       // TypeScript type checking
       useTypeScript &&
--- a/src/App.js
+++ b/src/App.js
@@ -65,7 +65,7 @@ export const App = () => {
   );
   const trackSettings = useSelector((state) => state.view.trackSettings);
 
-  // Send scope to service worker for tile caching
+  // set user scope for service worker caching
   useEffect(() => {
     if (navigator?.serviceWorker?.controller && user?.id) {
       navigator.serviceWorker.controller.postMessage({
--- a/src/App.js
+++ b/src/App.js
@@ -65,27 +65,6 @@ export const App = () => {
   );
   const trackSettings = useSelector((state) => state.view.trackSettings);
 
-  // set user scope for service worker caching
-  useEffect(() => {
-    if (navigator?.serviceWorker?.controller) {
-      if (user?.id) {
-        const scopeHash = selectedUserProfile?.id ?? user.id;
-        console.log('Setting SW scope to:', scopeHash);
-        navigator.serviceWorker.controller.postMessage({
-          type: 'SET_SCOPE',
-          scope: { hash: scopeHash }
-        });
-      } else {
-        console.log('Clearing SW scope - no user');
-        // Clear scope when user logs out
-        navigator.serviceWorker.controller.postMessage({
-          type: 'SET_SCOPE',
-          scope: { hash: null }
-        });
-      }
-    }
-  }, [user, selectedUserProfile]);
-
   const socket = useContext(SocketContext);
 
   const [isDragging, setDragState] = useState(false);
@@ -129,6 +108,28 @@ export const App = () => {
     disallowDragAndDrop(e);
     finishDrag(e);
   }, [disallowDragAndDrop, finishDrag]);
+
+
+  // set user scope for service worker caching
+  useEffect(() => {
+    if (navigator?.serviceWorker?.controller) {
+      if (user?.id) {
+        const scopeHash = selectedUserProfile?.id ?? user.id;
+        console.log('Setting SW scope to:', scopeHash);
+        navigator.serviceWorker.controller.postMessage({
+          type: 'SET_SCOPE',
+          scope: { hash: scopeHash }
+        });
+      } else {
+        console.log('Clearing SW scope - no user');
+        // Clear scope when user logs out
+        navigator.serviceWorker.controller.postMessage({
+          type: 'SET_SCOPE',
+          scope: { hash: null }
+        });
+      }
+    }
+  }, [user, selectedUserProfile]);
 
   useEffect(() => {
     /* use these catch blocks to provide error toasts if/as desired */
--- a/src/App.js
+++ b/src/App.js
@@ -67,11 +67,22 @@ export const App = () => {
 
   // set user scope for service worker caching
   useEffect(() => {
-    if (navigator?.serviceWorker?.controller && user?.id) {
-      navigator.serviceWorker.controller.postMessage({
-        type: 'SET_SCOPE',
-        scope: { hash: selectedUserProfile?.id ?? user.id }
-      });
+    if (navigator?.serviceWorker?.controller) {
+      if (user?.id) {
+        const scopeHash = selectedUserProfile?.id ?? user.id;
+        console.log('Setting SW scope to:', scopeHash);
+        navigator.serviceWorker.controller.postMessage({
+          type: 'SET_SCOPE',
+          scope: { hash: scopeHash }
+        });
+      } else {
+        console.log('Clearing SW scope - no user');
+        // Clear scope when user logs out
+        navigator.serviceWorker.controller.postMessage({
+          type: 'SET_SCOPE',
+          scope: { hash: null }
+        });
+      }
     }
   }, [user, selectedUserProfile]);
 
--- a/src/App.js
+++ b/src/App.js
@@ -115,13 +115,11 @@ export const App = () => {
     if (navigator?.serviceWorker?.controller) {
       if (user?.id) {
         const scopeHash = selectedUserProfile?.id ?? user.id;
-        console.log('Setting SW scope to:', scopeHash);
         navigator.serviceWorker.controller.postMessage({
           type: 'SET_SCOPE',
           scope: { hash: scopeHash }
         });
       } else {
-        console.log('Clearing SW scope - no user');
         // Clear scope when user logs out
         navigator.serviceWorker.controller.postMessage({
           type: 'SET_SCOPE',
--- a/src/App.js
+++ b/src/App.js
@@ -65,6 +65,16 @@ export const App = () => {
   );
   const trackSettings = useSelector((state) => state.view.trackSettings);
 
+  // Send scope to service worker for tile caching
+  useEffect(() => {
+    if (navigator?.serviceWorker?.controller && user?.id) {
+      navigator.serviceWorker.controller.postMessage({
+        type: 'SET_SCOPE',
+        scope: { hash: selectedUserProfile?.id ?? user.id }
+      });
+    }
+  }, [user, selectedUserProfile]);
+
   const socket = useContext(SocketContext);
 
   const [isDragging, setDragState] = useState(false);
--- a/src/Map/index.js
+++ b/src/Map/index.js
@@ -605,7 +605,7 @@ const Map = ({ children, onMapLoad, socket }) => {
           }
         }
 
-        // Remove any remaining file extensions from the src path
+        // Sanitize the src path to remove internal icon ID suffixes
         src = src.replace(/(\.svg|\.png|\.jpg).*$/, '$1');
 
         try {
--- a/src/Map/index.js
+++ b/src/Map/index.js
@@ -589,7 +589,6 @@ const Map = ({ children, onMapLoad, socket }) => {
       // querying from the root /static/ dir of the host means this is one of our static assets, let's get it
       // if the map says it's missing.
       if (id.includes('/static/')) {
-        // Parse filepath to extract path and dimensions
         const dimensions = {};
         const match = id.match(/^(.*?)(?:-([^-.]+)-([^-.]+))?$/);
 
--- a/src/Map/index.js
+++ b/src/Map/index.js
@@ -589,9 +589,28 @@ const Map = ({ children, onMapLoad, socket }) => {
       // querying from the root /static/ dir of the host means this is one of our static assets, let's get it
       // if the map says it's missing.
       if (id.includes('/static/')) {
-        const src = id.replace(/(\.svg|\.png|\.jpg).*$/, '$1');
+        // Parse filepath to extract path and dimensions
+        const dimensions = {};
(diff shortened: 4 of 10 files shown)


src/SpatialFeaturesLayer/index.js
```diff
@@ -19,6 +18,8 @@ export const LINES_LAYER_ID = 'spatial-features-lines';
 export const POLYGONS_LAYER_ID = 'spatial-features-polygons';
 // const POLYGONS_LABELS_LAYER_ID = 'spatial-features-polygon-labels';
 
+const BEFORE_LAYER_ID = 'sky';
```
luixlive: Question: is the `sky` layer always available? Even in user's custom basemaps?
By reading the [docs](https://docs.mapbox.com/mapbox-gl-js/api/map/#addlayer-parameters-beforeid) I get the impression that the `beforeId` parameter of `addLayer` must always be an existing layer, or there could be an exception. So just wanted to be sure that you tested that scenario. Or maybe the `sky` layer is always there, not sure.

src/App.js
```diff
@@ -65,6 +65,27 @@ export const App = () => {
   );
   const trackSettings = useSelector((state) => state.view.trackSettings);
 
+  // set user scope for service worker caching
+  useEffect(() => {
```
luixlive: nit: most of our components have a defined internal architecture just to keep them organized and consistent.
```javascript
const component = () => {
  // External library hooks

  // Redux selectors
  const trackSettings = useSelector((state) => state.view.trackSettings);

  // Context / state / refs
  const [map, setMap] = useState(null);

  // Callbacks
  const jumpToStartingLocation = useCallback((map) => {...}, [...]);

  // Lifecycle-attached logic (useEffects)
  useEffect(...);

  return <...>;
};
```

You can see it actually in this file. So this `useEffect` should be a few lines below to keep that consistency.

src/App.js
```diff
@@ -65,6 +65,27 @@ export const App = () => {
...
+    if (navigator?.serviceWorker?.controller) {
+      if (user?.id) {
+        const scopeHash = selectedUserProfile?.id ?? user.id;
+        console.log('Setting SW scope to:', scopeHash);
```
luixlive: The logic of the `useEffect` is clean and concise. But do we want to keep these `console.log`s? Once we know it works well, I feel they are just adding clutter to our terminal (which is already pretty bloated).

src/sw-custom.js
```diff
@@ -13,29 +46,33 @@ if ('function' === typeof importScripts) {
...
-    self.addEventListener('activate', function(event) {
+    self.addEventListener('activate', function (event) {
       event.waitUntil(self.clients.claim()); // Become available to all pages
     });
 
     workbox.core.clientsClaim();
 
-    self.addEventListener( 'fetch', function ( event ) {
-      if ( event.request.url.match( '^.*(\/admin\/).*$' ) ) {
+    self.addEventListener('fetch', function (event) {
```
luixlive: What is the intention of this `fetch` listener? There's a conditional to return `false`, which doesn't mean or do anything. It seems like we can completely remove this listener.

src/sw-custom.js
```diff
@@ -13,29 +46,33 @@ if ('function' === typeof importScripts) {
...
-    workbox.precaching.precacheAndRoute([]);
+    try {
+      workbox.precaching.precacheAndRoute([]);
+    } catch (error) {
+      console.warn('Initial precaching setup failed:', error);
+    }
 
     // Font caching
     workbox.routing.registerRoute(
       new RegExp('https://fonts.(?:.googlepis|gstatic).com/(.*)'),
```
luixlive: Sadly it seems like we haven't been caching our `googleapis` fonts for a while because there is a typo here 😞 `googlepis`

src/sw-custom.js
```diff
@@ -59,20 +96,144 @@ if ('function' === typeof importScripts) {
...
-          new workbox.expiration.Plugin({
+          new workbox.expiration.ExpirationPlugin({
             maxEntries: 60,
             maxAgeSeconds: 20 * 24 * 60 * 60, // 20 Days
           }),
         ],
       })
     );
+
+    // ---- Tiles: user-scoped disk cache (works with cookies or Authorization) ----
```
luixlive: Are these comments AI generated? Probably we can get rid of them jeje

src/sw-custom.js
```diff
@@ -59,20 +96,144 @@ if ('function' === typeof importScripts) {
...
+    // Intercept only your tile requests. Adjust the path/pattern if needed.
+    // Examples:
+    //   ^/tiles/          -> relative to your origin
+    //   https://api.example.com/tiles/
+    const tileRouteMatch = ({ url, request }) => {
```
luixlive, line 97: I see very different code styles within this file. In some places we use arrow functions, in some other regular functions. Also some functions are locally defined, others are at the top of the file. This file could use a bit of organization and consistent styling, and I'm pretty sure a LLM agent could do it in ~10 seconds 😄

src/sw-custom.js
```diff
@@ -1,6 +1,39 @@
...
+    }
+    if (data.type === 'PURGE_SCOPE' && data.hash) {
+      // Best-effort delete; log errors for debugging
+      caches.delete(TILE_CACHE_NAME_PREFIX + String(data.hash)).catch((err) => console.warn('Failed to delete cache:', err));
+      // If the current scope matches the purged hash, clear it
+      if (CURRENT_SCOPE_HASH === String(data.hash)) CURRENT_SCOPE_HASH = null;
+    }
+  });
+
+  function responseIsVectorTile(resp) {
```
luixlive: This function is never called ☠️

luixlive (changes requested):
Interesting PR. I'm a bit dusty with service workers, but this new cache looks good. Was going to approve because most of my comments are small, but I really think the service worker file can be cleaned up a little. Also I agree with Copilot, lots of `console.log`s that we probably don't need.

src/sw-custom.js
JoshuaVulcan: Resolved as per PL this morning

src/sw-custom.js
JoshuaVulcan: Fixed!

src/sw-custom.js
JoshuaVulcan: Fixed!

src/sw-custom.js
JoshuaVulcan: RIP

src/SpatialFeaturesLayer/index.js
JoshuaVulcan: It's always there, but I found a better before_layer_id for it now

.github/workflows/develop-workflow.yml
```diff
@@ -49,7 +49,7 @@ jobs:
         run: ci-helpers/makeversion.py
 
       - name: Install dependencies
-        run: yarn --ignore-scripts
+        run: yarn install
```
luixlive, line 52: Q: Why are we doing this change? Note that `yarn install` and just `yarn` is the same, but I wonder if there's a reason behind removing `--ignore-scripts` flag.

react-app-config/webpack.config.js
```diff
@@ -670,7 +670,7 @@ module.exports = function (webpackEnv) {
           // Bump up the default maximum size (2mb) that's precached,
           // to make lazy-loading failure scenarios less likely.
           // See https://github.com/cra-template/pwa/issues/13#issuecomment-722667270
-          maximumFileSizeToCacheInBytes: 5 * 1024 * 1024,
+          maximumFileSizeToCacheInBytes: 10 * 1024 * 1024,
```
luixlive, line 673: Q: Tile images can take 10 mb? Do you think caching files that large could cause a memory issue?

src/ducks/subjects.js
```diff
@@ -81,23 +81,21 @@ export const clearSubjectData = () => ({
 });
 
 export const fetchSubjectGroups = () => dispatch => axios.get(SUBJECT_GROUPS_API_URL)
-  .then(response => dispatch(fetchSubjectGroupsSuccess(response)));
+  .then(response => dispatch(fetchSubjectGroupsSuccess(response)))
+  .catch(_error => dispatch(fetchSubjectGroupsError())); // Fallback to empty array on error
```
luixlive, line 85: nit: The method `fetchSubjectGroupsError` is a bit deceiving. Makes you feel there's a special handling for errors but all we do is call the "success" action creator with an empty array. We can modernize and keep the intention clearer like this:
```javascript
export const fetchSubjectGroups = () => async (dispatch) => {
  let subjectGroups;
  try {
    subjectGroups = await axios.get(SUBJECT_GROUPS_API_URL);
  } catch (_error) {
    subjectGroups = [];
  }

  dispatch(fetchSubjectGroupsSuccess(subjectGroups));
};
```

src/ducks/subjects.js
```diff
@@ -81,23 +81,21 @@ export const clearSubjectData = () => ({
...
 
 const fetchMapSubjectsSuccess = response => ({
   type: FETCH_MAP_SUBJECTS_SUCCESS,
   payload: response.data,
 });
-/*
```
luixlive, line 90: Any good reasons to keep unused code commented? 😅

src/Map/index.js
```diff
@@ -589,9 +589,27 @@ const Map = ({ children, onMapLoad, socket }) => {
...
+          if (width && width !== 'x') {
+            dimensions.width = Number(width);
+          }
+          if (height && height !== 'x') {
+            dimensions.height = Number(height);
+          }
+        }
+
+        // Remove any remaining file extensions from the src path
+        src = src.replace(/(\.svg|\.png|\.jpg).*$/, '$1');
```
luixlive, line 609: I think there's a logical error here. The comment suggests that you are trying to remove the file extension, but this line actually doesn't do anything. It searches for the match of the regex `/(\.svg|\.png|\.jpg).*$/` and the replaces it with itself `$1`:
Image

Probably you want to change the `$1` for an empty string?
`src = src.replace(/(\.svg|\.png|\.jpg).*$/, '');`

src/Map/index.js
luixlive, line 628: nit: I feel like this `useEffect` makes me dizzy. With some help of ChatGPT I was able to reduce it to something easier to read:
```javascript
useEffect(() => {
  if (map) {
    const handleMissingImage = async (event) => {
      if (!event?.id?.includes('/static/')) {
        try {
          // Parse optional width/height suffixes, e.g. "icon-32-32.png"
          const [, basePath, w, h] = event.id.match(/^(.*?)(?:-([^-.]+)-([^-.]+))?$/) || [];
          const width = w && w !== 'x' ? Number(w) : undefined;
          const height = h && h !== 'x' ? Number(h) : undefined;

          // Strip common extensions
          const src = basePath?.replace(/\.(svg|png|jpg)$/, '') ?? id;

          await addMapImage({ height, id, src, width });
        } catch (error) {
          console.warn('Failed to add map image:', { id, error });
        }
      }
    };

    map.on('styleimagemissing', handleMissingImage);
    return () => map.off('styleimagemissing', handleMissingImage);
  }
}, [map]);
```

Not sure if it works, but if you can get to something more concise and easy to read it would be great!

src/MessageBadgeLayer/index.js
```diff
@@ -43,7 +43,7 @@ const calcMapMessages = (messages = [], subjectFeatureCollection) => {
 };
 
 const SOURCE_ID = 'MESSAGE_BADGES';
-const LAYER_ID = `${SOURCE_ID}_LAYER`;
+export const LAYER_ID = `${SOURCE_ID}_LAYER`;
```
luixlive: nit: we no longer need to export this.

src/SpatialFeaturesLayer/index.js
```diff
@@ -26,10 +32,11 @@ const DEFAULT_LINE_PAINT_COLOR = [
 const DEFAULT_POLYGON_FILL_COLOR = [
   'case',
   ['has', 'fill'], ['get', 'fill'],
-  ['has', 'color'], ['get', 'color'],
+  ['has', 'fill-color'], ['get', 'fill-color'],
   ['has', 'fill_color'], ['get', 'fill_color'],
+  ['has', 'color'], ['get', 'color'],
   ['has', 'stroke'], ['get', 'stroke'],
-  '#ff6600'
+  'rgba(255, 102, 0, 0)'
```
luixlive, line 39: Q: Some changes in this file feel a bit random to me, but probably is just because I don't have the context of why we needed to do them. But why are we changing the styling of the layers? Colors, stroke width, etc... I'm just curious.

src/SpatialFeaturesLayer/index.js
```diff
@@ -361,6 +368,15 @@ const SpatialFeaturesLayer = ({ onFeatureClick }) => {
     }
   }, [map, polygonLayerFilter]);
 
+  useEffect(() => {
+    if (!!map && !map.hasImage('marker-icon')) {
```
luixlive: nit: could be reduced to `!map?.hasImage('marker-icon')`

src/App.js
```diff
@@ -109,6 +109,26 @@ export const App = () => {
     finishDrag(e);
   }, [disallowDragAndDrop, finishDrag]);
 
+
+  // set user scope for service worker caching
+  useEffect(() => {
```
luixlive, line 114: nit: I know I already made you move this, sorry 😅 But I just thought it would be nice to clean a bit our `App` component by hiding some of these implementations behind a hook. Not urgent, not a blocker. But if you like the idea, it would be awesome to put this in a `useSetServiceWorkerScope` or something. That way we would simplify this component but also it would make logic incredibly easy to test! (Even for a Cursor agent jeje)

src/sw-custom.js
```diff
@@ -59,20 +82,65 @@ if ('function' === typeof importScripts) {
...
+        return strategy.handle({ event, request });
+      }
+    );
+    // ---------------------------------------------------------------------------
+
   } else {
     console.error('Workbox could not be loaded. No offline support');
   }
 
-  workbox.precaching.precacheAndRoute(self.__WB_MANIFEST);
```
luixlive: Love it! This file ended up being simpler, easier to navigate and smaller. I just notice there's a bunch of empty lines at the end haha.

luixlive (approved):
Good to go 🚀 

Everything looks fine. I left a bunch of questions (some out of curiosity, some just food for thought to see if we can improve the code) and a few nits, but nothing to block it.

react-app-config/webpack.config.js
JoshuaVulcan: 🤷 I was just trying to ensure we catch them all. It's unlikely in a real environment.

src/sw-custom.js
JoshuaVulcan: I'll clean those up!

src/SpatialFeaturesLayer/index.js
JoshuaVulcan: Previously, we had no default value for this, meaning it would be transparent. So I matched that functionality for default features here, leaving them transparent if no value comes from the server.

src/Map/index.js
JoshuaVulcan: Sorry, my comment made things worse instead of clearer. I mean this removes the trailing part of the image path used to specify dimension bounds as a function of image ID internally, like /path/to/file.svg-10-x etc, and forwards them along so addImage respects the dimensions requested.

src/ducks/subjects.js
JoshuaVulcan: You might have been reading the diff wrong, this is me deleting this commented-out code

PR #1397, opened 2025-11-03
hotfix:removing case normalization for disabled option eval

JoshuaVulcan:
### What does this PR do?
- See title

### How does it look
- N/A

### Relevant link(s)
* https://allenai.atlassian.net/browse/ERA-12182

### DIFF
--- a/src/SchemaFields/index.js
+++ b/src/SchemaFields/index.js
@@ -271,7 +271,7 @@ export const SelectWidget = ({
 
   const getOptionValue = (optionValue) => isPlainObject(optionValue) ? optionValue.value : optionValue;
 
-  const isOptionDisabled = (option) => (schema?.inactive_enum ?? []).includes(option.value.toLowerCase());
+  const isOptionDisabled = (option) => (schema?.inactive_enum ?? []).includes(option.value);
 
   const selected = enumOptions.find((item) => value
     ? item.value === (isPlainObject(value) ? value.value : value)
--- a/src/SchemaFields/index.js
+++ b/src/SchemaFields/index.js
@@ -1,4 +1,4 @@
-import React, { useCallback, useMemo, useRef, useState } from 'react';
+import React, { useCallback, useRef, useState } from 'react';
 import Button from 'react-bootstrap/Button';
 import { canExpand, getInputProps, getTemplate, getUiOptions } from '@rjsf/utils';
 import Form from 'react-bootstrap/Form';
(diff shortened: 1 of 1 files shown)


src/SchemaFields/index.js
```diff
@@ -276,7 +271,7 @@ export const SelectWidget = ({
 
   const getOptionValue = (optionValue) => isPlainObject(optionValue) ? optionValue.value : optionValue;
 
-  const isOptionDisabled = (option) => schemaInactiveEnumLowercase.includes(option.value.toLowerCase());
+  const isOptionDisabled = (option) => (schema?.inactive_enum ?? []).includes(option.value.toLowerCase());
```
luixlive: Two comments on this line:

nit: It seems like `schema` is always defined, I guess it's injected by `rjsf`, so we don't need the optional chain `schema.inactive_enum`.

Bug: I see that we are still doing `option.value.toLowerCase()`, so if there's an uppercase letter in any of the values in the `schema.inactive_enum`, we can know for sure that this logic will fail to match the option, resulting in not disabling those inactive options. All this has a smell, we should ask ourselves why we had this `toLowerCase` logic in the first place. There must be a reason,. Removing the `toLowerCase` may sound like a quick fix to the reported bug, but maybe we are injecting other issues.

luixlive (changes requested):

src/SchemaFields/index.js
JoshuaVulcan: `filteredEnumOptions` has an existing guard clause for the same property, I'm following that convention. All other feedback addressed!

luixlive (approved):
> `filteredEnumOptions` has an existing guard clause for the same property, I'm following that convention. All other feedback addressed!

Oh I think there's a confusion. Your optional chaining is protecting against an `undefined` `schema`. `filteredEnumOptions` is not doing that:
```javascript
const filteredEnumOptions = enumOptions.filter((option) => {
  const itemDisabled = schema.inactive_enum && schema.inactive_enum.includes(option.value);

  return !itemDisabled || (!!itemDisabled && !!originalValues.includes(option.value));
});
```

The guard clause there is for `schema.inactive_enum`. But you can see that we are not adding optional chaining like `schema?.inactive_enum` which means we trust on `schema` always being defined. Anyway this is just a nit.

About removing the `toLowerCase` I still have the same concerns as before. I think we should dig a bit and understand why we decided to make the values lower case in the first place before just removing that logic 🤔 Maybe it was unnecessary, I couldn't tell, so I will approve it. But I think it's a good idea that you and @AlanCalvillo do some test around this to make sure everything is fine 👍

PR #1398, opened 2025-11-03
ERA-12069: ER Web Customization via Tenant Feature Flags

luixlive:
### What does this PR do?
Add support for 4 new system configuration flags: `analyzers_enabled`, `events_enabled`, `spatial_features_enabled`, and `subjects_enabled`.

To make this and future permission-related work easier, a new file was created with a set of hooks to easily know if the user has specific permissions and if the system configuration allows the feature to be shown: `src/hooks/userPermissions/index.js`.

### Relevant link(s)
* [ERA-12069](https://allenai.atlassian.net/browse/ERA-12069)
* [Env]()

1- Read the new file (but not new reducer) `src/ducks/system-config/index.js` with the `systemConfig` reducer that stores the system config flags. It's very easy to understand and fully tested.
2- Then read the new `src/hooks/userPermissions/index.js` with utilities to easily load booleans that tell if the user has a specific permission:
```javascript
// The hooks have no dependencies and they export all permissions for the feature.
const { hasEventsCreatePermission } = useEventsPermissions();
const { hasObservationsExportPermission } = useObservationsPermissions();
const { hasPatrolsUpdatePermission } = usePatrolsPermissions();
const { hasMessagesReadPermission } = useMessagesPermissions();
```
3- Now you have enough context to understand the changes in the rest of the files where we refactor old code to match the new pattern, and add new conditions around the app to show / hide features related to analyzers, events, spatial features, and subjects based on the new system config flags.

**Notes:** Probably unnecessary to say, but we should thoroughly test these new flags in the feature environment (@JoshuaVulcan will help creating those flags in the tenant for this feature env, we already have a branch and [PR](https://github.com/PADAS/das/pull/3636))

[ERA-12069]: https://allenai.atlassian.net/browse/ERA-12069?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/AddItemButton/AddItemModal/index.js
```diff
@@ -20,33 +19,39 @@ export const ADD_TAB_KEYS = { ADD_REPORT: 'reports', ADD_PATROL: 'patrols' };
...
 
-  const patrolTypes = useSelector((state) => state.data.patrolTypes);
+  const { hideAddPatrolTab, hideAddReportTab } = useContext(AddItemContext);
 
   const storedActiveTabKey = getStoredTab() || ADD_TAB_KEYS.ADD_REPORT;
 
   const [activeTabKey, setActiveTabKey] = useState(storedActiveTabKey);
 
-  const patrolsEnabled = !!patrolFlagEnabled && !!hasPatrolWritePermissions && !!patrolTypes.length;
+  const canAddEvents = hasEventsCreatePermission && eventTypes.length > 0 && !hideAddReportTab;
```
luixlive, line 34: Hide the "Add Events" tab from our `AddItemModal` if the `events_enabled` system config is disabled:
Image

src/AddItemButton/index.js
```diff
@@ -53,17 +57,18 @@ const AddItemButton = ({
...
-    formProps,
-    hideAddPatrolTab,
-    hideAddReportTab,
-    onAddPatrol,
-    onAddReport,
-    patrolData,
-    reportData,
-  };
-  return <AddItemContext.Provider value={addItemContextValue}>
+  return (hasEventsCreatePermission || hasPatrolsCreatePermission) ? <AddItemContext.Provider
```
luixlive, line 60: Don't render the `AddItemButton` at all if user doesn't have permissions (or system config flags are disabled) for both events and patrols:
Image

src/constants/index.js
```diff
@@ -131,16 +131,19 @@ export const DEFAULT_SHOW_NAMES_IN_MAP_CONFIG = {
 };
 
 export const SYSTEM_CONFIG_FLAGS = {
-  PATROL_MANAGEMENT: 'patrol_enabled',
```
luixlive, line 135: Added new flags and sorted all of them alphabetically:
```javascript
export const SYSTEM_CONFIG_FLAGS = {
  ALERTS: 'alerts_enabled',
  ANALYZERS: 'analyzers_enabled',                                       // new
  DAILY_REPORT: 'daily_report_enabled',
  DEFAULT_EVENT_FILTER_FROM_DAYS: 'default_event_filter_from_days',
  DEFAULT_PATROL_FILTER_FROM_DAYS: 'default_patrol_filter_from_days',
  EULA: 'eula_enabled',
  EVENTS: 'events_enabled',                                             // new
  GEOPERMISSIONS: 'geopermissions_enabled',
  KML_EXPORT: 'export_kml_enabled',
  PATROL_MANAGEMENT: 'patrol_enabled',
  SPATIAL_FEATURES: 'spatial_features_enabled',                         // new
  SUBJECTS: 'subjects_enabled',                                         // new
  TABLEAU: 'tableau_enabled',
};
```

src/DetailViewComponents/ActivitySection/index.js
```diff
@@ -84,16 +87,18 @@ const ActivitySection = ({
     />,
   })), [attachmentsToAdd, onDeleteAttachment]);
 
-  const containedReportsRendered = useMemo(() => containedReports.map((containedReport) => ({
```
luixlive, line 87: Hide patrol contained events if the new `events_enabled` system config is disabled:
Image

src/ducks/system-config/index.js
```diff
@@ -0,0 +1,87 @@
+import { DEFAULT_SHOW_TRACK_DAYS, SYSTEM_CONFIG_FLAGS } from '../../constants';
```
luixlive, line 1: New independent reducer to keep the system config flags logic organized and testable.

src/GlobalMenuDrawer/index.js
```diff
@@ -60,28 +54,27 @@ export const USERS_GUIDE_SITE_URL = 'https://support.earthranger.com/en_US/earth
...
-  const alertsEnabled = useSelector((state) => state.view.systemConfig.alerts_enabled);
+  const alertsEnabled = useSelector((state) => state.view.systemConfig[SYSTEM_CONFIG_FLAGS.ALERTS]);
+  const dailyReportEnabled = useSelector((state) => state.view.systemConfig[SYSTEM_CONFIG_FLAGS.DAILY_REPORT]);
   const drawer = useSelector((state) => state.view.drawer);
   const eventTypes = useSelector((state) => state.data.eventTypes);
+  const kmlExportEnabled = useSelector((state) => state.view.systemConfig[SYSTEM_CONFIG_FLAGS.KML_EXPORT]);
   const selectedUserProfile = useSelector((state) => state.data.selectedUserProfile);
   const serverData = useSelector((state) => state.data.systemStatus.server);
-  const tableauEnabled = useSelector((state) => state.view.systemConfig.tableau_enabled);
+  const subjectsEnabled = useSelector((state) => state.view.systemConfig[SYSTEM_CONFIG_FLAGS.SUBJECTS]);
```
luixlive, line 69: Add checks to the new system configs `events_enabled` and `subjects_enabled` to conditionally render export buttons in our `GlobalMenuDrawer`:
Image

src/hooks/usePermissions/index.js
```diff
@@ -0,0 +1,94 @@
+import { useMemo } from 'react';
```
luixlive, line 1: New hooks file that provides simple hooks to get specific permissions from the current user or profile and that overrides the permission values with the system config flags:
```javascript
// Destructure the permission you need: create, delete, export, read, update.
const { hasEventsCreatePermission, hasEventsReadPermission } = useEventsPermissions();
```

src/Map/index.js
```diff
@@ -657,29 +660,30 @@ const Map = ({ children, onMapLoad, socket }) => {
 
       <ClustersLayer onShowClusterSelectPopup={onShowClusterSelectPopup} />
 
-      <EventsLayer
+      {hasEventsReadPermission && <EventsLayer
```
luixlive, line 690: Since these system configs are client only features, we can't trust the DAS backend or tile server to not return the events, subjects, features, and analyzers data. So to make sure they don't render in the map, we conditionally render the layers.

src/SideBar/MapLayersTab/index.js
```diff
@@ -27,7 +29,23 @@ const MapLayersTab = () => {
   const dispatch = useDispatch();
   const { t } = useTranslation('components', { keyPrefix: 'sideBar.mapLayersTab' });
 
+  const analyzersEnabled = useSelector((state) => state.view.systemConfig[SYSTEM_CONFIG_FLAGS.ANALYZERS]);
```
luixlive, line 32: `MapLayersTab` now checks if the system config flag of each of its sub-tabs is enabled in order to render them. So now the tabs `Subjects`, `Features`, `Analyzers`, and `Events` render conditionally:
Image

src/SideBar/index.js
```diff
@@ -48,83 +48,103 @@ const SideBar = () => {
...
+  const socket = useContext(SocketContext);
 
   const patrolsFeed = useFetchPatrolsFeed();
   const reportsFeed = useReportsFeed();
 
-  const socket = useContext(SocketContext);
-
-  const sideBarRef = useRef();
-
+  const analyzersEnabled = useSelector((state) => state.view.systemConfig[SYSTEM_CONFIG_FLAGS.ANALYZERS]);
```
luixlive, line 56: `SideBar` now conditionally renders three of its items: `Events`, `Patrols`, and `Map Layers`. The first two just depend on the new permissions hook that already checks the system config flags. The `Map Layers` link has a more complex condition. Basically if none of the map features is enabled, we hide it:
```javascript
const showLayersTab = analyzersEnabled || spatialFeaturesEnabled || subjectsEnabled || hasEventsReadPermission;
```

Image

src/hooks/usePermissions/index.js
```diff
@@ -0,0 +1,94 @@
...
+  const isFeatureEnabled = useMemo(() => {
+    switch (permissionKey) {
+    case PERMISSION_KEYS.EVENTS:
+      return eventsEnabled;
+
+    case PERMISSION_KEYS.PATROLS:
+      return patrolManagementEnabled;
+
+    default:
+      return true;
```
JoshuaVulcan: Seems like the default should be `false` for feature flags, right? Unless otherwise known...

JoshuaVulcan (approved):
In general, I approve, although I think `usePermissions` could be improved and made a bit more generic to avoid 1. extraneous selections (that event state and patrol state selector instantiating for each instance of `usePermissions`, even when not about event or patrol permissions, def has a smell) and 2. hard-coding lists of permission types. We can chat on Monday, but nothing blocking approval nor merge.

src/hooks/usePermissions/index.js
luixlive: Hmm if we set the default as `false` then any feature that is no Patrols or Events would be disabled all the time. For example: `useObservationsPermissions()` would return `false` for all its flags since it doesn't have a feature flag.

I'm going to work on a small change that will remove this `switch` statement following your first suggestion to not read event and patrol state every time we use this hook. Maybe after those changes the code is easier to understand 👍

luixlive:
@JoshuaVulcan just sent a commit that removes the selectors that read patrols and events system config files in the general `usePermissions` hook, instead we read them locally in the `useEventsPermissions` and `usePatrolsPermissions` hooks: https://github.com/PADAS/das-web-react/pull/1398/commits/0c12bda829d7ca890c035be9069932e8023d2d19

About the hardcoding, not sure what you mean, but I guess the fact that we list the 5 type of supported permissions like this:
```javascript
return useMemo(() => ({
  hasCreatePermission: permissionSet.has(PERMISSIONS.CREATE),
  hasDeletePermission: permissionSet.has(PERMISSIONS.DELETE),
  hasExportPermission: permissionSet.has(PERMISSIONS.EXPORT),
  hasReadPermission: permissionSet.has(PERMISSIONS.READ),
  hasUpdatePermission: permissionSet.has(PERMISSIONS.UPDATE),
}), [permissionSet]);
```

Is that right? If that's the case, well I'm not sure, do we support any other type of permission? We only have these 5 in our `PERMISSIONS` constant (that has been there for years) and I don't think we are using any other kind of permission in the code. I feel this is safe, and I like the fact that you can use the hooks from anywhere in the code and just get the permission you need instead of having to specifically request for it:
```javascript
// With the implementation in this PR.
const { hasMessagesReadPermission } = useMessagesPermissions();


// If we don't want to hardcode the 5 types of permissions. We need to import PERMISSIONS and pass a parameter.
import { PERMISSIONS } from '@/constants';

const hasMessagesReadPermission = useMessagesPermissions(PERMISSIONS.READ);
```

luixlive:
I'm merging this PR since it has an approval and we are about to close the sprint. But if there's any remaining feedback I'll be glad to send another commit in a separate PR @JoshuaVulcan

PR #1399, opened 2025-11-05
ERA-12138: Fix extra click navigating between two popup-capable items on the map

AlanCalvillo:
### What does this PR do?
- Adds checks if the feature under the click is a timepoint or features, if so, always make the shouldHidePopup flag false.

### How does it look
between timepoints

https://github.com/user-attachments/assets/61859226-7e64-45bd-860b-0556adad4c3b

same issue happened on Spatial Features

https://github.com/user-attachments/assets/f999c835-c797-49da-a0a8-973409dd2a7e

### Relevant link(s)
* Tracking tickets: [ERA-12138](https://allenai.atlassian.net/browse/ERA-12138)

[ERA-12138]: https://allenai.atlassian.net/browse/ERA-12138?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Map/index.js
+++ b/src/Map/index.js
@@ -431,19 +431,11 @@ const Map = ({ children, onMapLoad, socket }) => {
   }, [fetchMapData]);
 
   // Helper function to check if a feature should keep the popup open
-  const isPopupReplacementFeature = useCallback((feature) => {
-    const { layer } = feature;
-
-    if (layer.id.includes(LAYER_IDS.TRACK_TIMEPOINTS)) {
-      return true;
-    }
-
-    if ([SYMBOLS_LAYER_ID, LINES_LAYER_ID, POLYGONS_LAYER_ID, POLYGONS_OUTLINE_LAYER_ID].includes(layer.id)) {
-      return true;
-    }
-
-    return false;
-  }, []);
+  const doesFeatureOpenPopup = useCallback(
+    (feature) => feature.layer.id.includes(LAYER_IDS.TRACK_TIMEPOINTS)
+       || [SYMBOLS_LAYER_ID, LINES_LAYER_ID, POLYGONS_LAYER_ID, POLYGONS_OUTLINE_LAYER_ID].includes(feature.layer.id),
+    []
+  );
 
   const onMapClick = useMemo(() => withLocationPickerState((event) => {
     event.preventDefault();
@@ -470,8 +462,7 @@ const Map = ({ children, onMapLoad, socket }) => {
     }
 
     // Determine if we should hide the existing popup
-    const hasPopupReplacementFeature = featuresAtPoint.some(isPopupReplacementFeature);
-    const shouldHidePopup = !hasClusters && !hasPopupReplacementFeature;
+    const shouldHidePopup = !hasClusters && !featuresAtPoint.some(doesFeatureOpenPopup);
 
     // Handle popup visibility
     if (popup) {
@@ -491,7 +482,7 @@ const Map = ({ children, onMapLoad, socket }) => {
     handleMultiFeaturesAtSameLocationClick,
     hidePopup,
     hideUnpinnedTrackLayers,
-    isPopupReplacementFeature,
+    doesFeatureOpenPopup,
     map,
     popup,
     withLocationPickerState,
--- a/src/Map/index.js
+++ b/src/Map/index.js
@@ -67,12 +67,11 @@ import MessageBadgeLayer from '../MessageBadgeLayer';
 import MapImagesLayer from '../MapImagesLayer';
 import SleepDetector from '../SleepDetector';
 import ClustersLayer from '../ClustersLayer';
-import {
+import SpatialFeaturesLayer, {
   SYMBOLS_LAYER_ID,
   LINES_LAYER_ID,
   POLYGONS_LAYER_ID,
   POLYGONS_OUTLINE_LAYER_ID,
-  SpatialFeaturesLayer,
 } from '../SpatialFeaturesLayer';
 
 
@@ -457,7 +456,7 @@ const Map = ({ children, onMapLoad, socket }) => {
 
     // Check if clicking on a timepoint (track layer point) - don't hide popup so it can be replaced
     const timepointAtClick = map.queryRenderedFeatures(event.point)
-      .find((item) => item.layer.id.includes(LAYER_IDS.TRACK_TIMEPOINTS_SYMBOLS));
+      .find((item) => item.layer.id.includes(LAYER_IDS.TRACK_TIMEPOINTS));
     if (timepointAtClick) {
       shouldHidePopup = false;
     }
--- a/src/Map/index.js
+++ b/src/Map/index.js
@@ -67,7 +67,14 @@ import MessageBadgeLayer from '../MessageBadgeLayer';
 import MapImagesLayer from '../MapImagesLayer';
 import SleepDetector from '../SleepDetector';
 import ClustersLayer from '../ClustersLayer';
-import SpatialFeaturesLayer from '../SpatialFeaturesLayer';
+import {
+  SYMBOLS_LAYER_ID,
+  LINES_LAYER_ID,
+  POLYGONS_LAYER_ID,
+  POLYGONS_OUTLINE_LAYER_ID,
+  SpatialFeaturesLayer,
+} from '../SpatialFeaturesLayer';
+
 
 import AddItemButton from '../AddItemButton';
 import MapRulerControl from '../MapRulerControl';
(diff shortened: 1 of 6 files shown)


src/Map/index.js
```diff
@@ -67,7 +67,14 @@ import MessageBadgeLayer from '../MessageBadgeLayer';
 import MapImagesLayer from '../MapImagesLayer';
 import SleepDetector from '../SleepDetector';
 import ClustersLayer from '../ClustersLayer';
-import SpatialFeaturesLayer from '../SpatialFeaturesLayer';
+import {
+  SYMBOLS_LAYER_ID,
+  LINES_LAYER_ID,
+  POLYGONS_LAYER_ID,
+  POLYGONS_OUTLINE_LAYER_ID,
+  SpatialFeaturesLayer,
```
luixlive: There is an issue here. `SpatialFeaturesLayer` is the default export of this file, not a named export 🤔 The import should look like:
```javascript
import SpatialFeaturesLayer, {
  SYMBOLS_LAYER_ID,
  LINES_LAYER_ID,
  POLYGONS_LAYER_ID,
  POLYGONS_OUTLINE_LAYER_ID,
} from '../SpatialFeaturesLayer';
```

src/constants/index.js
```diff
@@ -99,7 +99,7 @@ export const LAYER_IDS = {
   STATIC_SENSOR: 'static_sensor',
   SUBJECT_SYMBOLS: 'subject-symbol-layer',
   TOPMOST_STYLE_LAYER: 'feature-separation-layer',
-  TRACK_TIMEPOINTS_SYMBOLS: 'track-layer-timepoints',
+  TRACK_TIMEPOINTS_SYMBOLS: 'track-layer-points',
```
luixlive: Question (and a nit): Any reason for this change?

I've been wanting to formalize the pattern of the source and layer names, since right now they feel messy and random haha. It would be nice to follow an approach like in `COORDINATE_REFERENCE_SYSTEM_BBOX: 'coordinate-reference-system-bbox-layer'`.

So this would become: `TRACK_TIMEPOINTS: 'track-timepoints-layer'`.

Or if we really need to specify the layer type (symbols): `TRACK_TIMEPOINTS_SYMBOLS: 'track-timepoints-symbols-layer'`.

luixlive (changes requested):
Nice to see you dropping PRs!

Submitting a change request because I think I'm seeing a possible bug (incorrect import).

On a separate note, following our boy scout rule (improving the code we touch) this is a good opportunity to redo a bit that `onMapClick` method that is kind of ugly. Maybe with some help of an agent you can improve it a bit, maybe even its performance since I see many calls to `map.queryRenderedFeatures` that could be integrated into a single query + iteration!

Also I see a good amount of tests of the new code, but they are pretty verbose on the comments. Some are not necessary at all, like
```javascript
// hidePopup should NOT have been called
expect(hidePopupMock).not.toHaveBeenCalled();
```

I can infer that myself by reading the `expect` statement 😄

src/constants/index.js
AlanCalvillo: cursor firstly suggested this:

```
const timepointAtClick = map.queryRenderedFeatures(event.point)
      .find((item) => item.layer.id.includes('track-layer-points-'));
    if (timepointAtClick) {
      shouldHidePopup = false;
    }
```
but then i realized we had these constants, i noted it looked pretty similar to cursor's suggestion, but it wasn't being used anywhere in the code. so i decided to match the suggestion. 

i think it could potentially be used here https://github.com/PADAS/das-web-react/blob/552f702c84ba1d221250449c7e9285dc2f7c6355/src/TracksLayer/index.js#L23
and here
https://github.com/PADAS/das-web-react/blob/552f702c84ba1d221250449c7e9285dc2f7c6355/src/PatrolTrackLayer/index.js#L17

and, yes i can update the name.

luixlive: Hmm, this comment brings more questions now. If this layer id is not being used anywhere in the code, then why are we expecting for a layer with that id to exist at all? We are querying features from our map that belong to a layer id that we are not defining 🤔 

And yes, those two code lines you shared are using magic strings instead of a constant, definitely a possible improvement. But I also searched in the code for `track-layer-points-` and I don't see that string anywhere else, so again it seems like we are trying to query by layer ids that don't exist at all. This sounds like useless code, maybe we should take this a bit further and debug if those methods ever find anything at all (I guess not since we are not defining those layers).

Anyway, fixing those other two places are out of the scope, but coming back to Cursor's solution here, it seems like it's querying features from a layer that doesn't exist. Did you test this manually and see if this query ever returned something?
```javascript
const timepointAtClick = map.queryRenderedFeatures(event.point)
      .find((item) => item.layer.id.includes(LAYER_IDS.TRACK_TIMEPOINTS));
```

luixlive:
With your latest changes this PR seems solid. I would still try asking cursor to improve the readability and performance of the `onMapClick` method. I feel it can do something cool. But before approving, please take a look at my latest response in our discussion above. There's a smell here with those layer ids that we are querying but don't seem to be defined anywhere.

I was wondering if maybe that layer was moved to the tiles server. But I just took a look at https://github.com/PADAS/das-web-react/pull/1319 and I don't see any changes (removals or additions) related to them.

src/constants/index.js
AlanCalvillo: when i click a timepoint, the value of `timepointAtClick` is something like:
```
{
  id: "track-layer-points-7f74aef5-a431-4112-aba7-984ee002778f",
  type: "symbol",
  source: "track-source-7f74aef5-a431-4112-aba7-984ee002778f-points",
  layout: {
    "icon-allow-overlap": false,
    "icon-anchor": "bottom",
    "icon-size": 0.15,
    "icon-rotate": -77.38522670467313,
    "icon-image": "track_arrow",
    "icon-pitch-alignment": "map",
    "icon-rotation-alignment": "map",
  },
  paint: {
    "icon-opacity": 1,
  },
}
```
so seems like it does find those layers. could they have been assigned here? https://github.com/PADAS/das-web-react/blob/552f702c84ba1d221250449c7e9285dc2f7c6355/src/TracksLayer/track.js#L79
(ain't used to mapbox idiosyncrasies)

src/Map/index.js
```diff
@@ -430,61 +430,72 @@ const Map = ({ children, onMapLoad, socket }) => {
     fetchMapData();
   }, [fetchMapData]);
 
+  // Helper function to check if a feature should keep the popup open
+  const isPopupReplacementFeature = useCallback((feature) => {
```
luixlive: nit: Two suggestions on this method.
1- The word "replacement" makes sense in the context of this PR, but we may reuse the function for other things so we could use a more generalized name like `doesFeatureOpenPopup`.
2- We can simplify it a lot. It only returns a boolean so we can make it a simple OR operation.

```javascript
const doesFeatureOpenPopup = useCallback(
  (feature) => feature.layer.id.includes(LAYER_IDS.TRACK_TIMEPOINTS)
    || [SYMBOLS_LAYER_ID, LINES_LAYER_ID, POLYGONS_LAYER_ID, POLYGONS_OUTLINE_LAYER_ID].includes(feature.layer.id),
  []
);
```

src/Map/index.js
```diff
@@ -430,61 +430,72 @@ const Map = ({ children, onMapLoad, socket }) => {
...
+
+    return false;
+  }, []);
+
   const onMapClick = useMemo(() => withLocationPickerState((event) => {
     event.preventDefault();
     event.originalEvent.stopPropagation();
 
+    // Query features once for performance
+    const featuresAtPoint = map.queryRenderedFeatures(event.point);
```
luixlive, line 445: Nice!

src/Map/index.js
```diff
@@ -430,61 +430,72 @@ const Map = ({ children, onMapLoad, socket }) => {
...
-        LINES_LAYER_ID,
-        POLYGONS_LAYER_ID,
-        POLYGONS_OUTLINE_LAYER_ID
-      ].includes(item.layer.id));
-    if (spatialFeatureAtClick) {
-      shouldHidePopup = false;
-    }
+    // Determine if we should hide the existing popup
+    const hasPopupReplacementFeature = featuresAtPoint.some(isPopupReplacementFeature);
+    const shouldHidePopup = !hasClusters && !hasPopupReplacementFeature;
```
luixlive: Hard to visualize the resulting refactor of `onMapClick` here with all the diff. But I just checked it in a separate file and it looks way better! Just an extremely nit, we are defining a variable that's only being used once in the next line and it's not even hard to understand or read. We could simply: 
```javascript
const shouldHidePopup = !hasClusters && !featuresAtPoint.some(isPopupReplacementFeature);
```

luixlive (approved):
Nice! Ready to ship. Left a couple minor nits if you find a couple minutes to be more picky with the code.

PR #1400, opened 2025-11-08
ERA-12164: User profiles list needs a scroll container

AlanCalvillo:
### What does this PR do?
- Adds a container to the profiles list and adds auto-scroll if it overflows
- Add overflow-y to the whole menu, so Cooke Settings and log-out menu items are reachable on small viewports.
- Adds tests to the component.

### How does it look
https://github.com/user-attachments/assets/695ab610-8175-4e7c-b32c-55dc3371dcba

### Relevant link(s)
* Tracking tickets: [ERA-12164](https://allenai.atlassian.net/browse/ERA-12164)

- fruit of vibe-coding, be as fussy as you can

[ERA-12164]: https://allenai.atlassian.net/browse/ERA-12164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/public/locales/en-US/top-bar.json
+++ b/public/locales/en-US/top-bar.json
@@ -61,10 +61,10 @@
     "WARNING": "Warning"
   },
   "userMenu": {
-    "userAndProfilesHeader": "User and Profiles",
     "cookieSettingsItem": "Cookie Settings",
     "logoutItem": "Log out",
     "toggleLabel": "Open User Settings",
-    "toggleTitle": "User Settings"
+    "toggleTitle": "User Settings",
+    "userAndProfilesHeader": "User and Profiles"
   }
 }
--- a/public/locales/en-US/top-bar.json
+++ b/public/locales/en-US/top-bar.json
@@ -61,6 +61,7 @@
     "WARNING": "Warning"
   },
   "userMenu": {
+    "userAndProfilesHeader": "User and Profiles",
     "cookieSettingsItem": "Cookie Settings",
     "logoutItem": "Log out",
     "toggleLabel": "Open User Settings",
--- a/src/UserMenu/index.js
+++ b/src/UserMenu/index.js
@@ -22,7 +22,7 @@ const UserMenu = ({
 
   const cookieSettingsRef = useRef();
 
-  const displayUser = selectedUserProfile.username ? selectedUserProfile : user;
+  const displayUser = selectedUserProfile?.username ? selectedUserProfile : user;
 
   const onLogOutItemClick = () => {
     onLogOutClick();
@@ -47,14 +47,16 @@ const UserMenu = ({
 
       <Dropdown.Menu>
         {!!userProfiles.length && <>
-          {[user, ...userProfiles].map((profile, index) => <Dropdown.Item
-            active={profile.username === displayUser.username ? 'active' : null}
-            aria-label={profile.username}
-            key={`${profile.id}-${index}`}
-            onClick={() => onProfileClick(profile)}
-          >
-            {profile.username}
-          </Dropdown.Item>)}
+          <div className={styles.profilesList}>
+            {[user, ...userProfiles].map((profile, index) => <Dropdown.Item
+              active={profile.username === displayUser.username ? 'active' : null}
+              aria-label={profile.username}
+              key={`${profile.id}-${index}`}
+              onClick={() => onProfileClick(profile)}
+            >
+              {profile.username}
+            </Dropdown.Item>)}
+          </div>
 
           <Dropdown.Divider />
         </>}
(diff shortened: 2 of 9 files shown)


src/UserMenu/index.js
```diff
@@ -47,14 +47,16 @@ const UserMenu = ({
...
-            active={profile.username === displayUser.username ? 'active' : null}
-            aria-label={profile.username}
-            key={`${profile.id}-${index}`}
-            onClick={() => onProfileClick(profile)}
-          >
-            {profile.username}
-          </Dropdown.Item>)}
+          <div className={styles.profilesList}>
+            {[user, ...userProfiles].map((profile, index) => <Dropdown.Item
+              active={profile.username === displayUser.username ? 'active' : null}
```
luixlive: I know this code isn't new, but it can be improved. `active` is a boolean prop, and here we are either setting a string or null. Both are incorrect. I think the right approach is `active={profile.username === displayUser.username`.

Here the docs: https://react-bootstrap.netlify.app/docs/components/dropdowns/#dropdownitem

src/UserMenu/index.js
```diff
@@ -47,14 +47,16 @@ const UserMenu = ({
...
+              aria-label={profile.username}
```
luixlive: `aria-label` is the same than the item text. This is redundant and even confusing for screen readers, so it would be better to remove the `aria-label.`

src/UserMenu/index.js
```diff
@@ -47,14 +47,16 @@ const UserMenu = ({
...
+              key={`${profile.id}-${index}`}
```
luixlive: Aren't profile ids unique? It's an antipattern in React to use `index` as ids in the `key` prop.

src/UserMenu/index.test.js
```diff
@@ -0,0 +1,215 @@
...
+    test('displays the current user username', () => {
+      renderUserMenu();
+      expect(screen.getByText('john.doe')).toBeInTheDocument();
+    });
+
+    test('displays selected user profile when provided', () => {
+      const selectedProfile = { id: 'profile-1', username: 'jane.smith' };
+      renderUserMenu({ selectedUserProfile: selectedProfile });
+
+      const toggleButton = screen.getByTestId('user-menu-toggle-btn');
```
luixlive: Nice testing coverage! However, I see that we use stuff like query selectors an test ids. They work, but we lose the opportunity to make sure our code is accessible as part of our testing strategy. It would be better to use:
```javascript
screen.getByRole('button', { name: 'Text or label of the button' });
```

This way we make sure that our button has the right role and a good associated name. This applies to the rest of tests in this file. Here the hierarchical order in which it's preferred to query: https://testing-library.com/docs/queries/about/#priority

luixlive (approved):
Nice! I provided a couple items of feedback over the code. And I have an extra suggestion to take this to the next level: adding a menu header (and possibly a menu divider). Here the docs, extremely easy to do: https://react-bootstrap.netlify.app/docs/components/dropdowns/#menu-headers

That way over menu would render like:
Image

public/locales/en-US/top-bar.json
```diff
@@ -61,6 +61,7 @@
     "WARNING": "Warning"
   },
   "userMenu": {
+    "userAndProfilesHeader": "User and Profiles",
```
luixlive: Extremely nit: we sort i18n keys alphabetically so files are easier to navigate visually.

I know we fail to achieve it. Some files aren't correctly sorted, some files have missing translations... This is a great opportunity to create a technical debt ticket and put agents to work! We can prompt an agent to sort the files and while doing so, tell us which translations are missing in some languages.

src/UserMenu/index.js
```diff
@@ -47,11 +47,11 @@ const UserMenu = ({
...
         {!!userProfiles.length && <>
+          <Dropdown.Header>{t('userAndProfilesHeader')}</Dropdown.Header>
           <div className={styles.profilesList}>
-            {[user, ...userProfiles].map((profile, index) => <Dropdown.Item
-              active={profile.username === displayUser.username ? 'active' : null}
-              aria-label={profile.username}
-              key={`${profile.id}-${index}`}
+            {[user, ...userProfiles].map((profile) => <Dropdown.Item
+              active={profile.username === displayUser.username}
+              key={`${profile.id}`}
```
luixlive: Another nit: this can be simplified to `key={profile.id}`. No need for the template string anymore.

luixlive (approved):
Another couple nits. Extremely picky. Feel free to merge this PR whenever you are ready 🎉

public/locales/en-US/top-bar.json
AlanCalvillo: following up here:
https://allenai.atlassian.net/browse/ERA-12246

PR #1415, opened 2025-12-05
fix:update layer filter when polygon filter changes

JoshuaVulcan:
### What does this PR do?
- Adds filter update for polygon outline layer on polygon hide/show toggle

### How does it look
- N/A. It's a one-liner.


luixlive (approved):
Easy.

Should we uncomment line 389? Or is there a reason to keep that comment there?
```Javascript
// map?.getLayer?.(POLYGONS_LABELS_LAYER_ID),
```

PR #1416, opened 2025-12-08
ERA-12342: das-web-react: Add transformSchemaToFormElements with support for conditional sections

luixlive:
### What does this PR do?
Introduce a new transformation method that outputs a unified `formElements` structure with full support for conditional sections.

### Relevant link(s)
* [ERA-12342](https://allenai.atlassian.net/browse/ERA-12342)

Note that this PR only adds the new utility `transformSchemaToFormElements` but it doesn't implement it yet. I just created the method, its utilities, and thorough unit tests. All the changes in this PR are additions.

For that reason, I'd recommend to not review in Github, but pull the branch locally and open `src/utils/v2-event-schemas/transformSchemaToFormElements/index` and from there start the reviewing the transformation flow. Basically:
1- `transformSchemaToFormElements`: the starting point that receives the schema, initializes the `formElements` object and adds the active sections one by one.
2- `transformSection`: adds a section to the `formElements` transforming its schema data. Then it adds the active section children one by one.
3- `transformHeader`: adds a header to the `formElements` transforming its schema data.
4- `transformField`: adds a field to the `formElements` transforming its schema data depending on the field type.

`transformSchemaToFormElements` is the evolution of `makeFieldsFromSchema`. It's more resilient (sets defaults when form element properties are missing to avoid failures), throws insightful errors, modularizes the transformation flow, and has total coverage of unit testing, including corner cases.

[ERA-12342]: https://allenai.atlassian.net/browse/ERA-12342?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JoshuaVulcan (approved):
Looks good to me! I would personally organize some of these error classes into a single file, but I get the consistency/layout established here.

PR #1419, opened 2025-12-15
ERA-12344: das-web-react: Replace makeFieldsFromSchema with transformSchemaToFormElements and gate section rendering by condition evaluation

luixlive:
### What does this PR do?
Remove the legacy `makeFieldsFromSchema` and replace it with `transformSchemaToFormElements` to have the new `formElements` object that supports conditions. Also, use `evaluateSectionConditions` to calculate the visible sections depending on the current state of the form data.

### How does it look
image

### Relevant link(s)
* [ERA-12344](https://allenai.atlassian.net/browse/ERA-12344)
* [Env](https://era-12344.dev.pamdas.org)

`SchemaForm` component is the best place to start. Along with the main ticket changes, there are a few refactors in the file so it would be a good idea to fully review the file locally (<200 lines).
- New `getVisibleSectionIds` method to filter the sections using `evaluateSectionConditions`.
- Replaced `fields` with `formElements`.

From there, the rest of the changed files will be easier to review.

[ERA-12344]: https://allenai.atlassian.net/browse/ERA-12344?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/__test-helpers/fixtures/event-schemas.js
```diff
@@ -369,6 +368,7 @@ export const snareSchemaV2 = {
     },
     required: ['number_of_snares_found'],
     type: 'object',
+    unevaluatedProperties: false,
```
luixlive, line 371: I updated this repo to use `unevaluatedProperties` instead of `additionalProperties` in its fixtures and tests to match the latest v2 schemas.

src/ReportFormSummary/V2SchemaFormSummary/index.js
```diff
@@ -28,47 +28,37 @@ const FieldSummary = ({ field, formData, id }) => {
...
+        : <FieldSummary field={formElements[fieldId]} formData={formData} id={fieldId} key={fieldId} />)}
     </div>}
   </div>
 </div>;
 
-// For V2 schemas, we have a customized rendering of the details using utility functions like makeFieldsFromSchema and
-// getHumanizedFieldValue following the schema definition.
 const V2SchemaFormSummary = ({ eventSchema, formData }) => {
-  const fields = useMemo(() => makeFieldsFromSchema(eventSchema), [eventSchema]);
+  const formElements = useMemo(() => transformSchemaToFormElements(eventSchema), [eventSchema]);
```
luixlive, line 56: This is one of the two places that were using `makeFieldsFromSchema`. Now it is using `transformSchemaToFormElements` in order to properly render the summary of the v2 event type data:
Image

src/ReportManager/DetailsSection/SchemaForm/fields/Numeric/index.js
```diff
@@ -25,8 +25,8 @@ const Numeric = ({ autofillDefaultInput, details, error, id, onFieldChange, valu
     </label>
 
     <NumericInput
-        max={details.maxInput ?? null}
-        min={details.minInput ?? null}
```
luixlive, line 29: In the new utility `transformSchemaToFormElements` we are already defining default values or `null`s, so we don't need to add defensive code anywhere else. Also, the `placeholder` was standardized to `hint` since it was different for some fields.

src/ReportManager/DetailsSection/SchemaForm/utils/useSchemaValidations/index.js
```diff
@@ -5,6 +5,7 @@ import { useTranslation } from 'react-i18next';
 
 const ajv = new Ajv2020({ allErrors: true });
 addFormats(ajv);
+ajv.addKeyword({ keyword: 'x-section', schemaType: 'string' });
```
luixlive, line 8: We use AJV in strict mode, so extra properties are not supported. This line adds support for the custom `x-section` property in conditions.

src/ReportManager/DetailsSection/SchemaForm/index.js
```diff
@@ -21,12 +22,19 @@ export const FIELDS = {
   [FORM_ELEMENT_TYPES.TEXT]: Text,
 };
 
+const getVisibleSectionIds = (formElements, formData) =>
```
luixlive, line 25: Utility that evaluates the conditions of every section against a given `formData` object and returns an array with the sections that pass the validations. Basically, it returns which sections should be rendered (visible).

src/ReportManager/DetailsSection/SchemaForm/index.js
```diff
@@ -53,20 +62,14 @@ const SchemaForm = ({
...
 
-  // This ref works as a flag to trigger a useEffect and call onFormDataChange asynchronously when there are changes in
-  // the form data, so we can keep the onSectionFieldChange dependency array empty.
-  const shouldSendFormDataChangeRef = useRef(false);
-
   const [fieldErrors, setFieldErrors] = useState({});
-  const [formData, setFormData] = useState(initialFormData);
 
-  const fields = useMemo(() => makeFieldsFromSchema(schema), [schema]);
+  const formElements = useMemo(() => transformSchemaToFormElements(schema), [schema]);
```
luixlive, line 67: Here we replace `makeFieldsFromSchema` with  `transformSchemaToFormElements`.

src/ReportManager/DetailsSection/SchemaForm/index.js
```diff
@@ -85,22 +88,38 @@ const SchemaForm = ({
...
+    const newVisibleSectionIds = getVisibleSectionIds(formElements, newFormData);
+    const hiddenSectionIds = visibleSectionIds.filter((id) => !newVisibleSectionIds.includes(id));
+    const hiddenFields = hiddenSectionIds.flatMap((sectionId) => [
+      ...formElements[sectionId].details.leftColumn,
+      ...formElements[sectionId].details.rightColumn,
+    ]).filter((fieldId) => fieldId in formData);
+    if (hiddenFields.length > 0) {
+      // The field change resulted in a section being hidden. Remove the hidden
+      // section's fields from the form data.
+      hiddenFields.forEach((fieldId) => delete newFormData[fieldId]);
```
luixlive: Above we have a new logic on every field change to validate if the change will break a passing condition. In that case, the section will be hidden and the fields of that section should be removed from the form data.

src/ReportManager/ReportDetailView/index.js
```diff
@@ -311,7 +311,7 @@ const ReportDetailView = ({
       reportToSubmit = {
         ...reportChanges,
         id: reportForm.id,
-        event_details: { ...originalReport.event_details, ...reportChanges.event_details },
```
luixlive, line 314: We were merging the original event details with the new event details, but now that we have conditions, the original report may contain data of a conditional field that should not be included anymore (the condition was broken) so now I'm just injecting the current state of the event details form.

mock-api/api/v2.0/activity/eventtypes/fixtures.js
```diff
@@ -97,6 +96,7 @@ const eventtypeSchemaFixtures = {
       },
       required: ['animal_species', 'number_of_animals'],
       type: 'object',
+      unevaluatedProperties: false,
```
JoshuaVulcan, line 99: Small but great change 👍

src/ReportManager/DetailsSection/SchemaForm/fields/Text/index.js
```diff
@@ -45,7 +45,7 @@ const Text = ({ autofillDefaultInput, details, error, id, onFieldChange, value =
       className={`${styles.input} ${STYLES[details.inputType]}`}
       id={id}
       onChange={(event) => onFieldChange(id, event.currentTarget.value || undefined)}
-      placeholder={details.placeholder}
+      placeholder={details.hint}
```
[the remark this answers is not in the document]
luixlive: Yep, we did this shift a long time ago (Tiffany decided to label these as "hint") but we had already implemented a couple fields that were using the word "placeholder". I'm trying to standardize them.

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/index.js
```diff
@@ -11,32 +11,27 @@ import SortableList from './SortableList';
...
-// within other collections, so we propagate values, errors and breadcrumbs and their changes to the parent and the
-// children.
 const Collection = ({
   blurLocationMarker,
   breadcrumbs,
   details,
   error,
-  fields,
   focusLocationMarker,
+  formElements,
```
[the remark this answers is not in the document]
luixlive: All tests are passing and I did a couple tests 👍

src/ReportFormSummary/V2SchemaFormSummary/index.js
```diff
@@ -28,47 +28,37 @@ const FieldSummary = ({ field, formData, id }) => {
...
 
-  return fields[ROOT_CANVAS_ID]?.details.fields.map((sectionId) => <SectionSummary
-    details={fields[sectionId].details}
-    fields={fields}
+  return formElements[ROOT_CANVAS_ID]?.details.sections.map((sectionId) => <SectionSummary
```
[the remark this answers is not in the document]
luixlive: There's no need to worry about legacy schemas. The `formElements` object is generated in `transformSchemaToFormElements`, and this renaming doesn't affect the schema at all. We can be confident that this property will be called `sections` for all schemas.

src/ReportManager/DetailsSection/SchemaForm/fields/Numeric/index.js
```diff
@@ -25,8 +25,8 @@ const Numeric = ({ autofillDefaultInput, details, error, id, onFieldChange, valu
...
+        max={details.maxInput}
+        min={details.minInput}
```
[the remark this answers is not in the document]
luixlive: We are still doing the same, but in the transformation method instead of in React components, which is cleaner in my opinion.

src/utils/v2-event-schemas/getHumanizedFieldValue/index.js
```diff
@@ -7,28 +7,26 @@ import { shouldUse12HourFormat } from '../../datetime';
...
   .find((option) => option.const === value)?.title;
 
-// Utility to calculate a human readable version of the field values. For example, render a date-time like
-// 2020/01/01 12:00 PM instead of 2020-01-01T12:00:00Z.
+// Utility to calculate a human readable version of the field values. For
+// example, render a date-time like 2020/01/01 12:00 PM instead of
+// 2020-01-01T12:00:00Z.
 const getHumanizedFieldValue = (field, value, defaultHumanizedValue, language, coordinatesRepresentation, t) => {
-  if (!value) {
+  if (value === undefined) {
```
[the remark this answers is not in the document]
luixlive: Exactly! We were returning the default for falsy values that were valid like 0 or `false`. Now we only return the default if the value is explicitly `undefined` as it should be.

src/ReportManager/DetailsSection/SchemaForm/index.js
```diff
@@ -21,12 +22,19 @@ export const FIELDS = {
...
+    ));
+
 const SchemaForm = ({
   autofillDefaultInputs,
   eventId,
   eventLocation,
+  formData,
   hideMapLocationMarkers,
-  initialFormData,
   onFormDataChange,
```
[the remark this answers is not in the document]
luixlive: This is a good comment. I did some manual tests and it worked, but I will leave this conversation unresolved so we can make sure to do some thorough testing. Anyway, QA should do those tests independently of this change.

src/ReportManager/DetailsSection/SchemaForm/index.js
```diff
@@ -85,22 +88,38 @@ const SchemaForm = ({
...
+      ...formElements[sectionId].details.leftColumn,
+      ...formElements[sectionId].details.rightColumn,
+    ]).filter((fieldId) => fieldId in formData);
+    if (hiddenFields.length > 0) {
+      // The field change resulted in a section being hidden. Remove the hidden
+      // section's fields from the form data.
+      hiddenFields.forEach((fieldId) => delete newFormData[fieldId]);
+    }
+
+    onFormDataChange(newFormData);
```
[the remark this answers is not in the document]
luixlive: Nice suggestion!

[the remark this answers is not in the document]
luixlive: While working on this I noticed that there was an edge case bug in my implementation to remove fields from hidden sections from the form data. Conditional sections can depend on fields in other conditional sections:
Section 3 --> Section 2 --> Section 1

This means that when there's a change in the field of Section 1, it can logically result in having to delete the values from the fields in Section 2 and Section 3. But in order to do that, we need to calculate the hidden sections in a loop:
1- Field in Section 1 is changed
2- Now we need to hide Section 2
2.1- Remove the fields in Section 2
3- Now we need to hide Section 3
3.1- Remove the fields in Section 3

And this can go on and on. Basically, we need a loop until the form data is stable and no more sections are being hidden. I just did that change, and after that, I decided that it's better to keep the `delete` in `currentHiddenFields.forEach((fieldId) => delete newFormData[fieldId]);` since it makes the code more readable. It's already pretty complex to now add the suggested filtering.

JoshuaVulcan (approved):
Very nice -- extremely clean integration, thanks to some nice thoughtful technical design @luixlive  🙂

PR #1432, opened 2026-01-13
Updated style + subject popup treatment for ropeless fishing gearsets

JoshuaVulcan:
### What does this PR do?
- Be contextual and descriptive: i.e. What was changed or added.

### How does it look
- Screenshot of the changes, if applicable

### Relevant link(s)
* Tracking tickets: [ERA-\<ticket number\>](https://allenai.atlassian.net/browse/ERA-<ticket-number>)
* [Hosted demo environments](https://era-<ticket-number>.pamdas.org)
* Etc

- Describes the best way to approach or start reviewing this PR

- Any pertinent information that should be considered


src/ducks/subjects.js
```diff
@@ -63,6 +64,22 @@ const cancelableMapSubjectsFetch = () => {
         .then((response) => {
           if (response) {
             dispatch(fetchMapSubjectsSuccess(response));
+
+            // Auto-pin tracks for ropeless_buoy_gearset subjects (buoy site feature)
+            // Only when zoomed in (z > 10) to avoid unnecessary track loading when zoomed out
+            const currentZoom = map?.getZoom();
+            if (currentZoom && currentZoom > 8) {
```
luixlive, line 71: No need for the first check:
```Javascript
if (map?.getZoom() > 8) {
}
```

Also I see this code being repeated in `fetchSubjectGroups`. Maybe we could abstract this logic in a separate method? Or at least some part of it, like `getBuoyGearsetSubjects` that returns an empty array for zoom < 8.

src/Map/index.js
```diff
@@ -729,11 +728,9 @@ const Map = ({ children, onMapLoad, socket }) => {
       {subjectHeatmapAvailable && <SubjectHeatLayer />}
       {showReportHeatmap && <ReportsHeatLayer />}
 
-      {subjectTracksVisible && <TracksLayer onPointClick={onTimepointClick} showTimepoints={showTrackTimepoints} />}
+      {subjectTracksVisible && <TracksLayer onPointClick={onTimepointClick} onTrackLabelClick={onSelectSubject} showTimepoints={showTrackTimepoints} />}
```
luixlive, line 731: This will call `onSelectSubject` for all tracks, not only Buoy subjects, meaning that this code will always be called when the user clicks a track:
```Javascript
if (tracks_available) {
  dispatch(
    updateTrackState({ visible: [...subjectTrackState.visible, id] })
  );
}
```

Do we want that? If the user has some tracks pinned and clicks them, would their state change to visible?

src/MessagesModal/index.js
```diff
@@ -50,7 +51,7 @@ const MessagesModal = ({ onSelectSubject, selectedSubject = null }) => {
     <Modal.Header style={{ alignItems: 'center', h5: { margin: 0 }, height: '4rem' }}>
       {selectedSubject
         ? <h5 style={{ alignItems: 'center', display: 'flex' }}>
-          {selectedSubject.name}
+          {calcDisplayNameForSubject(selectedSubject.name)}
```
luixlive: This method expects to receive a subject, and it will extract the name property:
```Javascript
export const calcDisplayNameForSubject = (subject) => {
  const primaryName = subject.name || subject.title || '';

  ...
};
```

So I think this line is wrong, it should be:
```Javascript
{calcDisplayNameForSubject(selectedSubject)}
```

src/reducers/index.js
```diff
@@ -129,7 +130,7 @@ const rootReducer = combineReducers({
     patrolTrackState: patrolTracksReducer,
     mapImages: mapImagesReducer,
     mapIsLocked: mapLockStateReducer,
-    showMapNames: displayMapNamesReducer,
+    showMapNames: persistReducer(mapLabelPreferencesPersistenceConfig, displayMapNamesReducer),
```
luixlive, line 133: Why don't we follow the same pattern of the other persisted reducers of calling the persistence config with the same name? `persistReducer(displayMapNamesPersistenceConfig, displayMapNamesReducer)`

src/SideBar/SettingsPane/MapTab/MainMapSettingsView/MapMarkersFieldSet/index.js
```diff
@@ -17,7 +17,12 @@ const MapMarkersFieldSet = () => {
   });
 
   const hasUserLocation = useSelector((state) => !!state.view.userLocation);
-  const showMapNames = useSelector((state) => state.view.showMapNames);
+  const showMapNames = useSelector((state) => {
```
luixlive: ⚠️ this selector will impact this component's performance hugely. Selector functions run every time there is a Redux store update (in any reducer), and a selector with `useSelector` will trigger a re-render if the returned value changes. Here, the returned value will always be different because you are returning a new object: `{ ...state.view.showMapNames }`. In other words, every single time our store gets updated, this component will be re-rendered unnecessarily.

Instead, you should delete the unwanted property outside of the selector:
```Javascript
const showMapNames = useSelector((state) => state.view.showMapNames);

const cleanShowMapNames = { ...showMapNames };
delete cleanShowMapNames._persist;
```

src/SoundNotificationsPlayer/RadioStateChangeToAlarmNotificationPlayer/index.js
```diff
@@ -9,13 +9,14 @@ import DateTime from '../../DateTime';
...
 import * as styles from './styles.module.scss';
+import { calcDisplayNameForSubject } from '../../utils/subjects';
 
 const AlarmRadioStateToastMessage = ({ onClickJumpToLocation, subject }) => {
   const { t } = useTranslation('components', { keyPrefix: 'soundNotificationsPlayer' });
 
   return <div className={styles.alarmRadioStateToast}>
     <div>
-      {t('alarmRadioStateToastMessage', { subjectName: subject.name })}
+      {t('alarmRadioStateToastMessage', { subjectName: calcDisplayNameForSubject(subject.name) })}
```
luixlive: Here too you are passing the subject's name instead of the full subject object, as the function expects it.

src/SubjectControls/index.js
```diff
@@ -119,7 +119,7 @@ const SubjectControls = ({
         content: SubjectHistoricalDataModal,
         subjectId: subject.id,
         subjectIsStatic: subjectIsStatic(subject),
-        title: t('historicalDataModalTitle', { subjectName: subject.name }),
+        title: t('historicalDataModalTitle', { subjectName: calcDisplayNameForSubject(subject.name) }),
```
luixlive: x3

src/SubjectMessagePopup/index.js
```diff
@@ -14,7 +14,7 @@ const SubjectMessagePopup = ({ data }) => {
   const { subject, message } = data;
   return <>
     <h6 onClick={() => jumpToLocation(getSubjectLastPositionCoordinates(subject))}>
-      <ChatIcon /> {subject.name}
+      <ChatIcon /> {calcDisplayNameForSubject(subject.name)}
```
luixlive: x4

src/SubjectMessagesPopover/index.js
```diff
@@ -26,7 +27,7 @@ const SubjectMessagesPopover = ({ className = '', subject = null, ...restProps }
   const popover = <Popover className={styles.popover}>
     <Popover.Header>
       <h6>
-        <ChatIcon /> {subject.name}
+        <ChatIcon /> {calcDisplayNameForSubject(subject.name)}
```
luixlive: x5

src/SubjectMessagesPopup/index.js
```diff
@@ -18,7 +19,7 @@ const SubjectMessagesPopup = ({ data }) => {
   return <>
     <div className={styles.header}>
       <h6>
-        <ChatIcon /> {properties.name}
+        <ChatIcon /> {calcDisplayNameForSubject(properties.name)}
```
luixlive: x6

src/SubjectPopup/index.js
```diff
@@ -18,6 +18,8 @@ import * as styles from './styles.module.scss';
 
 const STORAGE_KEY = 'showSubjectDetailsByDefault';
 
+
```
luixlive, line 21: I see a few empty lines in this file, plus another place where we are passing the name instead of the subject to the method `calcDisplayNameForSubject(properties.name)`.

src/SubjectsLayer/index.js
```diff
@@ -74,23 +74,68 @@ const SubjectsLayer = ({ mapImages = {}, onSubjectClick }) => {
     }
   }]);
 
+  // Override icon-image and icon-size for ropeless_buoy_gearset subjects to use ferry-11
+  // Note: icon-allow-overlap and icon-ignore-placement don't support data expressions,
+  // so we set them globally to ensure buoys are always visible
+  const buoyIconLayout = {
```
luixlive, line 80: The indentation, new lines, commas, in these blocks makes it very difficult to understand the Mapbox logical expressions. We could do something a bit smoother for readability:
```Javascript
const buoyIconLayout = {
  // Icon image
  'icon-image': [
    'case',

    // Ropeless buoy gearset: ferry-11
    ['==', ['get', 'subject_subtype'], 'ropeless_buoy_gearset'],
    'ferry-11',

    // Default: build icon name dynamically "image-width-height"
    [
      'concat',
      ['get', 'image'],
      '-',
      ['case', ['has', 'width'], ['get', 'width'], 'x'],
      '-',
      ['case', ['has', 'height'], ['get', 'height'], 'x'],
    ],
  ],

  // Icon size
  'icon-size': [
    'interpolate',
    ['exponential', 0.5],
    ['zoom'],

    // Zoom level 0
    0,
    [
      // Ropeless buoy gearset: 1.5
      'case',
      ['==', ['get', 'subject_subtype'], 'ropeless_buoy_gearset'],
      1.5,

      [
        'case',
        ['in', 'generic', ['get', 'image']],
        0.1 / 3,
        0.2 / 3,
      ],
    ],

    // Zoom level 11
    11,
    [
      // Ropeless buoy gearset: 1.85
      'case',
      ['==', ['get', 'subject_subtype'], 'ropeless_buoy_gearset'],
      1.85,

      [
        'case',
        ['in', 'generic', ['get', 'image']],
        0.4 / 3,
        0.8 / 3,
      ],
    ],

    // Zoom level 14
    14,
    [
      // Ropeless buoy gearset: 2.5
      'case',
      ['==', ['get', 'subject_subtype'], 'ropeless_buoy_gearset'],
      2.5,

      [
        'case',
        ['in', 'generic', ['get', 'image']],
        0.5 / 3,
        1 / 3,
      ],
    ],
  ],

  // These don't support data expressions, so we set them globally to ensure buoys are always visible.
  'icon-allow-overlap': true,
  'icon-ignore-placement': true,
};
```

src/TracksLayer/track.js
```diff
@@ -117,17 +138,110 @@ const TrackLayer = ({
...
+  const trackLabelSource = useMemo(() => {
+    if (!isRopelessBuoyGearset || !trackData.track.features.length) return null;
+
+    return {
+      type: 'FeatureCollection',
+      features: [trackData.track.features[0]]
+    };
+  }, [isRopelessBuoyGearset, trackData.track.features]);
+
+  useMapSources(trackLabelSource ? [{ id: trackLabelSourceId, data: trackLabelSource }] : [], { tolerance: 1.5, type: 'geojson', lineMetrics: true });
```
luixlive, line 178: `useMapSources` and `useMapLayers` don't cleanup resources properly because it's logically impossible to make them call the map function in this order: Add source, add layer -> remove layer, remove source.

If you call `useMapSources` and then `useMapLayers`, the component will try to remove the source before cleaning its layers.

If you call useMapLayers` and then `useMapSources`, the component will try to add a layer to a source that hasn't been created yet.

For that reason I've been avoiding them, to avoid cluttering our console and memory leaks. The cleanest approach is to trigger the `map` functions in `useEffect`s with clean dependency arrays like in: https://github.com/PADAS/das-web-react/blob/develop/src/ReportManager/DetailsSection/SchemaForm/utils/useMapLocationMarkers/index.js

src/TracksLayer/track.js
```diff
@@ -117,17 +138,110 @@ const TrackLayer = ({
...
+    }
+  }]);
+
+  // Add event handlers for track label layer
+  const onLabelClick = useCallback((event) => {
+    if (!subject || !onTrackLabelClick) return;
+
+    const { lngLat } = event;
+    // Create synthetic layer object matching the subject layer structure
+    const syntheticLayer = {
```
luixlive, line 209: I'm a bit lost here, why do we need a synthetic layer? Why not to send the actual one?

.env.development
```diff
@@ -1,4 +1,4 @@
-REACT_APP_DAS_HOST=''
+REACT_APP_DAS_HOST='https://buoy.pamdas.org'
```
luixlive: I guess we don't want to keep this string. Also in `.env.production`.

luixlive (commented):
I know you are still working on tests and maybe some tweaks, but since I was already going through the code, I left some minor comments I think would improve this PR a little 😄

luixlive:
I can see the ferries and their tracks in the env, all fine with that. I just gotta say that this logic of auto-pinning the buoy tracks feels kind of buggy IMO, by design. When I manually select only one, but move the map a little and the backend responds with more map subjects, they get automatically selected... Feels a bit annoying for me, but if that's what we really want, then all is ok from my side:
image

src/ducks/subjects.js
JoshuaVulcan: The fetchSubjectGroups code is old and unnecessary, I'll remove it shortly.

src/Map/index.js
JoshuaVulcan: The event binding is set up as such:

`useMapEventBinding('click', onLabelClick, trackLabelLayerId, isRopelessBuoyGearset && trackLabelText !== '');`

So the event won't be fired unless the click is on a non-empty label for ropeless fishing gear. that seemed like an exclusive enough code path to me

JoshuaVulcan:
> I can see the ferries and their tracks in the env, all fine with that. I just gotta say that this logic of auto-pinning the buoy tracks feels kind of buggy IMO, by design. When I manually select only one, but move the map a little and the backend responds with more map subjects, they get automatically selected... Feels a bit annoying for me, but if that's what we really want, then all is ok from my side:

Totally agreed. it's a unique and generally janky use case. but it is indeed the desired behavior.

src/SubjectsLayer/index.js
JoshuaVulcan: IMO this is much more legible as-is than suggested, though the GitHub code window in-browser seems to get a bit scrambled on it. Here's what it looks like in an IDE:

Screenshot 2026-01-20 at 3 18 12 PM

In any case, if this is just stylistic, I'm going to recommend a good ol' fashioned punt 😅

luixlive: Yeah it's just stylistic preference. I think these Mapbox expressions are hard to read by design haha.

I just feel some of the new lines are too complex to read in single lines, I think some new lines with different indentation, empty new lines to separate sub-expressions and a few comments could make it easier to go through. For example:
`0, ['case', ['==', ['get', 'subject_subtype'], 'ropeless_buoy_gearset'], 0.75, ['case', ['in', 'generic', ['get', 'image']], 0.1 / 3, 0.2`

But yeah, it's a matter of preference. Not a blocker.

luixlive (dismissed):
The only thing I would like to put emphasis on from my previous comments is the recommendation on deprecating `useMapSources` and `useMapLayers`. That pattern has been throwing warnings and errors for a few years in our console and it's simply not cleaning the map resources correctly. It's not a blocker, since anyway we are using them in other places, but in my opinion it's a matter of deciding if we increment that technical debt or continue to tackle it.

PR #1440, opened 2026-01-31
Hotfix: always render all sections and hide them using "hidden" attribute to avoid mounting and unmounting issues

luixlive:
**Issue:** Alan configured a multiple choice list field rendered as a checkbox list with the "Subjects from Subject Sub-Type" dynamic data. Then, a conditional section depends on if that choice list has a value. The conditional section renders two other fields, one of them being another multiple choice list field, but this one rendered as a dropdown. However, in the client app, when filling the form, the checkbox list items require two clicks in order to be rendered, but the conditional section shows up after the first click even though the checkbox is not checked yet.

**Explanation:** NASTY STUFF. And I can wash my hands here, I traced back the error (took me a while), and it seems like the bug is in "react-select” library. The issue doesn’t originate in the checkbox list that requires two clicks, it originates from the dropdown that is rendered in the left column of the conditional section. If you go to the EFE and change that dropdown to be a list (another checkbox list since it’s a multi-select), things work as expected. The issue only happens when we render a “react-select” dropdown in a conditional section.

**Fix:** Moving away from "react-select” would be great, but that will take time and a good amount of code. By trying different things, I noticed that the issue is related to the mounting of the "react-select” component. You see, conditional sections are not rendered at all if the condition doesn’t pass the evaluation. Once the evaluation passes, sections and all their content are mounted. A trivial fix is to change that paradigm, and instead always render all sections, but hide with the ”hidden” HTML attribute (similar to display: none) sections with failing evaluations. This way, the “react-select” components are always rendered, and the issue is gone.

⚠️ Important ⚠️: We need to merge down these changes to `develop` branch.


src/ReportManager/DetailsSection/SchemaForm/fields/Section/index.js
```diff
@@ -22,6 +23,7 @@ const Section = ({
   return <div
       className={styles.section}
       data-testid={`schema-form-section-${id}`}
+      {...otherProps}
```
luixlive, line 26: Now sections receive extra properties that are spread to the `div`. This is how we inject the `hidden` attribute from the `SchemaForm` component in the the section parent `div`.

src/ReportManager/DetailsSection/SchemaForm/index.js
```diff
@@ -191,11 +191,12 @@ const SchemaForm = ({
   }, [formData, formElements, setLocationMarkers]);
 
   return <form onSubmit={onSubmit}>
-    {visibleSectionIds.map((sectionId) => <Section
+    {formElements[ROOT_CANVAS_ID]?.details.sections.map((sectionId) => <Section
```
luixlive, line 194: We used to only render "visible sections" (aka sections with passing conditions). Now we render all sections but hide the ones that have failing conditions using the `hidden` attribute.

JoshuaVulcan (approved):

PR #1446, opened 2026-02-05
ERA-12624: V2 schema forms with multiple fields with default values only fills one of the inputs

luixlive:
### What does this PR do?
Fix auto-filling of fields with default inputs in v2 schema forms.

### Relevant link(s)
* [ERA-12624](https://allenai.atlassian.net/browse/ERA-12624)

[ERA-12624]: https://allenai.atlassian.net/browse/ERA-12624?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/index.js
```diff
@@ -12,14 +12,7 @@ const INPUTS = {
   [CHOICE_LIST_ELEMENT_INPUT_TYPES.LIST]: List
 };
 
-const ChoiceList = ({
-  autofillDefaultInput: _autofillDefaultInput,
```
luixlive, line 16: The `autofillDefaultInput` prop was removed, I will explain more a bit down below.

In a couple field files the update simply consists on removing the ignored prop. They receive it from the `SchemaForm` renderer, but they don't need it because these field types don't even support default inputs.

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/SortableList/Item/index.js
```diff
@@ -112,6 +112,23 @@ const Item = ({
     shouldDeleteOnCancelRef.current = false;
   };
 
+  useEffect(() => {
```
luixlive, line 115: There are 3 scenarios where we want to automatically populate the default value in the inputs. This is the first one: when the user clicks the "Add Item" button of a collection. The collection item will have it's own form data, and it must be pre-populated with the default values of the item's fields.

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/index.js
```diff
@@ -25,17 +25,14 @@ const Collection = ({
 }) => {
   const { t } = useTranslation('reports', { keyPrefix: 'reportManager.detailsSection.schemaForm.fields.collection' });
 
-  // Ref to keep track of the temporal id of the last added item so we keep
-  // incrementing them when the user adds more items.
-  const lastAddedItemIdRef = useRef(value.length - 1);
```
luixlive, line 30: I found a bug related to this logic of setting an id by keeping the last id in this `ref`. Seems like Gabo did some work a couple months back where he added this logic of decreasing the number of this ref when deleting an item, but it wasn't working properly. It was causing duplicated keys in React arrays during render, degrading the UX.

Instead of keeping track of the last id through a ref, now I do the calculation every time the user adds a new item to a collection:
```Javascript
const highestExistingItemId = items.reduce((highestItemId, item) => Math.max(highestItemId, item.id), -1);
```

It's literally an `O(n)` where `n < 10` that runs only when the user adds an item. That means: nothing to worry about, and we have cleaner code now 😄

src/ReportManager/DetailsSection/SchemaForm/fields/Numeric/index.js
```diff
@@ -1,24 +1,14 @@
...
 
 import * as styles from './styles.module.scss';
 
-const Numeric = ({ autofillDefaultInput, details, error, id, onFieldChange, value = '' }) => {
+const Numeric = ({ details, error, id, onFieldChange, value = '' }) => {
   const hasError = !!error;
   const hasDescription = !!details.description && !hasError;
   const label = details.isRequired ? `${details.label} *` : details.label;
 
-  const shouldAutofillDefaultInputRef = useRef(autofillDefaultInput && details.defaultInput);
```
luixlive, line 12: The only field types that supported default inputs before Booleans were Numeric and Text. This is the buggy code.

We were relying on a `useEffect` that triggered `onFieldChange` while mounting the field if it has a default input. The issue: if you have a form with 3 fields with default inputs, they are mounting at the same time (React render cycle), thus each one of them trigger the callback `onFieldChange`, which resulted in a race. Only one of the calls were being executed, affect the actual `formData` state in the next render. I'm cleaning all that code here.

src/ReportManager/DetailsSection/SchemaForm/fields/Section/index.js
```diff
@@ -9,22 +9,43 @@ const Section = ({
...
   };
 
-  return <div
-      className={styles.section}
-      data-testid={`schema-form-section-${id}`}
-      {...otherProps}
-    >
+  useEffect(() => {
+    const sectionBecameVisible = previousHiddenRef.current && !hidden;
+    if (sectionBecameVisible) {
```
luixlive, line 29: Here we have the second scenario where we want to automatically populate the default values in the form fields: when a hidden section becomes visible. When does a hidden section becomes visible? Right now there's only once answer to that: conditional sections with conditions not being fulfilled in the last render, but being fulfilled in the current one. However, in the future we may have more scenarios were a section becomes visible, and this `useEffect` will cover all of them.

src/ReportManager/DetailsSection/SchemaForm/index.js
```diff
@@ -164,6 +164,31 @@ const SchemaForm = ({
     }
   };
 
+  useEffect(() => {
+    if (shouldAutofillDefaultInputs) {
```
luixlive, line 168: And the third and most common of the 3 scenarios were we want to automatically populate the default fields: when mounting the form for a new event. When the user starts creating a new event, and the form gets mounted, we automatically fill the values of those fields that have a default input FOR THE VISIBLE SECTIONS ONLY. We don't want to fill a field from a hidden section (conditional section that hasn't been rendered) and include that value in the form data, that wouldn't pass the JSON schema validation since the condition isn't fulfilled!

src/ReportManager/DetailsSection/SchemaForm/fields/Section/index.js
```diff
@@ -9,22 +9,43 @@ const Section = ({
...
+      const defaultFormData = sectionChildrenIds.reduce((accumulator, sectionChildId) => {
+        if (formElements[sectionChildId].details.defaultInput) {
+          accumulator[sectionChildId] = formElements[sectionChildId].details.defaultInput;
+        }
+        return accumulator;
+      }, {});
+
+      setDefaultFormData(defaultFormData);
+    }
+  }, [details.conditions?.length, details.leftColumn, details.rightColumn, formElements, hidden, setDefaultFormData]);
```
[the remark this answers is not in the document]
luixlive: Remember that all content of the `formElements` variable (including this `details` object that is actually a property from one of the form elements) never change. That information comes from the schema and will never be altered unless a change in the schema is done, which means the page should refresh entirely. 

Anyway, this is a good catch. That was in the dependency array because I was using it, but then removed it and forgot to clean it from here. Will remove it 👍

.env.development
```diff
@@ -1,4 +1,4 @@
-REACT_APP_DAS_HOST=''
+REACT_APP_DAS_HOST='https://root.dev.pamdas.org'
```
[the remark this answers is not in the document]
luixlive: Removing it!

src/ReportManager/DetailsSection/SchemaForm/index.js
```diff
@@ -164,6 +164,31 @@ const SchemaForm = ({
...
+        ];
+        sectionChildrenIds.forEach((sectionChildId) => {
+          if (formElements[sectionChildId].details.defaultInput) {
+            accumulator[sectionChildId] = formElements[sectionChildId].details.defaultInput;
+          }
+        });
+        return accumulator;
+      }, {});
+
+      onFormDataChange(initialFormData);
```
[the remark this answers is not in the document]
luixlive: Nice suggestion, doing it. The following two are basically the same for the other two places where we set the defaults. Will do the changes.

JoshuaVulcan (approved):
Works for me, following the repro steps from the issue yesterday. @AlanCalvillo can we regress and release this AM?

PR #1460, opened 2026-02-18
Migrate PR workflow to inline Docker build with dual registry

andersondario-er:
## Summary

- Replace `gundi-workflows/build_docker.yml` with inline Docker build logic (das-web-react is a **public** repo and cannot call reusable workflows from serca-pipelines which is internal)
- Remove deploy label gate: build runs on **every PR**
- Add dual registry push: serca-artifact-registry (primary) + padas-app (secondary)
- Remove GitHub environment declaration
- Add GHA cache (`cache-from`/`cache-to`) for faster builds
- Image tags: `{sha}` and `pr-{sha}` pushed to both registries
- Add concurrency group with cancel-in-progress

## Test plan

- [ ] Open a test PR and verify config + build jobs run without deploy label
- [ ] Verify Docker image is pushed to both serca-artifact-registry and padas-app
- [ ] Verify image tags match expected format (`{sha}` and `pr-{sha}`)
- [ ] Verify GHA cache is populated on first run and used on subsequent pushes

### DIFF
--- a/.github/workflows/pr-envs.yml
+++ b/.github/workflows/pr-envs.yml
@@ -1,26 +1,30 @@
-name: Build das-server on Pull Request
+name: Pull Request Workflow
+
 on:
   pull_request:
     types:
       - labeled
       - synchronize
-      - reopened
+      - opened
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


andersondario-er:
Recreating with dash branch name (no slashes)

PR #1467, opened 2026-02-25
Migration to vite

luixlive:
### What does this PR do?
- Migrate our ejected Create React App configuration (Webpack + Babel) to Vite.
- Upgrade to Node 24
- Upgrade to Yarn 4.12
- Upgrade all the project dependencies


.github/workflows/develop-workflow.yml
```diff
@@ -33,7 +33,7 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@v4
         with:
-          node-version: 22
+          node-version: 24
```
luixlive: Now all our workflows and Docker images reference Node version 24.

.yarn/releases/yarn-4.12.0.cjs
luixlive, line 1: Upgrade to Yarn 4.12 🎉

__mocks__/scss.js
luixlive, line 1: All Jest configuration related files were moved to a new `jest-configuration/` folder to keep things organized and the root directory clean.

jest-config/transformers/babelTransform.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+const babelJest = require('babel-jest');
+
+module.exports = babelJest.createTransformer({
```
luixlive, line 5: We moved away from Babel in our building process, but we still need it for Jest. However, I moved away entirely from Create React App leftovers. I created our own Babel transformer with up-to-date presets and plugins.

jest-config/transformers/transformImportMetaEnv.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+// Custom Babel plugin to transform `import.meta.env.*` into `process.env.*`.
```
luixlive, line 3: Vite makes environment variables available to the client in `import.meta.env`, but Jest will expect them to be in `process.env`. This is a simple custom transformer that transforms that object.

react-app-config/webpack/persistentCache/createEnvironmentHash.js
luixlive, line 1: Now all the configuration files that I removed from the Create React App ejection!

src/AddItemButton/index.js
```diff
@@ -80,7 +80,7 @@ const AddItemButton = ({
 
     <button
       aria-label={t('defaultLabel')}
-      className={`${styles[`addItemButton-${variant}`]} ${className}`}
+      className={`${styles[`addItemButton${variant.charAt(0).toUpperCase() + variant.slice(1)}`]} ${className}`}
```
luixlive: A simple fix for a SCSS issue. Vite transformation will make CSS class names with dashes into camel case.

src/Auth0TokenManager/index.js
```diff
@@ -40,7 +40,7 @@ const Auth0TokenManager = () => {
           // Auth0Provider has processed the callback, now get the token
           const token = await getAccessTokenSilently({
             authorizationParams: {
-              audience: process.env.REACT_APP_AUTH0_AUDIENCE,
+              audience: import.meta.env.REACT_APP_AUTH0_AUDIENCE,
```
luixlive, line 43: Now we access the environment variables through `import.meta.env`.

src/ducks/event-schemas/sanitizeSchemas.js
```diff
@@ -11,50 +11,54 @@ const GLOBAL_UI_SCHEMA_CONFIG = {
   Details: { 'ui:widget': 'textarea' },
 };
 
-const convertSchemaEnumNameObjectsIntoArray = (schema) => {
-  const { properties } = schema;
-  const updates = Object.entries(properties).reduce((propsObject, [key, value]) => {
+const convertSchemaEnumNameObjectsIntoArray = (schema, path = '') => {
```
luixlive, line 14: Just a couple of the dependency upgrades required code changes. One of them was `rjsf`. Even though we are moving away from it in v2 schemas, I decided to still migrate to the new major version (v6). They did a few changes in the API, here more info: https://rjsf-team.github.io/react-jsonschema-form/docs/migration-guides/v6.x%20upgrade%20guide/

src/EventsLayer/index.js
```diff
@@ -2,7 +2,7 @@ import React, { memo, useCallback, useContext, useEffect, useMemo, useRef, useSt
 import { featureCollection } from '@turf/turf';
 import { useSelector } from 'react-redux';
 
-import ClusterIcon from '../common/images/icons/cluster-icon.svg';
+import ClusterIcon from '../common/images/icons/cluster-icon.svg?url';
```
luixlive, line 5: With the new Vite transformation, in order to import a SVG as URL, we need to add `?url`.

src/ReportedBySelect/index.js
```diff
@@ -1,6 +1,6 @@
-import React, { memo, useCallback, useMemo, useState } from 'react';
+import React, { memo, useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { components } from 'react-select';
-import { FixedSizeList } from 'react-window';
+import { List } from 'react-window';
```
luixlive, line 3: Another upgrade that required a few changes, `react-window`. They changed the library considerably in version 2. This file is the only place where we use it, and the changes basically consist on a basic migration: https://github.com/bvaughn/react-window/blob/main/CHANGELOG.md#200

src/SchemaFields/index.js
```diff
@@ -34,40 +34,36 @@ export const AddButton = ({ className, uiSchema: _uiSchema, ...restProps }) => {
 };
 
 export const ArrayFieldItemTemplate = ({
+  buttonsProps,
```
luixlive, line 37: More changes related to moving to `rjsf` 6.

.eslintrc
luixlive, line 1: Moved to Eslint 9, which means flat configuration.

vite.config.js
```diff
@@ -0,0 +1,47 @@
+import { defineConfig } from 'vite';
+import react from '@vitejs/plugin-react';
+import svgr from 'vite-plugin-svgr';
+
+export default defineConfig({
```
luixlive, line 5: A simple, clean Vite configuration. Our project tooling will never be a magic black box again 🥹

src/SubjectHistoricalDataModal/index.test.js
```diff
@@ -45,7 +45,13 @@ describe('SubjectHistoricalDataModal', () => {
       <SubjectHistoricalDataModal title='Historical data' subjectId='fake-id' fetchObservationsForSubject/>
     </Provider>);
 
-    expect(fetchObservationsForSubject).toHaveBeenCalledWith({ 'page': 1, 'page_size': ITEMS_PER_PAGE, 'subject_id': 'fake-id', sort_by: SORT_BY });
```
luixlive, line 48: I think @JoshuaVulcan sent a commit adding a new parameter "include_empty_location" when fetching observations but forgot to run the tests 😅  Fixed!

src/setupTests.js
```diff
@@ -1,9 +1,17 @@
 import '@testing-library/jest-dom';
 import 'jest-webgl-canvas-mock';
+import dotenv from 'dotenv';
 import ReactGA4 from 'react-ga4';
 
 import MockSocketContext, { SocketContext } from './__test-helpers/MockSocketContext';
 
+// Set test environment variables.
+dotenv.config();
```
luixlive, line 9: We have a Babel transform plugin that transforms all occurrences of `import.meta.env` in the code to `process.env`. Here we inject the environment variables we use into `process.env` so the code can find them.

vite.config.js
```diff
@@ -0,0 +1,47 @@
...
+
+  // Match CRA environment variable prefix.
+  envPrefix: 'REACT_APP_',
+
+  // Transform .js files as if they were .jsx files.
+  esbuild: {
+    loader: 'jsx',
+    include: /.*\.jsx?$/,
+    exclude: []
+  },
```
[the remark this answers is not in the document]
luixlive: I don't see any performance impact related to this. This esbuild loader only affects files processed by Vite, which excludes node modules by default.

.yarn/install-state.gz
JoshuaVulcan, line 1: Can we please exclude this file while we're in the business of cleaning things up?

src/AddItemButton/index.js
JoshuaVulcan: Pretty gross as an inline transform though, I can see this biting in the future. Can we fix the CSS classes upstream or encapsulate this better?

src/ReportedBySelect/index.js
```diff
@@ -108,30 +108,34 @@ const getOptionValue = ({ hidden, id }) => {
   return id;
 };
 
-const MenuRow = memo(({ child, style }) => <div style={style}>{child}</div>); /* eslint-disable-line react/display-name */
+const MenuRow = ({ children, index, style }) => <div style={style}>{children[index]}</div>;
```
JoshuaVulcan, line 111: This seems like a deoptimization, what's the point of the change?

src/ReportManager/DetailsSection/SchemaForm/fields/Boolean/index.js
```diff
@@ -18,7 +18,7 @@ const Boolean = ({ details, error, id, onFieldChange, readOnly, value = false })
       aria-invalid={hasError}
       aria-required={details.isRequired}
       checked={value}
-      className={styles.switch}
+      className={styles.switchField}
```
JoshuaVulcan, line 21: This is just a stylistic change?

src/SchemaFields/index.js
```diff
@@ -34,40 +34,36 @@ export const AddButton = ({ className, uiSchema: _uiSchema, ...restProps }) => {
...
-  hasRemove,
-  index,
-  onDropIndexClick,
-  onReorderClick,
   readonly,
   registry,
   uiSchema,
 }) => {
   const { MoveDownButton, MoveUpButton, RemoveButton } = registry.templates.ButtonTemplates;
+  const { hasMoveDown, hasMoveUp, hasRemove, onMoveDownItem, onMoveUpItem, onRemoveItem } = buttonsProps || {};
```
JoshuaVulcan, line 46: non-blocking: I'm not a huge fan of bundling properties like this, it adds more steps to reason through the interface and usually shows stress-points where things need additional componentization.

src/SchemaFields/index.test.js
```diff
@@ -9,9 +9,9 @@ describe('the ExternalLinkField field',  () => {
     render(
       <ExternalLinkField
         formData='https://testing-this-neat-thing.biz'
-          idSchema={{ id: 'link-1' }}
-          schema={{ title: 'here is your great link' }}
-        />
+        fieldPathId={{ $id: 'link-1', path: [] }}
```
JoshuaVulcan, line 12: Calling this an `id` but it's an object with multiple properties seems misleading.

src/utils/event-schemas.js
```diff
@@ -1,4 +1,7 @@
 export const getLinearErrorPropTree = (errorProperty) => {
+  if (errorProperty == null || typeof errorProperty !== 'string') {
```
JoshuaVulcan, line 2: What specific issues is this guarding against? Seems like a place to sniff out type consistency issues further upstream.

Dockerfile.mt
```diff
@@ -1,7 +1,5 @@
-FROM europe-west3-docker.pkg.dev/serca-artifact-registry/virtual-docker/node:22-alpine AS build
+FROM europe-west3-docker.pkg.dev/serca-artifact-registry/virtual-docker/node:24-alpine AS build
 ARG ENV_FILE=.env.production
-ARG REACT_APP_GA4_TRACKING_ID
-ENV REACT_APP_GA4_TRACKING_ID=$REACT_APP_GA4_TRACKING_ID
```
JoshuaVulcan, line 4: Why remove the google analytics tracking ID ARG+ENV declarations here?

JoshuaVulcan (approved):
Nothing blocking, just a few (mostly stylistic) questions and thoughts. This absolutely rips (complimentary) locally compared to before. Nicely done 👏

src/ReportedBySelect/index.js
luixlive: This change is necessary after updating `react-window` to the latest version. We are using the "Fixed row heights" pattern: https://react-window.vercel.app/list/fixed-row-height

In those docs you can see that the way to send props to the children changed. The props you desire to read in the children need to be arrays ordered like the rows, and send them through `rowProps`. This is the example of the docs:
```Javascript
import { List } from 'react-window';

const RowComponent = ({ index, names, style }) => (
  <div className="flex items-center justify-between" style={style}>
    {names[index]}
    <div className="text-slate-500 text-xs">{`${index + 1} of ${names.length}`}</div>
  </div>
);
 
const Example = ({ names } => (
  <List
    rowComponent={RowComponent}
    rowCount={names.length}
    rowHeight={25}
    rowProps={{ names }}
  />
);
```

If you change the `names` variable from the example with our `children`, you get the same pattern. There's no way around it, at least not a clean one. I'm not a big fan either, but I'm not sure if it's really a de-optimization. I mean, we are passing an array by reference, it's not like this consumes more memory, and at the end we are rendering the same, so it doesn't involve more processing either. It's just a weird pattern they stablished.

src/ReportManager/DetailsSection/SchemaForm/fields/Boolean/index.js
luixlive: Not really. It seems like Vite didn't like the name `switch`. It wasn't resolving it properly. I didn't investigate further, but my theory is that the issue comes from `switch` being a reserved keyword in Javascript (you know, `switch` / `case` ). Webpack and it's SASS transformer didn't have issue with this name, but for some reason Vite does 🤷‍♂️

src/SchemaFields/index.js
luixlive: Sorry, I'm not sure if I get your comment. You mean that you don't like the fact that all these properties come within an object instead of individual props as before?

If that's what you mean, again I agree, but this is something I didn't change. Just like the `react-window` change above, this is something changed in the latest version of `rjsf`.

Here you can see the breaking changes to the `ArrayFieldItemTemplate` (like ours): https://rjsf-team.github.io/react-jsonschema-form/docs/migration-guides/v6.x%20upgrade%20guide/#arrayfielditemtemplate-breaking-changes

`The ArrayFieldItemTemplateProps was refactored to extract the following props into out a new ArrayFieldItemButtonsTemplateProps`

src/SchemaFields/index.test.js
luixlive: Another change that comes from the new version of `rjsf`: https://rjsf-team.github.io/react-jsonschema-form/docs/migration-guides/v6.x%20upgrade%20guide/#idschema-replacement-breaking-change

src/utils/event-schemas.js
luixlive: Actually, this code was added by Cursor. I followed the migration guide and then asked Cursor to read the full doc (https://rjsf-team.github.io/react-jsonschema-form/docs/migration-guides/v6.x%20upgrade%20guide/) and validate that I did the correct changes. It helped tweaking a couple things, and one of those was adding this. I remember asking why, but didn't remember, so I just asked it again 😅 

> The defensive check is there because RJSF v6 changed how paths and errors work. The upgrade guide explains that IdSchema (string-based) was replaced by FieldPathId and path arrays, and that the way errors are built (e.g. via toErrorList() from @rjsf/utils) changed. So we can’t rely on every error object having a string property anymore; in some cases (e.g. root-level or certain schema constructs) it may be missing or in a different shape.
> 
> getLinearErrorPropTree expects a string (the old dot-notation path like ".properties['reportcounty'].enum") and uses .replace() and .split() on it. It’s used in transformErrors for every error and again in filterOutErrorsForHiddenProperties. If any error has property equal to null, undefined, or a non-string (e.g. an array), calling .replace() on it would throw and break the form. The condition guards against that by returning [] for invalid/missing property, so those errors are handled safely (e.g. no path segments / no field match) instead of causing a crash.

Seems like Cursor expect some edge scenarios where the error could be `null` or not a string. What I assume is that those are very specific errors, that probably we are not even hitting with our use case of the library. When I was testing this I didn't find anything weird, but I would expect to do a couple more thorough checks on erroneous scenarios during the regression.

Dockerfile.mt
luixlive: Actually, I don't understand we used to have those lines there. Our pipeline is designed to inject the tracking id from our `.env` files. The variable `REACT_APP_GA4_TRACKING_ID` is present in both `.env.development` and `.env.production`, with different values (for each environment of course).

You can see in the second line of this file the `ARG ENV_FILE=.env.production`. From there we read the whole file and inject it a couple lines below like:
```
RUN test -f "$ENV_FILE" && cp "$ENV_FILE" .env.production.local
```

So the lines:
```
ARG REACT_APP_GA4_TRACKING_ID
ENV REACT_APP_GA4_TRACKING_ID=$REACT_APP_GA4_TRACKING_ID
```

Weren't necessary. For some reason, having them worked well with Webpack. But with Vite, keeping them there was causing our value from the `.env` file being override with an empty string, breaking the app. Removing them works correctly in the deployed environments (at least from the PR envs).

src/AddItemButton/index.js
luixlive: Yeah, agreed. Found a cleaner fix, just import the default from the CSS Module. That one includes all the classes even if they have a hyphen @JoshuaVulcan

src/utils/event-schemas.js
JoshuaVulcan: > we can’t rely on every error object having a string property anymore; in some cases (e.g. root-level or certain schema constructs) it may be missing or in a different shape.
> so those errors are handled safely (e.g. no path segments / no field match) instead of causing a crash.

Just swallowing errors isn't much of an improvement on preventing a crash TBH.
If we're saying there are real error paths with this structure, can we surface them appropriately?

luixlive: Yeah agreed. The thing is that even myself don't fully get this piece of code, but I agree that if there's an error, it would be better to see the app breaking than just ignoring it. I'll remove these lines with the theory that this originated from either hallucination of wrong inference from part of the agent, since honestly I don't see any changes on the errors mentioned on the official upgrade guide from `rjsf`. No reason to change this code 👍

luixlive: I removed this code, and immediately got tests failing. And I kind of understood better what is going on. If I remove these lines, the tests fail because we assume that all errors come with a `property` property (literally, `error.property`). But we are getting an error that doesn't have it, which is this one:
```
errors [
  {
    stack: 'schema is invalid: data/required must NOT have fewer than 1 items'
  }
]
```

That originates from schemas with a `required` property somewhere with an empty array `[]`. And we have many fixtures like that. I investigated where this error comes from, and it is actually part of the meta-schema draft-04 that we use for `rjsf`:
```
{
  "id": "http://json-schema.org/draft-04/schema#",
  "$schema": "http://json-schema.org/draft-04/schema#",
  "description": "Core schema meta-schema",
  "definitions": {
    ...
    "stringArray": {
      "type": "array",
      "items": {"type": "string"},
      "minItems": 1,
      "uniqueItems": true
    }
  },
  "type": "object",
    ...
    "required": {"$ref": "#/definitions/stringArray"},
    ...
  },
  ...
}
```

You can see that the `stringArray` definition that is used for the `required` property expects `"minItems": 1`. What I really don't understand is why we weren't facing this issue before 😆 But I'm pretty sure it's related to the library update.

In other words, we were already "swallowing" this error before, just not consciously. So what are the next steps? We sure can't update all the schemas out there with empty required arrays. We could also create some sort of utility to iterate recursively schemas and remove `required` properties with empty arrays. Or create or own meta-schema. From my perspective, these all sound like a very big effort for something that has been working like this since forever, and that we are about to deprecate with the v2 migraitons!

I suggest to keep the defensive code to avoid test failures and runtime errors when loading schemas with empty arrays. It seems harmless, but again, we can see how this goes when running regression tests.

What do you think @JoshuaVulcan ?

luixlive: Ok I was able to prompt Cursor in `Ask` mode to run a full investigation on `rjsf` version 6 and `@rjsf/validator-ajv8` to understand why we are seeing this meta-schema error that we weren't seeing before with `rjsf` v5 and `@rjsf/validator-ajv6`. And I got the response. 

TLDR: `@rjsf/validator-ajv8` is more strict and runs the meta-schema validation, and the new defensive code is one valid way of dealing with the issue.

Here the complete response:

## 1. Why you see “schema is invalid: data/required must NOT have fewer than 1 items”

**Draft-04 says `required` cannot be an empty array.**

In the draft-04 meta-schema your project uses:

```20:25:node_modules/ajv-draft-04/dist/refs/json-schema-draft-04.json
    "stringArray": {
      "type": "array",
      "items": {"type": "string"},
      "minItems": 1,
      "uniqueItems": true
    }
```

and `required` is defined as:

```84:84:node_modules/ajv-draft-04/dist/refs/json-schema-draft-04.json
    "required": {"$ref": "#/definitions/stringArray"},
```

So whenever `required` is present, it must be an array of strings with **at least one item**. An empty array `required: []` is invalid.

Your schemas do use `required: []` in several places (e.g. in `src/__test-helpers/fixtures/event-schemas.js` at lines 329, 511, 583). The error is about that: “must NOT have fewer than 1 items” = the array has 0 items. So the problem is **empty `required` arrays**, not empty strings inside `required`.

---

## 2. Why this started after moving to RJSF v6 / `@rjsf/validator-ajv8`

**Ajv 8 validates the schema against the meta-schema when you compile.**

When the form runs, RJSF calls the validator, which ends up doing something like `this.ajv.compile(schema)`. In Ajv 8 (and with `ajv-draft-04`), `compile()` validates the schema against the draft-04 meta-schema by default. If the schema is invalid (e.g. `required: []`), **Ajv throws** instead of compiling.

So:

- **Before (e.g. `@rjsf/validator-ajv6` + Ajv 6):** Schema validation on compile was either off or looser, so `required: []` was accepted and you only got *instance* validation errors (with normal `property` paths).
- **After (`@rjsf/validator-ajv8` + Ajv 8 / `ajv-draft-04`):** The schema is validated on compile; `required: []` fails meta-schema validation, so `compile()` throws. That throw is a **schema** error, not a **form data** error.

So the new error comes from the **validator upgrade** (RJSF v6 + `@rjsf/validator-ajv8` and the way Ajv 8 compiles schemas), not from RJSF’s UI layer.

---

## 3. Why that error has no `property`

**Schema compilation errors are turned into a minimal error object that only has `stack`.**

Flow:

1. In `validator.js`, `rawValidation()` catches the exception from `this.ajv.compile(schema)` and returns:
   - `errors: undefined`
   - `validationError: <thrown Error>`
2. In `processRawValidationErrors.js`, that is handled as a schema error:

```118:121:node_modules/@rjsf/validator-ajv8/lib/processRawValidationErrors.js
    if (invalidSchemaError) {
        errors = [...errors, { stack: invalidSchemaError.message }];
    }
```

So the schema error is pushed as **only** `{ stack: invalidSchemaError.message }`. There is no `property`, `instancePath`, or other field. Your code that expects `error.property` to be a string is correct to guard with “defined and string” so it doesn’t break on this schema error.

---

## Summary

| What | Explanation |
|------|-------------|
| **Trigger** | Schemas use `required: []`. Draft-04 says `required` must have `minItems: 1`. |
| **Why now** | Ajv 8 (used by `@rjsf/validator-ajv8` and `ajv-draft-04`) validates the schema on `compile()` and throws for invalid schemas; the old Ajv 6–based validator did not (or was less strict). |
| **Why no `property`** | The thrown error is turned into `{ stack: message }` only; RJSF does not add a `property` for schema compilation errors. |

So:

- The defensive check in `getLinearErrorPropTree` (only use `errorProperty` when it’s a non-null string) is the right way to handle these schema-level errors.
- To fix the underlying issue you can either:
  - Stop sending `required` when there are no required fields (omit the key), or
  - Normalize schemas so you never pass `required: []` (e.g. replace `required: []` with omitting `required` before passing the schema to the validator).

If you want, we can next trace exactly where your app builds or receives those schemas so you can add that normalization in one place.

PR #1477, opened 2026-03-05
ERA-12491: das-web-react: Render Link Fields in EarthRanger Client Forms

luixlive:
### What does this PR do?
Add support for text field format validations in v2 schemas

### How does it look
Validations
image

Read-only with link button:
image

### Relevant link(s)
* [ERA-12491](https://allenai.atlassian.net/browse/ERA-12491)
* [Env](https://era-12491.pamdas.org)

[ERA-12491]: https://allenai.atlassian.net/browse/ERA-12491?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/common/images/icons/arrow-up-right-from-square.svg
luixlive, line 1: New icon for the URL Link:
Image

src/DateTimePicker/styles.module.scss
```diff
@@ -7,8 +7,6 @@
 
   &[aria-invalid="true"] {
     .datePicker {
-      border-color: colors.$bright-red;
```
luixlive, line 10: I noticed an issue with some of our custom inputs. When they are in an error state, and we focus them, the error state border takes precendence, making a weir two-color border:
Image

I fixed it:
Image

src/GpsInput/index.js
```diff
@@ -230,7 +230,7 @@ const GpsInput = ({
         aria-describedby={isInputValid && description ? descriptionId : undefined}
         aria-errormessage={!isInputValid ? errorMessageId : undefined}
         aria-expanded={isTextSearchOptionChecked ? placesFromSearchText?.length > 0 : undefined}
-        aria-invalid={!isInputValid}
+        aria-invalid={isInputValid ? 'false' : 'true'}
```
luixlive, line 233: Claude noticed an issue I've been making in several places in our codebase. `aria-invalid` should have a string as value, not a boolean. It was an accessibility issue unnoticeable in the UI, but it's nice to fix it. You will notice several places where I do this in the PR.

src/GpsInput/styles.module.scss
```diff
@@ -13,6 +13,14 @@
     padding: 0.25rem 0.5rem;
     width: 100%;
 
+    &[aria-invalid="true"] {
```
luixlive, line 16: Similar fix for the error / focus border issue:
Image

Fixed:
Image

src/ReportManager/DetailsSection/SchemaForm/fields/Boolean/index.js
```diff
@@ -7,15 +7,18 @@ import * as styles from './styles.module.scss';
 const Boolean = ({ details, error, id, onFieldChange, readOnly, value = false }) => {
   const hasError = !!error;
   const hasDescription = !!details.description && !hasError;
-  const label = details.isRequired ? `${details.label} *` : details.label;
```
luixlive, line 10: Another accessibility improvement suggested by Claude: remove the asterisk from the screen reader  readable text. It's role is of an adornment, not real text.

src/ReportManager/DetailsSection/SchemaForm/fields/Boolean/index.js
```diff
@@ -24,13 +27,13 @@ const Boolean = ({ details, error, id, onFieldChange, readOnly, value = false })
       readOnly={readOnly}
     />
 
-    {(hasDescription || hasError) && <p
-      aria-live={hasError ? 'assertive' : 'off'}
+    <p
```
luixlive, line 29: Also: not mount and unmount an "aria-live" region. So now the paragraph stays there, and doesn't take any space while it's empty. This way when there's an error, screen readers get notified through the "assertive" algorithm successfully.

You will see this and the previous change repeated in a bunch of files.

src/ReportManager/DetailsSection/SchemaForm/fields/ChoiceList/Dropdown.js
```diff
@@ -8,6 +8,13 @@ import * as styles from './styles.module.scss';
 
 const IndicatorSeparator = () => null;
 
+const Input = ({ selectProps, ...otherProps }) => <components.Input
```
luixlive, line 11: I noticed our choice list dropdowns had several aria attributes broken. React-select doesn't support properly a few of them, but there are ways around the issue like the one I implemented here (taking care of injecting them manually).

src/ReportManager/DetailsSection/SchemaForm/fields/Text/index.js
```diff
@@ -1,52 +1,80 @@
-import React, { memo } from 'react';
```
luixlive, line 1: After so many files of old code improvements, here are the changes actually related to the PR ticket 😅 

The text field is now more complex than before. Since we want a button inside the input (for URLs), and I don't want to do some nasty "position: absolute" stuff, I decided to create a custom input wrapper (a div) that behaves as the input itself (focus, hover, border, background...) while wrapping the input and the button. This file is now aware of the "format validation" of text fields and if the field has the URL format, is read only, and the value is a valid URL, we render the button for both, short and long inputs:
Image

The input and the link are semantically correct and accessible (keyboard navigation works well).

src/ReportManager/DetailsSection/SchemaForm/utils/useSchemaValidations/index.js
```diff
@@ -3,6 +3,8 @@ import addFormats from 'ajv-formats';
 import Ajv2020 from 'ajv/dist/2020';
 import { useTranslation } from 'react-i18next';
 
+import { TEXT_ELEMENT_ALPHANUMERIC_FORMAT_VALIDATION_PATTERN } from '../../../../../utils/v2-event-schemas/constants';
```
luixlive, line 6: This is the portion of the work related to validation. This hook, `useSchemaValidations`, is responsible of running AJV against the form data and parse the errors in an easy to read format for our code. I had to include the new validation errors:
- email format
- uri format
- uuid format
- alphanumeric pattern

This, plus the proper translations, result in:
Image

src/utils/v2-event-schemas/transformSchemaToFormElements/transformField/transformTextField/index.js
```diff
@@ -1,4 +1,14 @@
-import { TEXT_ELEMENT_INPUT_TYPES } from '../../../constants';
+import {
```
luixlive, line 1: And finally, the transformation. Now `transformSchemaToFormElements` handles the new format validations properly when transforming the v2 schema into the `formElements` object. But you already reviewed this code since this method also exists in the `report-form-builder`. It's almost a copy paste.

src/utils/string.js
```diff
@@ -4,6 +4,15 @@ export const fileNameFromPath = path => path.match(/[^\\/]+$/)[0];
 export const uuid = () => uuidv4();
 export const caseInsensitiveCompare = (str1, str2) => str1.toLowerCase() === str2.toLowerCase();
 
+export const getIsValidWebUrl = (string) => {
```
luixlive, line 7: New trivial utility to validate if a string is a web URL.

src/ReportManager/DetailsSection/SchemaForm/fields/Text/index.js
```diff
@@ -1,52 +1,80 @@
...
+        aria-describedby={hasDescription ? `${id}-description`: undefined}
+        aria-errormessage={hasError ? `${id}-description` : undefined}
+        aria-invalid={hasError ? 'true' : 'false'}
+        aria-required={details.isRequired}
+        id={id}
+        onChange={(event) => onFieldChange(id, event.currentTarget.value || undefined)}
+        placeholder={details.hint}
+        readOnly={readOnly}
+        ref={inputRef}
+        value={value}
```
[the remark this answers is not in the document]
luixlive: In React 19, `forwardRef` is no longer necessary: https://react.dev/reference/react/forwardRef

A `ref` can be passed as a normal prop.

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/index.js
```diff
@@ -217,14 +217,13 @@ const Collection = ({
...
-        id={`${id}-description`}
+    <p
+      aria-live="assertive"
+      className={`${styles.description} ${hasError ? styles.error : ''}`}
+      id={`${id}-description`}
     >
       {error?.message || details.description}
-    </p>}
-
+    </p>
```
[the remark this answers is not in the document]
luixlive: Ok, fixed. We were incorrectly adding one live region per field. Now this paragraph stays just as a description / error message correctly attached to the field through ARIA attributes, and I created a single form-level live section to announce form errors.

src/ReportManager/DetailsSection/SchemaForm/index.js
```diff
@@ -216,6 +224,18 @@ const SchemaForm = ({
   }, [formData, formElements, setLocationMarkers]);
 
   return <form onSubmit={onSubmit}>
+    <div className="sr-only" role="alert">
```
luixlive, line 227: After researching a little and some help from Claude, I decided to not add one live region per field, which could be annoying when there are several errors in the form. Now those description / errors are simple paragraphs connected to their field through ARIA attributes. And now we have this new form-level live region that announces form errors to screen readers through a `role="alert"`, with "sr-only" so it's invisible.

JoshuaVulcan (approved):
Functionally approved w/local smoke-testing. Let's see how this behaves in dev 👁️

PR #1495, opened 2026-03-25
ERA-12914: The headers event detail view doesn't honor priority

luixlive:
### What does this PR do?
Fix issue with event headers not showing the color of their priority.

Extra: add Cursor Rule files.

### How does it look
image

### Relevant link(s)
* [ERA-12914](https://allenai.atlassian.net/browse/ERA-12914)
* [Env](https://era-12914.dev.pamdas.org/)

[ERA-12914]: https://allenai.atlassian.net/browse/ERA-12914?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.cursor/rules/business.mdc
+++ b/.cursor/rules/business.mdc
@@ -1,0 +1,90 @@
+---
+alwaysApply: true
+---
+
+# Business Context
+
+**Organization:** EarthRanger  
+**Domain:** Wildlife Conservation Technology
+
+EarthRanger is a real-time conservation platform that unifies field data, tracked assets, and people on a shared map. Core workflows: recording field events, monitoring wildlife and vehicles, planning and documenting patrols, and layering reference geography for spatial decision-making.
+
+**DAS** (Domain Awareness System) is the legacy product name. EarthRanger is the official name today. The acronym remains visible in API paths, host URLs, and related identifiers — treat them as synonymous.
+
+**Events** and **reports** refer to the same domain object. **Events** is the preferred term.
+
+## Events
+
+Events are time-stamped operational records tied to a geographic point or polygon. They represent field situations: wildlife sightings, human–wildlife conflict, enforcement actions, community reports, infrastructure issues, or any workflow the tenant configures. Each event has **state** (e.g. active / resolved), **priority** (red / amber / green), and optionally a **reporter**.
+
+**Event categories** are a server-defined taxonomy grouping **event types**. Each event type defines presentation (display name, icon), defaults (state, priority), and the structure of captured data. Example: the category **Security** may contain types **Animal Control**, **Carcass**, and **Spoor** — actual names are deployment-specific.
+
+**Incident collections** are a special event class: a parent record (`incident_collection`) that groups multiple child events through a `contains` relationship, keeping one field response under a single umbrella (e.g. a snare, carcass, and arrest linked to the same incident). The collection aggregates display metadata such as priority.
+
+Every event type has a **form schema** expressed as JSON that defines fields, layout, and validation.
+
+Operators discover events through the **Events sidebar** (text search, filters, sort, date range). The map renders markers, clusters, optional heatmaps, and drawn polygon geometry.
+
+## Subjects, Observations, and Tracks
+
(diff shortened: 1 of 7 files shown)


.cursor/rules/business.mdc
luixlive, line 1: Not related to this ticket, but totally harmless: I added Cursor rule files to provide agents minimal context about the repository in each prompt.
- `business.mdc`: Always-applied domain guide for EarthRanger: product vocabulary (events, subjects, patrols, map, alerts, settings, etc.) so assistants align with how conservation operators use the app. Token count **1,425**.
- `development.mdc`: Always-applied engineering guide: stack (React, Redux, Vite, Mapbox, etc.), folder layout, routing/Redux/API patterns, and common yarn commands. Token count **1,106**.
- `testing.mdc`: Applied to `**/*.test.*` Jest + Testing Library + MSW conventions, where test config lives, and how to structure and write accessible component/hook tests. Token count **564**.

Loading all of them takes around 3,000 tokens. Claude 3 Haiku has a context window limit of 200,000 tokens, and frontier models up to 1,000,000. So the token consumption is minimal and the context provided is of substantial value for agents to take better decisions.

.github/copilot-instructions.md
```diff
@@ -1,108 +1,30 @@
...
-- **Async operations**: Use `waitFor()` for async state updates
-- **Socket tests**: Use `MockSocketContext` for WebSocket simulation
-
-## Error Handling & Performance
-- **Error boundaries**: Wrap route-level components with ErrorBoundary
-- **Request lifecycle**: Ensure requests resolve properly, never leave hanging/obfuscated state
-- **Memory management**: Watch for memory leaks in map layers and event handlers
-- **Code style**: Extensive ESLint and stylelint rules configured - follow existing patterns
\ No newline at end of file
+## PR Review Guidelines
```
luixlive, line 1: All the content of `copilot-instructions` is already in the rule files, so this can be simplifier to point Copilot to read them and just provide basic instructions on good PR practices since the main role of Copilot for us is PR reviewer.

.github/pull_request_template.md
```diff
@@ -1,16 +1,11 @@
-### What does this PR do?
-- Be contextual and descriptive: i.e. What was changed or added.
+## What does this PR do?
```
luixlive, line 1: Trivial update in our PR template to match the EFE. It's basically the same but a bit simplified.

jest-config/mocks/scss.js
```diff
@@ -1,6 +1,17 @@
-module.exports = new Proxy(
+const proxy = new Proxy(
```
luixlive, line 1: The fix to the bug was to change the way we import the CSS Module since Vite transformer for SASS modules is different than Webpack's. However, in Jest we were using a mock for CSS modules that simply works as an identity object proxy, but didn't handle the `default` key scenario. This file updates our mock so Jest tests pass when doing the import as in this PR.

src/ReportManager/Header/index.js
```diff
@@ -10,7 +10,7 @@ import EventIcon from '../../EventIcon';
 import LocationJumpButton from '../../LocationJumpButton';
 import ReportMenu from './ReportMenu';
 
-import * as styles from './styles.module.scss';
+import styles from './styles.module.scss';
```
luixlive, line 13: The fix. The root cause is that for some reason, SASS transformer in Vite doesn't play well with classes that have a "-" in the middle. For example, here we have `priority-100`. They don't include it in the functions when importing like:
```Javascript
```

But when importing the default, we have the entire object, so the fix is to replace that with:
```Javascript
```

This means that it would be better to replace every single CSS module import in the repository to the second example, since it's more reliable. But that is way out of the scope of a hotfix, and not really necessary. We can use the new import as needed.

JoshuaVulcan (approved):
Looks good, I'm surprised static analysis and/or testing wouldn't have caught it sooner honestly 😅

PR #1498, opened 2026-03-25
ERA-12922: Enums in Checkboxes fields render values instead of display

luixlive:


.cursor/rules/testing.mdc
```diff
@@ -32,8 +32,8 @@ alwaysApply: false
 
 ### Testing
 
-- Use `render` from `app/test-utils.jsx` for components, with a custom `renderComponent` helper.
-- Use `renderHook` from `app/test-utils.jsx` for hooks.
+- Use `render` from `src/test-utils.jsx` for components, with a custom `renderComponent` helper.
```
luixlive, line 35: Unrelated typo fix.

src/ducks/event-schemas/sanitizeSchemas.js
```diff
@@ -122,7 +122,9 @@ const generateSchemaAndUiSchemaForCheckbox = (definition, schema) => {
       uniqueItems: true,
     },
     uiSchemaEntry: {
-      'ui:enumNames': titleMap.map((item) => item.name),
+      items: {
```
luixlive, line 125: Here's the fix. This is not well documented, but it seems like the `uiSchema` for a checkbox widget requires the `ui:enumNames` wrapped by an `items` object. I have the theory that this only applies for the bootstrap version of the library (the one we use), because if you use the [playground](https://rjsf-team.github.io/react-jsonschema-form/) and provide the schemas without this, the checkbox list is rendered properly:
Image

See how the `uiSchema` doesn't wrap the `ui:enumNames` with the items object, and they still render properly. If you wrap them with the items object, the checkbox list still renders ok, so my guess is that it should be optional. But for us, we need to add it to make it work.

I initially thought it was an issue in our custom CheckboxWidget, but that's not the case since the issue is there even if we don't use a custom widget.

package.json
```diff
@@ -5,7 +5,7 @@
   "buildbranch": "generated_by_build",
   "buildnum": "also_generated_by_build",
   "dependencies": {
-    "@auth0/auth0-react": "^2.15.1",
+    "@auth0/auth0-react": "^2.16.0",
```
luixlive, line 8: Since I couldn't find the issue, in a desperate try to blame an outdated dependency, I updated all the minor and patch versions haha.

src/ducks/event-schemas/sanitizeSchemas.js
```diff
@@ -122,7 +122,9 @@ const generateSchemaAndUiSchemaForCheckbox = (definition, schema) => {
...
+        'ui:enumNames': titleMap.map((item) => item.name),
+      },
       'ui:widget': CheckboxesWidget,
     },
```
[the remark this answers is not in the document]
luixlive: Great feedback. I'll say that it's out of a hotfix scope given we don't have any coverage on these files. It would be awesome tho, we could at least vibe code some. But again, we are moving away from v1s so the effort doesn't seem worth it.

package.json
```diff
@@ -50,8 +50,8 @@
     "react-error-boundary": "^6.1.1",
     "react-fast-compare": "^3.2.2",
     "react-flip-toolkit": "^7.2.4",
-    "react-ga4": "^2.1.0",
-    "react-i18next": "^16.6.2",
+    "react-ga4": "^3.0.1",
+    "react-i18next": "^16.6.6",
     "react-infinite-scroller": "^1.2.6",
```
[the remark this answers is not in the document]
luixlive: I didn't add it in the PR description but left a comment there. `react-ga4` hadn't received an update in 3 years till yesterday! No kidding, I was planning on moving away from the library and use a vanilla configuration. But with this version change they moved to ES modules so we can stay a bit longer I guess.

JoshuaVulcan (approved):
Good catch.

PR #1503, opened 2026-03-26
Hotfix: Fix timezone for time fields in v2 forms

luixlive:


src/ReportManager/DetailsSection/SchemaForm/fields/DateTime/index.js
```diff
@@ -67,37 +67,13 @@ const INPUTS = {
...
-
   const Input = INPUTS[details.inputType];
 
   const hasError = !!error;
   const hasDescription = !!details.description && !hasError;
   const label = details.isRequired ? `${details.label} *` : details.label;
 
-  // Date-time and time input types have a timezone offset, so we correct the input value to the current user timezone
-  // before rendering it.
-  useEffect(() => {
```
luixlive, line 80: We used to fix the timezone offset at the field level, which was the cause of the bug. All the fields with times tried to update the form data in the same render cycle, and only one was winning the race.

src/ReportManager/DetailsSection/SchemaForm/utils/getFormDataWithFixedTimezones/index.js
```diff
@@ -0,0 +1,60 @@
+import { format, isValid, parseISO } from 'date-fns';
+
+import {
+  DATE_TIME_ELEMENT_INPUT_TYPES,
+  FORM_ELEMENT_TYPES,
+} from '../../../../../utils/v2-event-schemas/constants';
+
+const correctDateTimeValue = (value, inputType) => {
```
luixlive, line 8: New utility to fix the timezone of the form data recursively once and for all.

src/ReportManager/DetailsSection/SchemaForm/index.js
```diff
@@ -165,25 +167,33 @@ const SchemaForm = ({
...
-      // a new event and the initial form data hasn't been set. Set the initial
-      // form data from the visible fields.
-      const visibleFieldIds = visibleSectionIds.flatMap((sectionId) => [
-        ...formElements[sectionId].details.leftColumn,
-        ...formElements[sectionId].details.rightColumn,
-      ]);
-      const defaultFormData = getDefaultFormData(visibleFieldIds, formElements);
-      if (Object.keys(defaultFormData).length > 0) {
-        onFormDataChange(defaultFormData);
+    if (shouldCalculateInitialData) {
```
luixlive, line 174: This useEffect used to calculate the initial data just for new events (basically, set the default inputs from the schema as the values), but now it also does it for existing events, by fixing the timezone offset of the time fields. It fixes all the time field values once.

JoshuaVulcan (approved):

PR #1504, opened 2026-03-26
Hotfix: Fix timezone for time fields in v2 forms

luixlive:


src/UserMenu/index.test.js
```diff
@@ -8,6 +8,7 @@ import UserMenu from './';
 describe('UserMenu', () => {
   let onLogOutClickMock;
   let onProfileClickMock;
+  let osanoBackup;
```
luixlive, line 11: Same exact PR than https://github.com/PADAS/das-web-react/pull/1503 but pointing to `release-2.136.1` and avoid a ton of conflicts. The only exception is this file that wasn't modified in the other PR. After moving to Osano cookie manager, a few tests were broken. I fixed them here.

JoshuaVulcan (approved):

PR #1509, opened 2026-03-27
[ERA-12587] [wip] Buoy Gear UI Updates

JoshuaVulcan:
## What does this PR do?
- Removes Buoy treatment for subjects and subject tracks
- Adds a separate Buoy component stack for displaying ropeless fishing gear

### Relevant link(s)
- https://allenai.atlassian.net/browse/ERA-12587
- https://era-12587.dev.pamdas.org/

## Notes
Acceptance-testing-wise, let's withhold performance judgement until [this PR](https://github.com/PADAS/das/pull/3816) is merged and available for us to test with.


.env.development
```diff
@@ -1,4 +1,4 @@
-REACT_APP_DAS_HOST=''
+REACT_APP_DAS_HOST='https://buoy.pamdas.org'
```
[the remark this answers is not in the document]
JoshuaVulcan: this is specifically for our feature environment and will be reverted before merge 👍

src/App.js
```diff
@@ -164,6 +165,15 @@ export const App = () => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
+  useEffect(() => {
+    dispatch(fetchAllGear());
+    const intervalId = window.setInterval(() => {
+      dispatch(fetchAllGear());
+    }, 600000);
```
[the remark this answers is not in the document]
JoshuaVulcan: On a ten minute interval? Literally never

public/locales/en-US/components.json
```diff
@@ -213,6 +213,20 @@
     "closeMapLayersButtonLabel": "Close Map Layers",
     "closePatrolFeedButtonLabel": "Close Patrol Feed",
     "closeSettingsButtonLabel": "Close Settings",
+    "closeGearTabButtonLabel": "Close Gear",
```
luixlive: nit: dumb maybe but it helps me rapidly scan the i18n files to sort the keys alphabetically. They almost are! This is one of the things I love of agents, sorting the keys in all the i18n files modified in this PR is a 2 line prompt away haha.

src/utils/gear.js
```diff
@@ -0,0 +1,264 @@
...
+export const GEAR_POINT_ROLE_SINGLE = 'single';
+export const GEAR_POINT_ROLE_TRAWL_END = 'trawl_end';
+
+/**
+ * Normalize one GET /api/v1.0/gear response body. Deployments differ:
+ * - DRF pagination: { count, next, previous, results }
+ * - Bare array (or copied payload): [...]
+ * - v1 envelope used elsewhere: { data: { results, next } } or { data: [...] }
+ */
+export const parseGearListPagePayload = (data) => {
```
luixlive: The code being so defensive and ignorant of the response format (even the comment above the method is explicit about it) is like we don't know or own API. We are always calling the same endpoint, with the same format and parameters, why do we need this utility at all? We should trust the contract.

src/ducks/gear/index.js
```diff
@@ -0,0 +1,195 @@
...
+        dispatch({ type: FETCH_GEAR_APPEND_PAGE, payload: rows });
+      }
+      hasNext = hasNextPage;
+      page += 1;
+    }
+
+    dispatch(fetchGearSuccess(merged));
+    return merged;
+  } catch (error) {
+    dispatch(fetchGearError(error));
```
luixlive: I already added a similar comment somewhere else, but why does this code read the response like if we don't know our API? It's like: _if the error response is not here, maybe it's here, or there, and make sure it's a string, otherwise maybe an object with the message in this property, if not add this default..._ hahaha. Don't we know the contract of this endpoint? This makes the code hard to follow, to test, and eventually, to debug.

src/ducks/gear/index.js
```diff
@@ -0,0 +1,195 @@
...
+  allIds: [],
+  byId: {},
+  error: null,
+  hiddenGearIds: [],
+  initialLoadInProgress: false,
+  loading: false,
+  tabEligible: false,
+};
+
+const mergeGearResults = (results) => {
```
luixlive: I see we defined the utilities in `'../../utils/gear'` but this one and `mergeGearAppendRows` are hanging here. Also I'd suggest to make these utilities agnostic of their implementations. The name "results" doesn't tell explicitly what it contains, it's just bleeding from the fact that the parent did a fetch request.

src/GearLayer/index.js
```diff
@@ -0,0 +1,87 @@
+import React, { memo, useContext, useMemo } from 'react';
+import { useSelector } from 'react-redux';
+
+import { MapContext } from '../App';
+import { LAYER_IDS, SOURCE_IDS } from '../constants';
+import { useMapEventBinding } from '../hooks';
+import useMapLayers from '../hooks/useMapLayers';
```
luixlive: I've been moving away from these two hooks because they don't handle addition and removal of sources and layers correctly. Sometimes they even throw warnings and errors in the console for this same reason.

You can see an example with the Coordinate Reference System box layer of how to properly add and remove sources and layers: https://github.com/PADAS/das-web-react/blob/develop/src/Map/layers/useCrsBoundingBoxLayer/index.js#L93 It's just a couple lines of code in a single `useEffect`, while `useMapLayers` and `useMapSources` attach like 10 `useEffect`s. They don't only cause memory leaks in the map, but they're also less performant.

src/GearPopup/index.js
```diff
@@ -0,0 +1,64 @@
...
+
+  const primaryLabel = gearHumanReadableLabel(gear);
+  const popupTitle = gear.manufacturer
+    ? `${gear.manufacturer}: ${primaryLabel}`
+    : primaryLabel;
+  const { coordinates } = data;
+
+  return <>
+    <div className={styles.header}>
+      <h6 className={styles.title} data-testid="gear-popup-title">{popupTitle}</h6>
```
luixlive: An `h6` feels random. We should maintain good semantics and add the proper heading level. For a dialog that steals focus, an `h2` is usually appropriate because it's not the page heading, but hierarchically it doesn't' have any headings above. It depends on the context of course.

src/GearPopup/index.js
```diff
@@ -0,0 +1,64 @@
...
+
+    <div className={styles.body}>
+      {coordinates && <GpsFormatToggle
+        className={styles.gpsFormatToggle}
+        lngLat={{ latitude: coordinates[1], longitude: coordinates[0] }}
+        name="gearPopup-gpsFormatToggle"
+      />}
+
+      <dl className={styles.details}>
+      <dt>{t('typeLabel')}</dt>
```
luixlive: Can we add the proper indentation here? <dl> is wrapping all the <dt> / <dd> pairs, but it doesn't look like it.

src/Map/index.js
```diff
@@ -125,6 +126,7 @@ const Map = ({ children, onMapLoad, socket }) => {
   const hiddenFeatureIDs = useSelector(state => state.data.mapLayerFilter.hiddenFeatureIDs);
   const eventFilter = useSelector(state => state.data.eventFilter);
   const eventsEnabled = useSelector((state) => state.view.systemConfig[SYSTEM_CONFIG_FLAGS.EVENTS]);
+  const gearTabEligible = useSelector((state) => state.data.gear.tabEligible);
```
luixlive: The name `tabEligible` assigns an intent to the variable. What it really is, is a boolean that is `true` when gear length is more than zero, so a more explicit and scalable name would be something like `hasGear`. That way the variable tells you what it really means and is more versatile for other purposes.

src/Map/index.js
```diff
@@ -275,6 +277,21 @@ const Map = ({ children, onMapLoad, socket }) => {
...
+  const onSelectGear = withLocationPickerState(
+    ({ coordinates: coordinatesOverride, event, layer }) => {
+      const { geometry, properties } = layer;
+      const coordinates = coordinatesOverride ?? (
+        geometry.type === 'Point'
+          ? geometry.coordinates
+          : [event.lngLat.lng, event.lngLat.lat]
+      );
+
+      window.setTimeout(() => showPopup('gear', { geometry, properties, coordinates }));
```
luixlive: The timeout doesn't have a time value attached. If this is a patch to a render problem, we could add a small comment.

src/selectors/subjects/index.js
```diff
@@ -29,10 +30,11 @@ export const selectMapSubjectsFeatureCollection = createSelector(
     // Calculate the subject features to show on the map.
     const features = [];
     mapSubjects.forEach((subjectId) => {
-      if (!hiddenSubjectIDsSet.has(subjectId) && subjectStore[subjectId]) {
+      const subjectRecord = subjectStore[subjectId];
+      if (!hiddenSubjectIDsSet.has(subjectId) && subjectRecord && !isGearSubjectSubtype(subjectRecord)) {
```
luixlive, line 37: We should update the comments here and in `selectHydratedSubjectGroupsWithLastPositionTime`. They document these complex selectors pretty well, but now they miss the fact that gear subjects should be ignored.

src/selectors/tracks/index.js
```diff
@@ -16,6 +17,7 @@ const selectSubjectTrackState = (state) => state.view.subjectTrackState;
 const selectTimeOfDayTimeZone = (state) => state.view.trackSettings.timeOfDayTimeZone;
 const selectTimeSliderState = (state) => state.view.timeSliderState;
 const selectTracks = (state) => state.data.tracks;
+const selectSubjectStore = (state) => state.data.subjectStore;
```
luixlive, line 20: Same in this file about adding small comments. Reading these selectors before was a nightmare, I've been refactoring them to make them readable, and to follow the proper practices for right memoization (we still have a few selectors causing memory leaks). Let's keep them nice!

src/SideBar/GearTab/index.js
```diff
@@ -0,0 +1,243 @@
...
+  const byId = useSelector((state) => state.data.gear.byId);
+  const fetchError = useSelector((state) => state.data.gear.error);
+  const hiddenGearIds = useSelector((state) => state.data.gear.hiddenGearIds);
+  const loading = useSelector((state) => state.data.gear.loading);
+  const initialLoadInProgress = useSelector((state) => state.data.gear.initialLoadInProgress);
+  const mapLayerFilter = useSelector((state) => state.data.mapLayerFilter);
+
+  const errorMessage = useMemo(() => {
+    if (!fetchError) return null;
+    const data = fetchError.response?.data;
```
luixlive: Fetch API technical details are bleeding here! We shouldn't have to read `fetch.response.data`. Redux thunk should be a wrapper around that, so our store is clean like: `state.data.gear.error`  -> "Error message", and our React component are completely agnostic of how the store was populated.

src/SideBar/SidebarListSortingControls/index.js
```diff
@@ -0,0 +1,192 @@
...
+import { ReactComponent as SortLinesIcon } from '../../common/images/icons/sort-lines.svg';
+
+import { MAP_LAYER_SORT_OPTIONS, MAP_LAYER_SORT_VALUES, SORT_DIRECTION } from '../../constants';
+import {
+  setMapLayersGrouped,
+  setMapLayersSortBy,
+  setMapLayersSortDirection,
+} from '../../ducks/map-layer-filter';
+
+import * as styles from '../MapLayersTab/Filters/styles.module.scss';
```
luixlive: We should bring the sidebar list sorting controls styles to its own CSS module. It seems pretty nasty to have them under a child component that doesn't use them anymore just to import them from here.

src/TracksLayer/track.js
```diff
@@ -61,7 +60,7 @@ const TrackLayer = ({
   lineLayout = {},
   linePaint = {},
   onPointClick,
-  onTrackLabelClick,
+  onTrackLabelClick: _onTrackLabelClick,
```
luixlive: We should remove this prop from here and from the parent instead of leaving dead code.

src/utils/gear.js
```diff
@@ -0,0 +1,264 @@
...
+      const bEmpty = b.manufacturerKey === '';
+      if (aEmpty !== bEmpty) return aEmpty ? 1 : -1;
+      return a.manufacturerKey.localeCompare(b.manufacturerKey, undefined, { sensitivity: 'base' });
+    });
+};
+
+/** Ms since epoch from `gear.last_updated`, or null if missing/invalid. */
+export const getGearLastUpdatedMs = (gear) => {
+  if (!gear?.last_updated) return null;
+  const ms = new Date(gear.last_updated).getTime();
```
luixlive: I see some poor variable names in this file: `ms`, `a`, `b`, `g`. We should be a bit more verbose so the methods are easy to read 👍

src/App.js
luixlive: Let's move this magic number to a constant 😄

luixlive (changes requested):
Very big PR! Great work honestly. A dedicated tab for Buoy users sounds amazing.

I provided a bunch of comments that I feel would improve the code and would take 1 minute to fix each. But something very important, I barely saw a bunch of unit tests just for a couple files. We have new component files, map callbacks, layers, ducks... That have no testing protection!

I know we want to merge this fast, so I won't get picky in my next review. But given the amount of code touched, I think should apply a bunch of these recommendations and add more tests before merging 👍

luixlive (commented):
Code-wise I'm ready to approve 💯  I just wanted to do some manual tests of the new tab, layers, etc... But seems like the backend env is down @JoshuaVulcan

JoshuaVulcan:
> Code-wise I'm ready to approve 💯 I just wanted to do some manual tests of the new tab, layers, etc... But seems like the backend env is down @JoshuaVulcan

@luixlive 

I had @victorlujanearthranger review this (he's our internal stakeholder and coordinator w/NOAA on this) -- he has access to production Buoy so he could log-in and validate. We're signed-off and good to go from an acceptance test and product owner perspective.

luixlive (approved):

PR #1516, opened 2026-03-30
ERA-12933: Add Community Input

jeslefcourt:
## Summary
- Adds a new `CommunityPage` component for Community Input functionality
- Updates `ReportManager` with location picker improvements and detail view refactoring
- Extends event schemas/types duck and event utilities to support community input

## Test plan
- [ ] Verify Community Input page renders correctly via computer and mobile devices.
- [ ] Test mobile usability for all input types.
- [ ] Confirm location picker works as expected in report forms
- [ ] Smoke test report creation, viewing and editing flows throughout typical web event entry processes


jeslefcourt:
@JoshuaVulcan We also need to prevent the heavy libraries (mapbox, turf) from loading.  The download size for this simple form is kinda ridiculous right now.  I had Claude do that, but I didn't like its solution.  Wanted to ask you what your preferred method is for this.

I also plan to update this to grab individual event type icons instead of the 15 MB icon sprite.

jeslefcourt:
@JoshuaVulcan Went ahead and made the change to avoid mapbox from being loaded, since that was also loading the sprite.  There's still an unnecessary 3 MB load of Turf, but fixing that touches a ton of code.

public/locales/en-US/components.json
```diff
@@ -14,6 +14,11 @@
     "defaultLabel": "Create Event or Patrol",
     "defaultTitle": "Create"
   },
+  "communityPage": {
```
luixlive, line 17: We are missing translations in every other language for all the new keys in this file (‎`public/locales/en-US/components.json`).

public/locales/en-US/details-view.json
```diff
@@ -36,13 +36,16 @@
     "cancelButton": "Cancel",
     "saveButton": "Save",
     "stateResolveButton": "Save and resolve",
-    "stateReopenButton": "Save and reopen"
+    "stateReopenButton": "Save and reopen",
+    "stateReviewButton": "Save and review",
```
luixlive, line 39: I don't see this translation being used anywhere in the app 🤔

public/locales/ne-NP/details-view.json
```diff
@@ -1,80 +1,83 @@
...
+    "collapseClosedButtonTitle": "बन्द गरिएको",
+    "collapseOpenButtonLabel": "बन्द गर्नुहोस्",
+    "collapseOpenButtonTitle": "बिस्तार गरिएको"
+  },
+  "footer": {
+    "cancelButton": "रद्द गर्नुहोस्",
+    "saveButton": "सुरक्षित गर्नुहोस्",
+    "stateResolveButton": "सुरक्षित गरि समाधान गर्नुहोस्",
+    "stateReopenButton": "सुरक्षित गरि फेरी खोल्नुहोस्",
+    "stateReviewButton": "Save and review",
```
luixlive: The new keys were added to the nepali translation files, but in english. None of them is translated, not only in this file but also in the others.

public/locales/pt/details-view.json
```diff
@@ -1,66 +1,69 @@
...
+    "downloadButtonTooltip": "Baixar",
+    "deleteButtonTooltip": "Apagar",
+    "fullViewButtonTooltip": "Visão geral"
+  },
+  "footer": {
+    "cancelButton": "Cancelar",
+    "saveButton": "Salvar",
+    "stateResolveButton": "Salvar e resolver",
+    "stateReopenButton": "Salvar e reabrir",
+    "stateReviewButton": "Save and review",
```
luixlive: Same thing for portuguese files. Not translated!

public/locales/sw/details-view.json
```diff
@@ -36,13 +36,16 @@
     "cancelButton": "Ghairi",
     "saveButton": "Hifadhi",
     "stateResolveButton": "Hifadhi na Tatua",
-    "stateReopenButton": "Hifadhi na Fungua tena"
+    "stateReopenButton": "Hifadhi na Fungua tena",
+    "stateReviewButton": "Save and review",
```
luixlive: And same thing for swahili. All texts are in english.

public/locales/en-US/reports.json
```diff
@@ -246,7 +252,9 @@
       "saveSplitButton": {
         "saveAndReopenItem": "Save and reopen",
         "saveAndResolveItem": "Save and resolve",
-        "title": "Save"
+        "title": "Save",
+        "saveAndReviewItem": "Save and review",
```
luixlive, line 256: Another translation that doesn't seem to be used anywhere.

src/AddItemButton/AddItemModal/TypesList/styles.module.scss
```diff
@@ -18,7 +18,9 @@
     display: flex;
     flex-wrap: wrap;
     line-height: normal;
+    list-style: none;
```
luixlive, line 21: nit: The two new styles in this class `list-style` and `padding` seem to be unnecessary. The list was already rendering with no list style and no padding. No harm in keeping them, but probably a flag that agents are doing weird non-requested changes.

src/CommunityPage/index.js
```diff
@@ -0,0 +1,144 @@
...
+
+const CommunityPage = () => {
+  const { t } = useTranslation('components', { keyPrefix: 'communityPage' });
+  const dispatch = useDispatch();
+  const navigate = useNavigate();
+  const { value, '*': eventTypeValue } = useParams();
+
+  const allEventTypes = useSelector((state) => state.data.eventTypes);
+  const eventsByCategory = useMemo(() => {
+    const creatableTypes = allEventTypes.filter((t) => !t.is_collection && !t.readonly);
```
luixlive: nit: Better to use an explicit name over a tiny one `t` --> `eventType`. And not only because code readability, but also this one is shadowing our i18n `t` function.

src/CommunityPage/index.js
```diff
@@ -0,0 +1,144 @@
...
+      .catch(() => { setIsUnauthorized(true); });
+    dispatch(fetchEventTypes({ community_input: value }, { skipAuth: true }))
+      .then(() => setIsLoading(false))
+      .catch(() => setIsLoading(false));
+    dispatch(fetchEventsSchema({ community_input: value }, { skipAuth: true }));
+  }, [dispatch, value]);
+
+  useEffect(() => {
+    if (isLoading || isUnauthorized || eventTypeValue) return;
+    const creatableTypes = allEventTypes.filter((t) => !t.is_collection && !t.readonly);
```
luixlive: nit: We do this filtering two times. Here and in `eventsByCategory`'s `useMemo`. We could create its own variable and reuse it:
```Javascript
const creatableEventTypes = useMemo(
  () => allEventTypes.filter((eventType) => !eventType.is_collection && !eventType.readonly),
  [allEventTypes]
);
```

src/CommunityPage/index.js
```diff
@@ -0,0 +1,144 @@
...
+    if (creatableTypes.length === 1) {
+      navigate(`/community/${value}/${creatableTypes[0].value}`, { replace: true });
+    }
+  }, [allEventTypes, eventTypeValue, isLoading, isUnauthorized, navigate, value]);
+
+  const schemaFetchExtraParams = useMemo(() => ({ community_input: value }), [value]);
+  const schemaFetchAxiosConfig = useMemo(() => ({ skipAuth: true }), []);
```
luixlive: nit: This useMemo is memoizing a constant, we can simply move it at the top of the file and save the memoization costs:
const SCHEMA_FETCH_AXIOS_CONFIG = { skipAuth: true };

src/CommunityPage/index.js
```diff
@@ -0,0 +1,144 @@
...
+    const creatableTypes = allEventTypes.filter((t) => !t.is_collection && !t.readonly);
+    return creatableTypes.length ? [{ value: 'all', display: '', types: creatableTypes }] : [];
+  }, [allEventTypes]);
+
+  const [communityName, setCommunityName] = useState('');
+  const [searchText, setSearchText] = useState('');
+  const [showSubmittedModal, setShowSubmittedModal] = useState(false);
+  const [isLoading, setIsLoading] = useState(true);
+  const [isUnauthorized, setIsUnauthorized] = useState(false);
+  const typesListRef = useRef(null);
```
luixlive: This ref is passed and assigned but never used.

src/CommunityPage/index.js
```diff
@@ -0,0 +1,144 @@
...
+  const schemaFetchAxiosConfig = useMemo(() => ({ skipAuth: true }), []);
+
+  const onClickType = useCallback((reportType) => {
+    navigate(`/community/${value}/${reportType.value}`);
+  }, [navigate, value]);
+
+  const onBack = useCallback(() => {
+    navigate(`/community/${value}`);
+    setShowSubmittedModal(true);
+    setTimeout(() => setShowSubmittedModal(false), 3000);
```
luixlive: If the user navigates to another page of the app during these 3 seconds, we'll get a state update on an unmounted component. We should store the timeout id and add a new `useEffect` that cleans the timeout when unmounting.
```Javascript
const timeoutRef = useRef(null);
...
const onBack = useCallback(() => {
  navigate(`/community/${value}`);
  setShowSubmittedModal(true);
  timeoutRef.current = setTimeout(() => setShowSubmittedModal(false), 3000);
}, [navigate, value]);
...
useEffect(() => () => clearTimeout(timeoutRef.current), []);
```

And a nit: Would be nice to replace the magic number (3000) with a constant "SUBMITTED_MODAL_TIMEOUT" or something similar.

src/CommunityPage/index.js
```diff
@@ -0,0 +1,144 @@
...
+  }, [navigate, value]);
+
+  return <SidebarScrollProvider>
```
luixlive, line 162: BTW, I see a lot of linter errors in this file.

src/CommunityPage/index.test.js
```diff
@@ -0,0 +1,384 @@
...
+    render(
+      <Provider store={buildStore(types)}>
+        <CommunityPage />
+      </Provider>
+    );
+
+  beforeEach(() => {
+    const { fetchCommunityInfo } = require('../ducks/community');
+    const { fetchEventTypes } = require('../ducks/event-types');
+    const { fetchEventsSchema } = require('../ducks/event-schemas');
```
luixlive: Requiring inside a `beforeEach` goes against our patterns. We import, mock the module, and then mock the implementation of the function:
```Javascript
import { fetchCommunityInfo } from '.../ducks/community';

jest.mock('../ducks/community', () => ({
  ...jest.requireActual('../ducks/community'),
  fetchCommunityInfo: jest.fn(),
}));

beforeEach(() => {
  fetchCommunityInfo.mockImplementation(() => Promise.resolve({ name: 'Test Community' }));
});
```

There are several examples of this pattern in our test files 😄

src/DasIcon/index.js
```diff
@@ -1,4 +1,5 @@
 import React, { memo } from 'react';
+import { API_URL, API_V2_URL } from '../constants';
```
luixlive: `API_URL` is being imported but not used.

src/DasIcon/index.js
```diff
@@ -24,6 +25,19 @@ export const calcIconUrl = (type, iconId) => {
 };
 
 const DasIcon = ({ type, iconId, color = 'gray', dispatch: _dispatch, className, ...rest }) => {
+  if (window.location.pathname.startsWith('/community') && type === 'events' && iconId) {
```
luixlive: Instead of reading `window.location`, we should use react-router's `const location = useLocation();`. `window.location` is a global object that is not synchronized to the React components' lifecycle, which could cause small bugs.

src/CommunityPage/index.js
```diff
@@ -0,0 +1,144 @@
...
+    <Modal centered show={showSubmittedModal} onHide={() => setShowSubmittedModal(false)}>
+      <Modal.Header closeButton />
+      <Modal.Body className={styles.submittedModalBody}>{t('formSubmitted')}</Modal.Body>
+    </Modal>
+
+    <div className={styles.pageContainer}>
```
luixlive, line 168: Here we have a ternary operator nested in another ternary operator that are hard to follow. It would be nicer to have a couple conditionals before the `return` like:
```Javascript
if (isUnauthorized) {
  return <div className={styles.pageContainer}>
    <EarthRangerLogo className={styles.earthRangerLogo} />

    <span className={styles.loadingError}>{t('invalidCommunityUrl')}</span>
  </div>;
}

if (isLoading) {
  return <div className={styles.pageContainer}>
    <MoonLoader color="#333" size={50} />
  </div>;
}

if (!eventTypeValue && eventsByCategory[0]?.types.length === 1) {
  return <div className={styles.pageContainer}>
    <EarthRangerLogo className={styles.earthRangerLogo} />
  </div>;
}

return ...
```

src/DetailViewComponents/ActivitySection/index.js
```diff
@@ -199,7 +200,7 @@ const ActivitySection = ({
   }, [notes, notesToAdd, onExpandCard]);
 
   return <div data-testid="detailView-activitySection" ref={ref}>
-    <div className={styles.sectionHeader}>
+    {!isCommunity && <div className={styles.sectionHeader}>
```
luixlive, line 203: Two things here:
1- Don't we want the header? It provides a nice visual separation between the form and the extra notes / attachments, and a couple helpful buttons.
Image

2- It seems like some styles were broken on the activity section. I can see a bullet as a list style in the notes (see the screenshot above).

src/ducks/event-schemas/index.js
```diff
@@ -23,28 +28,62 @@ export const FETCH_EVENT_TYPE_SCHEMA_V2_SUCCESS = 'FETCH_EVENT_TYPE_SCHEMA_V2_SU
 export const FETCH_EVENT_TYPE_SCHEMA_FAILURE = 'FETCH_EVENT_TYPE_SCHEMA_FAILURE';
 
 // Action creators
-export const fetchEventsSchema = () => async (dispatch) => {
-  const response = await axios.get(EVENTS_SCHEMA_API_URL);
-
+export const fetchEventsSchema = (params = {}, config = {}) => async (dispatch) => {
```
luixlive: We only call this action creator in two places in the code:
```Javascript
// src/App.js
dispatch(fetchEventsSchema());

// src/CommunityPage/index.js
dispatch(fetchEventsSchema({ community_input: value }, { skipAuth: true }));
```

We only have two modes for fetching schemas: normal "fetch all" and the new "fetch community". Supporting extra parameters and deep configuration is overengineered, we can simply send a parameter `community_input_value` and simplify the changes here:
```Javascript
export const fetchEventsSchema = (communityInputValue = null) => async (dispatch) => {
  const response = communityInputValue
    ? await axios.get(COMMUNITY_EVENTS_SCHEMA_API_URL(communityInputValue), { skipAuth: true  })
    : await axios.get(EVENTS_SCHEMA_API_URL);

  dispatch({ payload: response.data.data, type: FETCH_EVENTS_SCHEMA_SUCCESS });
};
```

src/ducks/event-schemas/index.js
```diff
@@ -23,28 +28,62 @@ export const FETCH_EVENT_TYPE_SCHEMA_V2_SUCCESS = 'FETCH_EVENT_TYPE_SCHEMA_V2_SU
...
+  const { community_input, ...restParams } = params;
+  const url = community_input ? COMMUNITY_EVENTS_SCHEMA_API_URL(community_input) : EVENTS_SCHEMA_API_URL;
+  const response = await axios.get(url, { params: restParams, ...config });
   dispatch({ payload: response.data.data, type: FETCH_EVENTS_SCHEMA_SUCCESS });
 };
 
-export const fetchEventTypeSchema = (eventTypeValue, eventId) => async (dispatch, getState) => {
+export const fetchEventTypeSchema = (eventTypeValue, eventId, extraParams = {}, axiosConfig = {}) => async (dispatch, getState) => {
```
luixlive: Same thing. Just pass the `communityInputValue` and add the proper axios configuration here, avoiding technical details bleeding upwards. Right now the parents need to inject stuff like `schemaFetchAxiosConfig` which is not part of their responsibility and is passed through props a couple levels down.

src/ducks/event-schemas/index.js
```diff
@@ -23,28 +28,62 @@ export const FETCH_EVENT_TYPE_SCHEMA_V2_SUCCESS = 'FETCH_EVENT_TYPE_SCHEMA_V2_SU
...
+          location: locationParam,
+          pre_render: true,
+          ...restExtraParams,
+        },
+        ...axiosConfig,
+      });
+
+      const rawSchema = response.data;
+
+      if (rawSchema?.schema) {
```
luixlive: My understanding is that community schemas will be v2. If we will only support v2s, then this condition is unnecessary, we can just add a ternary to set the right URL in the v2 condition block.

And even more than that, we could better wire the community schema fetching so we don't need a separate condition, we can simply leave the flow as it was and inside the v2 flow blocks just add a couple ternary operators to properly fetch the community schemas from their endpoint.

src/ducks/event-schemas/index.js
```diff
@@ -23,28 +28,62 @@ export const FETCH_EVENT_TYPE_SCHEMA_V2_SUCCESS = 'FETCH_EVENT_TYPE_SCHEMA_V2_SU
...
-export const fetchEventTypeSchema = (eventTypeValue, eventId) => async (dispatch, getState) => {
+export const fetchEventTypeSchema = (eventTypeValue, eventId, extraParams = {}, axiosConfig = {}) => async (dispatch, getState) => {
   dispatch({ type: FETCH_EVENT_TYPE_SCHEMA });
 
   const state = getState();
   const eventType = selectEventTypeByValue(state, eventTypeValue);
   const userLocationCoords = state.view.userLocation?.coords;
+  const { community_input, ...restExtraParams } = extraParams;
+
+  const locationParam = userLocationCoords
```
luixlive, line 45: nit: I guess the reason to extract the `locationParam` variable out here was to reuse it, but we aren't. Let's reuse this variable in the condition blocks where we fetch v1s and v2s (lines 81 and 105).

src/ducks/event-schemas/index.js
```diff
@@ -60,20 +99,35 @@ export const fetchEventTypeSchema = (eventTypeValue, eventId) => async (dispatch
...
       });
 
-      dispatch({
-        payload: { eventId, eventTypeValue, schema: response.data },
-        type: FETCH_EVENT_TYPE_SCHEMA_V2_SUCCESS,
-      });
+      const rawSchema = response.data;
+
+      if (rawSchema?.schema) {
+        // v2 endpoint returned v1-format schema
```
luixlive: This should never happen. Seems like a hallucination. We should revert this block.

src/ducks/event-types/index.js
```diff
@@ -9,21 +9,32 @@ const USE_EVENT_TYPES_V2_MOCK_API = import.meta.env.REACT_APP_MOCK_EVENTTYPES_V2
...
   `${USE_EVENT_TYPES_V2_MOCK_API ? '/api/v2.0/' : API_V2_URL}activity/eventtypes`;
+export const COMMUNITY_EVENT_TYPES_API_URL = (communityValue) =>
+  `${API_V2_URL}community/${communityValue}/eventtypes`;
 
 // Actions
 export const FETCH_EVENT_TYPES_SUCCESS = 'FETCH_EVENT_TYPES_SUCCESS';
 
 // Action creators
-export const fetchEventTypes = () => async (dispatch) => {
+export const fetchEventTypes = (params = {}, config = {}) => async (dispatch) => {
```
luixlive: Same thing I mentioned in the schemas duck, this change is overnegineered, we could make it simpler so we don't change our code so much by just passing a `communityInputValue` optional parameter and keep the parents agnostic of implementation details:
```Javascript
export const fetchEventTypes = (communityInputValue = null) => async (dispatch) => {
  if (communityInputValue) {
    const response = await axios.get(COMMUNITY_EVENT_TYPES_API_URL(communityInputValue), { skipAuth: true });

    const eventTypes = response.data.data.map((eventType) => ({ ...eventType, version: 2 }));

    dispatch({ payload: eventTypes, type: FETCH_EVENT_TYPES_SUCCESS });
  } else {
    const [eventTypesResponse, eventTypesV2Response] = await Promise.all([
      axios.get(EVENT_TYPES_API_URL),
      axios.get(EVENT_TYPES_V2_API_URL)
    ]);

    const eventTypes = [
      ...eventTypesResponse.data.data.map((eventType) => ({ ...eventType, version: 1 })),
      ...eventTypesV2Response.data.data.map((eventType) => ({ ...eventType, version: 2 })),
    ];

    dispatch({ payload: eventTypes, type: FETCH_EVENT_TYPES_SUCCESS });
  }
};
```

src/ducks/community.js
```diff
@@ -0,0 +1,9 @@
+import axios from 'axios';
+
+import { API_V2_URL } from '../constants';
+
+export const COMMUNITY_INPUT_API_URL = (communityValue) =>
+  `${API_V2_URL}community/${communityValue}`;
+
+export const fetchCommunityInfo = (communityValue, config = {}) => () =>
```
luixlive: This duck is missing its reducer. It's a fundamental part of the ducks architecture design. It makes sense storing the fetched community info in the Redux store so it's easily accessible from anywhere in the app. Right now we are fetching in from the `CommunityPage`, which means the data gets lost every time that component unmounts and that if we need to read the info from other place in the app it will either be impossible or require prop drilling.

Here's a trivial implementation:
```Javascript
import axios from 'axios';

import { API_V2_URL } from '../constants';
import globallyResettableReducer from '../../reducers/global-resettable';

export const COMMUNITY_API_URL = (communityValue) =>
  `${API_V2_URL}community/${communityValue}`;

// Actions
export const FETCH_COMMUNITY_SUCCESS = 'COMMUNITY.FETCH_COMMUNITY_SUCCESS';

// Action creators
export const fetchCommunity = (communityValue) => async (dispatch) => {
  const response = await axios.get(COMMUNITY_API_URL(communityValue), { skipAuth: true });

  dispatch({ payload: response.data.data, type: FETCH_COMMUNITY_SUCCESS });
};

// Reducer
export const INITIAL_STATE = { communityInfo: null };

const communityReducer = (state, action) => {
  switch (action.type) {
  case FETCH_COMMUNITY_SUCCESS:
    return action.payload;

  default:
    return state;
  }
};

export default globallyResettableReducer(communityReducer, INITIAL_STATE);
```

Then you just need to register the reducer in `src/reducers/index.js` and you will be able to select the data from anywhere in the app:
```Javascript
const community = useSelector((state) => state.data.community);
```

src/ducks/events.js
```diff
@@ -200,8 +207,9 @@ export const clearEventData = () => ({
   type: CLEAR_EVENT_DATA,
 });
 
-export const createEvent = event => (dispatch, getState) => {
-  const params = {};
+export const createEvent = (event, extraParams = {}) => (dispatch, getState) => {
```
luixlive: All the action creators changed in this file have the same pattern of receiving extra parameters configuration instead of simply the `communityValue`. We should follow an approach like the recommended one in the other ducks.

mt-nginx.conf
```diff
@@ -23,18 +23,7 @@ server {
 		try_files $uri =404;
 	}
 
-	# Serves config.js if present on disk, otherwise returns an empty 200.
-	# The app is expected to gracefully handle an empty response.
-	location = /config.js {
```
luixlive, line 28: Question: are these two blocks removed intentionally? They seem outside the scope of this PR.

.env.development
```diff
@@ -1,4 +1,4 @@
-REACT_APP_DAS_HOST=''
+REACT_APP_DAS_HOST='https://root.dev.pamdas.org'
```
luixlive: This change shouldn't be commited.

src/index.js
```diff
@@ -107,6 +108,16 @@ const RootApp = () => {
           </RequireAccessToken>}
         />
 
+        <Route
```
luixlive: `/community` is not a valid route, but we are rendering the `CommunityPage`. I think we should remove this block and let a simple 404 be shown. THat's the honest error, since users shouldn't be trying to access there.

src/App.js
```diff
@@ -44,7 +44,8 @@ import WithSocketContext, { SocketContext } from './withSocketConnection';
 import 'axios-progress-bar/dist/nprogress.css';
 import './App.scss';
 
-export const MapContext = createContext(null);
+import { MapContext } from './MapContext';
+export { MapContext };
```
luixlive: I see lots of files changed their imports to point directly to `/MapContext`. Is there a reason to still export it from here?

src/EditableItem/Footer.js
```diff
@@ -31,16 +27,45 @@ const Footer = ({
 }) => {
   const { t } = useTranslation('details-view', { keyPrefix: 'footer' });
   const { cancelTitle = t('cancelButton') } = restProps;
-  const stateButtonLabel = t(isActiveState ? 'stateResolveButton' : 'stateReopenButton');
+  const isInReview = data?.state === EVENT_FORM_STATES.REVIEW;
```
luixlive, line 31: `isActiveState` is an external prop, while `isInReview` is derived from data.state. Nothing prevents a caller from passing `isActiveState={true}` when the event is actually in review state, both the `isActiveState` branch and the `isInReview` branch would render, producing two "Save and resolve" items in the dropdown. Both should be derived from `data.state` the same way `isInReview` already is, making them mutually exclusive by construction and removing the external prop dependency.

src/EditableItem/Footer.js
```diff
@@ -31,16 +27,45 @@ const Footer = ({
...
 
   return <div className={styles.formButtons}>
     <Button type="button" onClick={onCancel} variant="secondary">{cancelTitle}</Button>
     {!readonly && <SaveButtonComponent className={styles.saveButton} disabled={saveDisabled} drop='down' variant='primary' type='submit' title={t('saveButton')} onClick={onSave}>
       {!onStateToggle && t('saveButton')}
-      {!!onStateToggle && <Dropdown.Item>
-        <StateButton state={data.state} isActive={isActiveState} onStateToggle={onStateToggle} label={stateButtonLabel} />
-      </Dropdown.Item>}
+      {!!onStateToggle && <>
+        {isActiveState && <Dropdown.Item>
```
luixlive: The conditions to render the dropdown items are weird and not mutually exclusive. We should render each item just once by combining all its conditions:
```Javascript
<>
  {(isActiveState || isInReview) && <Dropdown.Item>
    <StateButton
      targetState={EVENT_FORM_STATES.RESOLVED}
      onStateToggle={onStateToggle}
      label={t('stateResolveButton')}
    />
  </Dropdown.Item>}

  {isInReview && <Dropdown.Item>
    <StateButton
      targetState={EVENT_FORM_STATES.ACTIVE}
      onStateToggle={onStateToggle}
      label={t('stateActivateButton')}
    />
  </Dropdown.Item>}

  {!isActiveState && !isInReview && && <Dropdown.Item>
    <StateButton
      targetState={EVENT_FORM_STATES.ACTIVE}
      onStateToggle={onStateToggle}
      label={t('stateReopenButton')}
    />
  </Dropdown.Item>}
</>
```

src/EditableItem/styles.module.scss
```diff
@@ -362,6 +362,9 @@
     background: darken($primary-button-color, 10%);
     padding: 0 0.5rem;
   }
+  :global(.dropdown-item) button {
+    width: 100%;
+  }
```
luixlive: This will affect any button inside any Bootstrap `.dropdown-item` that is a descendant of `.formButtons`. We should go for a more targeted approach: add a class to `StateButton` and scope the rule to that.

src/EventItemContextMenu/index.js
```diff
@@ -110,20 +111,26 @@ const EventItemContextMenu = ({ children, className = '', report }) => {
...
     }
-  }, [dispatch, isActive, report, setStateContainedReport, t]);
+  }, [dispatch, report, setStateContainedReport, t]);
 
   return <ContextMenu className={className} disabled={isLoading} options={
     <>
-      <Dropdown.Item className={styles.option} onClick={updateReportState}>
-        {t(`updateReportStateItem.${isActive ? 'resolve': 'reopen'}`)} #{report.serial_number}
-      </Dropdown.Item>
+      {(isActive || isInReview) && <Dropdown.Item className={styles.option} onClick={() => updateReportState(EVENT_FORM_STATES.RESOLVED)}>
```
luixlive, line 125: Here seems to be a bug. This dropdown should match the one from `src/EditableItem/Footer/index.js`, but it doesn't. The condition branches that decide when to show certain items don't match.

src/GetUserLocationButton/index.js
```diff
@@ -41,13 +41,22 @@ const GetUserLocationButton = ({ onClick = null, onGet, ref, renderContent = nul
           (error) => {
             setIsLoading(false);
 
-            toast.error(t('errorToastMessage', { errorMessage: error.message }));
+            if (onError) {
```
luixlive: nit: I see the exact same code twice. Maybe we can reuse it?
```Javascript
const reportError = (error) => onError
  ? onError(error)
  : toast.error(t('errorToastMessage', { errorMessage: error.message }));
...
reportError(error);
```

src/hooks/useJumpToLocation/index.js
```diff
@@ -56,12 +55,13 @@ const useJumpToLocation = () => {
...
   }), [routerLocation.pathname]);
 
-  return (coords, zoom = 15) => {
+  return async (coords, zoom = 15) => {
     const isArrayCoords = Array.isArray(coords[0]);
 
     const padding = calcPadding(currentTab, isArrayCoords, itemId, isMediumLayoutOrLarger);
 
     if (isArrayCoords && coords.length > 1) {
+      const { LngLatBounds } = await import('mapbox-gl');
```
luixlive: This dynamic import is unjustified and will add latency to our calls to `jumpToLocation`.

The dynamic import means lazy loading the `LngLatBounds` from `mapbox-gl` in this line. But we are already importing this object in many other parts of the app, so there's no benefit, we are not reducing the bundle size. In contrast, now we have the downside of having an extra chunk and having to do an extra network request and wait for it before moving the map to the next location.

src/LocationPicker/MenuPopover/index.js
```diff
@@ -34,6 +38,7 @@ const MenuPopover = ({
 }) => {
   const { t } = useTranslation('components', { keyPrefix: 'locationPicker.menuPopover' });
 
+  const map = useContext(MapContext);
```
luixlive: Reading from the map context here is removing purity from our component. This is supposed to be a location picker field, used in forms and a few places in the app. It should be agnostic from business logic and parent context. But this change now requires it to be wrapped by the `MapContext`. Of course it will work because the app is always wrapped, but the component is not agnostic anymore and this complicates its testing.

The right way to achieve this would be to receive a prop in `LocationPicker` like `showPickMapLocationButton = true`, so the parent can override the default to `false` and here we simply render the button conditionally. The parent should be the one reading the map context and showing / hiding the button, which would keep our component pure and easier to test.

src/LocationPicker/index.js
```diff
@@ -142,12 +148,15 @@ const LocationPicker = ({
       />
     </div>
 
+    {locationError && <p className={styles.locationError} role="alert">{locationError}</p>}
```
luixlive: nit: Is there a reason to add this new error message in our component? It already renders the user's location button conditionally only when we have permissions, so in which scenario do we expect to get errors?

src/ReportManager/DetailsSection/SchemaForm/utils/getFormDataWithFixedTimezones/index.test.js
luixlive, line 1: This seems like an error. This file was removed and it's not in our `develop` branch. I don't know why this PR is adding it again. It's adding only the testing file actually, which is even weirder. Let's remove it.

src/ReportManager/DetailsSection/index.js
```diff
@@ -47,6 +47,9 @@ const DetailsSection = ({
   eventId,
   eventSchema = null,
   formValidator,
+  hideReportedBy = false,
+  hidePriority = false,
+  isCommunity = false,
```
luixlive, line 56: It seems to me that we don't need 3 props for this, since we are hiding all these fields just based on if we are in "community" mode or not. If that's the case, we can simplify this to a single prop `isCommunity` and then decide what to show or hide based on that.

src/ReportManager/DetailsSection/index.js
```diff
@@ -270,7 +273,7 @@ const DetailsSection = ({
       <button ref={submitFormButtonRef} type="submit" />
     </Form>}
 
-    {eventType?.version === 2 && eventSchema && <SchemaForm
+    {eventType?.version === 2 && eventSchema?.json && !(eventSchema instanceof Error) && <SchemaForm
```
luixlive: This has a smell, `eventSchema` shouldn't be an instance of `Error`. If we get an error while fetching the schema or something, we should handle it correctly there, but not pass an error masked as the schema.

src/ReportManager/DetailsSection/styles.module.scss
```diff
@@ -37,7 +37,7 @@
     display: block;
 
     @media (min-width: $md-layout-width-min) {
-      align-items: flex-end;
+      align-items: flex-start;
```
luixlive: nit: Any reasons to alter this style? I don't see changes when I test it in the app, but it worries me this could break the UI in some scenarios.

src/ReportManager/ReportDetailView/index.js
```diff
@@ -112,9 +116,17 @@ const ReportDetailView = ({
...
   onSaveAddedReport: onSaveAddedReportCallback = null,
   reportData = null,
   reportId,
+  hidePriority = false,
+  hideReportedBy = false,
+  isCommunity = false,
+  saveExtraParams = EMPTY_OBJECT,
+  schemaFetchAxiosConfig = EMPTY_OBJECT,
+  schemaFetchExtraParams = EMPTY_OBJECT,
+  skipSchemaFetch = false,
```
luixlive: All these many new props in a component as big as this one put me nervous 😅 Specially the ones with technical implementation details of other components, like `schemaFetchAxiosConfig`.

Anyway, I think I already added comments in the corresponding file for each of these props. If we work on each of those comments, eventually all these should be either removed or simplified 👍

src/ReportManager/ReportDetailView/index.js
```diff
@@ -665,10 +686,10 @@ const ReportDetailView = ({
   ]);
 
   useEffect(() => {
-    if (!!reportForm && !!eventType && !eventSchema) {
-      dispatch(fetchEventTypeSchema(reportForm.event_type, reportForm.id));
+    if (!skipSchemaFetch && !!reportForm && !!eventType && !eventSchema) {
```
luixlive: We don't need `skipSchemaFetch` props. We know we don't want to re-fetch the schema if we are in community mode, so `if (isCommunity && !!reportForm && !!eventType && !eventSchema) {` should be enough and save us one prop.

src/ReportManager/ReportDetailView/index.js
```diff
@@ -867,30 +893,57 @@ const ReportDetailView = ({
...
+                    onClick={onClickSaveButton}
+                    type="button"
+                    variant="primary"
+                  >
+                    {t('reportDetailView.saveSplitButton.title')}
                   </Button>
-                </Dropdown.Item>
-              </SplitButton>
+                : <SplitButton
+                    className={styles.saveButton}
```
luixlive, line 912: The same feedback from the `src/EditableComponent/Footer/index.js` file about improving the readability of when the dropdown items are rendered applies here.

Also, I see a couple linter errors in this file.

src/ReportManager/ReportDetailView/styles.module.scss
```diff
@@ -129,6 +131,10 @@
     background-color: unset;
   }
 
+  :global(.dropdown-item) button {
+    width: 100%;
+  }
```
luixlive, line 136: nit: The two changes in this file are funky. Making the position of `ReportDetailView` relative, having all the children it has, requires to make sure everything is inside it still renders properly in desktop and mobile. Also, we are using again a `:global` selector. As already mentioned in another file, this has the risk of altering how any dropdown item child of this component is rendered. Better to use more narrow-scoped selectors.

src/ReportManager/index.js
```diff
@@ -23,8 +23,9 @@ import ReportDetailView from './ReportDetailView';
 import * as styles from './styles.module.scss';
 
 const ADDED_REPORT_TRANSITION_EFFECT_TIME = 600;
+const EMPTY_OBJECT = {};
 
-const ReportManager = ({ onReportBeingAdded = null }) => {
+const ReportManager = ({ hidePriority = false, hideReportedBy = false, isCommunity = false, onBack = null, onReportBeingAdded = null, fallbackPath = null, newReportTypeId: newReportTypeIdProp = null, reportId: reportIdProp = null, saveExtraParams = EMPTY_OBJECT, schemaFetchAxiosConfig = EMPTY_OBJECT, schemaFetchExtraParams = EMPTY_OBJECT, skipSchemaFetch = false }) => {
```
luixlive: Holy... 😅  I guess same thing I mentioned in `src/ReportDetailView/index.js`. We need to go through each of the comments I've made and see how much this list of new props is reduced. And again, a lot of them can be omitted by using `isComunity` as a general flag.

src/DasIcon/index.js
```diff
@@ -24,6 +25,19 @@ export const calcIconUrl = (type, iconId) => {
...
+    const communityValue = window.location.pathname.split('/')[2];
+    return (
+      <img
+        alt=""
+        className={className || ''}
+        onError={(e) => { e.target.style.display = 'none'; }}
```
luixlive: `e.target.style.display = 'none';` is a React anti-pattern. It would be cleaner to conditionally render the icon based on if there are any errors:
```Javascript
const [hasError, setHasError] = useState(false);

if (hasError) return null;

if (window.location.pathname.startsWith('/community') && type === 'events' && iconId) {
  return (
    <img
      ...
      onError={() => setHasError(true)}
      ...
    />
  );
}

...
```

luixlive (changes requested):
Great work so far! Some feedback from my side:

**0- First of all, we should update this branch with the latest develop branches.**

**1- This branch has failing unit tests:**
Image

**2- My sincere opinion on the UI: we need the design team to get involved 😅** 
Community page before selecting event type:
Image

After selecting event type:
Image

**3- Code changes**
Lots of changes, lots of feedback items. Sorry in advance for so many comments. Some of them more important than others. As usual, I will mark as nitpicks the ones that are no blockers. I'm pretty sure by providing the text to an LLM and pointing to the specific line were the comment was added will make it easy for the agent to answer my questions or apply a fix.

You can provide the whole PR diff to the agent by copying all the text from https://patch-diff.githubusercontent.com/raw/PADAS/das-web-react/pull/1516.diff .  Then feed that to your agent and then start providing one by one each of the comments. Shouldn't take that long and the quality of the code should be improved considerably!

src/ducks/event-schemas/index.js
jeslefcourt: I dug into the backend before removing this — it's load-bearing. In `das`, the community schema endpoint (`retrieve_schema` in `community_input_public.py`) branches on the event type's version:

```python
if event_type.version == EventType.VersionChoices.VERSION_1:
    view = EventTypeSchemaView()          # returns v1-format { schema, definition }
    return view.get(request)
else:
    schema_service = EventTypeSchemaService()  # returns v2-format { json, ui }
```

So the community endpoint genuinely can return a v1-format schema whenever the event type is `VERSION_1`. Because we now tag every community event type as `version: 2` on the client, they all flow through the v2 branch here — including the backend-VERSION_1 ones — and the `if (rawSchema?.schema)` check is what catches those. Without it, VERSION_1 community event types stop rendering.

Honestly I think this v2 community endpoint is set up the way all the v2 endpoints should be — it returns the right data regardless of where/when the event type was created, so the client doesn't have to call different API versions for different data. So I'd like to keep this fallback. I did apply the structural part of your suggestion though — folded the community fetch into the v2 branch via a URL ternary, since community types already belong there.

jeslefcourt:
## Review feedback addressed

Thanks for the thorough review @luixlive (and Copilot). I've worked through all the comments — branch updated with latest `develop` and pushed. Summary of resolutions below; resolving the threads that are fully handled and leaving two open (noted at the end).

### Ducks / community data flow
- **fetchEventTypes / fetchEventsSchema / fetchEventTypeSchema** — simplified signatures to take a single `communityInputValue` (or none); `skipAuth` now lives inside the ducks instead of being injected by callers.
- **fetchEventTypeSchema** — folded the community fetch into the v2 branch via a URL ternary; reused the single `locationParam`. Kept the v1-format fallback inside the v2 branch (see the dedicated reply on that thread — the backend genuinely returns v1-format for VERSION_1 community event types).
- **community.js** — added the missing Redux reducer (`globallyResettableReducer`); `CommunityPage` now reads community info from the store.
- **events.js** — the save action creators (`createEvent`/`updateEvent`/`addNoteToEvent`/`uploadEventFile`) now take `communityInputValue` instead of an `extraParams` config object.

### CommunityPage
- Removed the unused `typesListRef`; added unmount cleanup for the submitted-modal timeout (+ a `SUBMITTED_MODAL_TIMEOUT` constant); flattened the nested ternaries into a `content` variable; un-shadowed the i18n `t`; extracted a single `creatableEventTypes` memo (was filtered twice); fixed the `react-hooks/refs` (temporal-id now derived via `useMemo`) and `jsx-indent` lint errors; switched the test mock pattern to top-level imports.

### State-change dropdowns (Footer / EventItemContextMenu / ReportDetailView)
- Unified all three on one transition matrix: **active → Resolve + Send to review; review → Activate + Resolve; resolved → Reopen** — each item rendered once (no more duplicate Resolve). Fixed a real bug in `EditableItem/Footer`: `isActiveState` was an external prop no caller passed, so active events wrongly showed only "Reopen"; it's now derived from `data` via `isReportActive`. Scoped the broad `:global(.dropdown-item)` CSS selectors to the specific button classes.

### ReportManager prop reduction
- Collapsed `hidePriority` / `hideReportedBy` to the single `isCommunity` flag (they were only ever passed together with it) and removed the dead `skipSchemaFetch` prop (never passed). `communityInputValue` is the one remaining community prop, used for both schema fetch and save.
- **DetailsSection** — `eventSchema` is no longer an `Error` masquerading as the schema: the schema reducer now stores `{ error }` on fetch failure and the component reads `eventSchema?.error`.

### Location components
- `GetUserLocationButton` — extracted a `reportError` helper (was duplicated). `useJumpToLocation` — replaced the dynamic `import('mapbox-gl')` with a static `LngLatBounds` import. `LocationPicker`/`MenuPopover` kept pure — the map check moved into `PickMapLocationButton` (self-hides when there's no map); removed the inline location-error message.

### DasIcon
- Replaced the `e.target.style.display = 'none'` DOM mutation with a `hasError` state that conditionally renders null.

### i18n
- Removed dead keys and stripped untranslated English placeholders from the non-English locales so they fall back to `en-US` (per `fallbackLng`); real translations to come via the translation process. **Note:** `stateReviewButton` and `saveAndReviewItem` were briefly removed as "unused" but then **re-added** — the agreed dropdown matrix gives the active state a "Send to review" option, so they're used again.
- Also repaired `pt/reports.json`, which was invalid JSON on the branch (a bad merge), by restoring it from `develop`.

### Env / infra / routing
- Reverted the `.env.development` host change; restored the accidentally-removed `/config.js` nginx blocks (kept the `community` route); removed the invalid bare `/community` route (now falls through); removed the `MapContext` re-export from `App.js` and repointed importers to `../MapContext`.

### Also
- Merged latest `develop`. Fixed the failing unit tests (the antimeridian `buildGeoSpanFilter` assertion was stale — the code change was correct; removed an orphan `getFormDataWithFixedTimezones` test whose implementation never existed). Full suite green: 2399 passing.

---

**Left open intentionally:**
- The v1-format-fallback thread (replied with backend evidence — I'd like to keep it).
- The Activity-section **header on community** — I kept it hidden as an intentional design choice and would rather loop in design before changing (per your overall note), so I left that thread for you/design.

One remaining `react-hooks/set-state-in-effect` lint warning in `CommunityPage`'s data-fetch effect predates this work; happy to take it in a follow-up.

src/DetailViewComponents/ActivitySection/index.js
jeslefcourt: Two parts:

1. **Header hidden on community** — that was intentional; the public community form is a stripped-down view. I agree with your overall point that design should weigh in here, so I'd rather flag it for them than flip it unilaterally. Leaving the `!isCommunity` guard for now.

2. **The bullet** — fixed. Confirmed in the running app it was the Activity-section `.list` `<ul>` rendering a `disc` on its note `<li>` (Chromium keeps the marker on a `display:flex` list when the item's UA display is `list-item`). Added `list-style: none; padding-left: 0;` to `.list`.

src/CommunityPage/index.js
```diff
@@ -0,0 +1,149 @@
...
+  }, [navigate, value]);
+
+  const onBack = useCallback(() => {
+    navigate(`/community/${value}`);
+    setShowSubmittedModal(true);
+    timeoutRef.current = setTimeout(() => setShowSubmittedModal(false), SUBMITTED_MODAL_TIMEOUT);
+  }, [navigate, value]);
+
+  let content;
+  if (isLoading || isUnauthorized || (!eventTypeValue && !isUnauthorized && eventsByCategory[0]?.types.length === 1)) {
```
luixlive: A couple comments here:

1. The third condition check is obscure. Defining a variable with an explicit name could help:
```Javascript
const explicitNameExplainingThisCheck = !eventTypeValue && !isUnauthorized && eventsByCategory[0]?.types.length === 1;

let content;
if (isLoading || isUnauthorized || explicitNameExplainingThisCheck) {
```

2. The condition is trying to merge two different states: loading and unauthorized. For that reason, its content has ternary operators, conditional renders, etc. Separating it in two conditions would give a clearer intent and a cleaner content definition:

```Javascript
if (isLoading) {
  content = (
    <div className={styles.loadingView}>
      <MoonLoader color="#333" size={50} />
    </div>
  );
} else if (isUnauthorized) {
  content = (
    <div className={styles.loadingView}>. // Probably a different name for the classname.
      <EarthRangerLogo className={styles.earthRangerLogo} />

      <span className={styles.loadingError}>
        {t('invalidCommunityUrl')}
      </span>
    </div>
  );
} else if (...) {
```

src/ducks/event-types/index.js
```diff
@@ -9,21 +9,30 @@ const USE_EVENT_TYPES_V2_MOCK_API = import.meta.env.REACT_APP_MOCK_EVENTTYPES_V2
...
   const [eventTypesResponse, eventTypesV2Response] = await Promise.all([
     axios.get(EVENT_TYPES_API_URL),
     axios.get(EVENT_TYPES_V2_API_URL)
   ]);
 
-  const eventTypes = [
-    ...eventTypesResponse.data.data.map((eventType) => ({ ...eventType, version: 1 })),
-    ...eventTypesV2Response.data.data.map((eventType) => ({ ...eventType, version: 2 })),
-  ];
+  const byValue = new Map();
```
luixlive, line 32: The addition of the community input  flow is above in an `if`. This change here seems unrelated and out of scope. Just changing the way in which merge the event types, but the previous approach feels cleaner to me.

src/ducks/events.js
```diff
@@ -21,6 +21,13 @@ export const EVENTS_API_URL = (
   : `${API_URL}activity/events`;
 export const EVENT_API_URL = `${API_URL}activity/event/`;
 
+const COMMUNITY_EVENTS_API_URL = (communityValue) =>
```
luixlive: Changes in this file seem to be missing unit tests. It'd be simple, yet valuable, to add some tests that validate the correct URL is used depending on if a `communityInputValue` is passed to the action creators or not.

src/hooks/useJumpToLocation/index.js
```diff
@@ -56,7 +56,7 @@ const useJumpToLocation = () => {
     itemId: getCurrentIdFromURL(routerLocation.pathname),
   }), [routerLocation.pathname]);
 
-  return (coords, zoom = 15) => {
+  return async (coords, zoom = 15) => {
```
luixlive: We made this method `async` but we are not using `await` inside it. Maybe old code that Claude forgot to clean?

src/PickMapLocationButton/index.js
```diff
@@ -74,6 +74,10 @@ const PickMapLocationButton = ({
     }
   }, [isPickingMapLocation, map, onCancel, onPick, setAppToShowMapMode]);
 
+  if (!map) {
```
luixlive, line 77: This component is useless if there is no map available, but from an accessibility and UI perspective, it would be better to still render it, just disabling the button. The logic to decide if the button is mounted or not should reside in the parent. But from here, it would be cleaner to just disable it if there's no map. A nice to have would be an `aria-description` stating why it is disabled.

src/ReportManager/DetailsSection/index.js
```diff
@@ -246,7 +247,7 @@ const DetailsSection = ({
     </div>
 
     {/* Legacy form renderer */}
-    {eventType?.version === 1 && !!jsonSchema && <Form
+    {!!eventSchema?.schema && !!jsonSchema && <Form
```
luixlive, line 259: In this file we removed the checks to `eventType.version === 1` like if we don't trust the `version` parameter any more. Instead we check if `eventSchema.schema` is truthy. Technically, it's not wrong, but checking the version number is semantically clearer. Why did we change it?

src/ReportManager/ReportDetailView/index.js
```diff
@@ -112,9 +114,15 @@ const ReportDetailView = ({
   isNewReport,
   newReportTypeId = null,
   onAddReport = null,
+  onBack = null,
   onSaveAddedReport: onSaveAddedReportCallback = null,
   reportData = null,
   reportId,
+  hidePriority = false,
+  hideReportedBy = false,
+  isCommunity = false,
```
luixlive, line 123: I still wonder if we really want `hidePriority`, `hideReportedBy`, `skipSchemaFetch`... as independent props instead of simply calculating them from the value of `isCommunity`.

src/ReportManager/index.js
```diff
@@ -24,7 +24,7 @@ import * as styles from './styles.module.scss';
 
 const ADDED_REPORT_TRANSITION_EFFECT_TIME = 600;
 
-const ReportManager = ({ onReportBeingAdded = null }) => {
+const ReportManager = ({ communityInputValue = null, isCommunity = false, onBack = null, onReportBeingAdded = null, fallbackPath = null, newReportTypeId: newReportTypeIdProp = null, reportId: reportIdProp = null }) => {
```
luixlive: 2 nits:

1. Maybe we can drop `isCommunity` prop and just use `communityInputValue`. We know that if `communityInputValue = null` we are not in community flow, if `communityInputValue` has a truthy value, we are in community flow.

2. Suffixing a prop with `Prop` doesn't help understand why we would receive the value from props vs from the search params. Better to give a more explicit name to the prop that says out loud an intention: 
```Javascript
const ReportManager = ({ communityCreatedEventTypeId }) => {
```

Btw, I see the "report" term, but we've been trying to move to "event" in every new PR.

src/CommunityPage/index.js
```diff
@@ -0,0 +1,149 @@
...
+  useEffect(() => () => clearTimeout(timeoutRef.current), []);
+
+  const onClickType = useCallback((reportType) => {
+    navigate(`/community/${value}/${reportType.value}`);
+  }, [navigate, value]);
+
+  const onBack = useCallback(() => {
+    navigate(`/community/${value}`);
+    setShowSubmittedModal(true);
+    timeoutRef.current = setTimeout(() => setShowSubmittedModal(false), SUBMITTED_MODAL_TIMEOUT);
```
luixlive, line 97: If `onBack` is called more than once, the previous timeout leaks. Fix: `clearTimeout(timeoutRef.current)` at the top of `onBack` before setting the new one.

src/CommunityPage/index.js
```diff
@@ -0,0 +1,149 @@
...
+    dispatch(fetchEventTypes(value))
+      .then(() => setIsLoading(false))
+      .catch(() => setIsLoading(false));
+    dispatch(fetchEventsSchema(value));
+  }, [dispatch, value]);
+
+  useEffect(() => {
+    if (isLoading || isUnauthorized || eventTypeValue) return;
+    if (creatableEventTypes.length === 1) {
+      navigate(`/community/${value}/${creatableEventTypes[0].value}`, { replace: true });
```
luixlive, line 84: Weird UX: when there is only one `creatableEventType`, `onBack` navigates to `/community/${value}`, which immediately triggers the auto-redirect effect back to `/community/${value}/${type}`.

src/CommunityPage/index.js
```diff
@@ -0,0 +1,149 @@
...
+    }
+    setIsLoading(true);
+    setIsUnauthorized(false);
+    dispatch(fetchCommunityInfo(value))
+      .catch(() => { setIsUnauthorized(true); });
+    dispatch(fetchEventTypes(value))
+      .then(() => setIsLoading(false))
+      .catch(() => setIsLoading(false));
+    dispatch(fetchEventsSchema(value));
+  }, [dispatch, value]);
```
luixlive, line 79: Race condition: `fetchCommunityInfo` and `fetchEventTypes` are independent dispatches. If `fetchEventTypes` resolves before `fetchCommunityInfo` rejects, there is a window where `isLoading=false && isUnauthorized=false`, flashing the (empty) community page before the error state is set.

src/DasIcon/index.js
```diff
@@ -24,6 +26,26 @@ export const calcIconUrl = (type, iconId) => {
 };
 
 const DasIcon = ({ type, iconId, color = 'gray', dispatch: _dispatch, className, ...rest }) => {
+  const location = useLocation();
+  const [hasError, setHasError] = useState(false);
+
+  if (location.pathname.startsWith('/community') && type === 'events' && iconId) {
+    if (hasError) {
+      return null;
```
luixlive, line 34: If a community icon sets `hasError` to `true`, then the `iconId` prop changes, the component stays mounted and `hasError` remains `true`, so `null` is returned for the new `iconId` even though it might load fine. Fix: add `useEffect(() => setHasError(false), [iconId])`.

src/DasIcon/index.js
```diff
@@ -24,6 +26,26 @@ export const calcIconUrl = (type, iconId) => {
...
+    }
+
+    const communityValue = location.pathname.split('/')[2];
```
luixlive: This URL parsing hardcodes the assumption that `communityValue` is always the third segment of the path. If the routing prefix ever changes this breaks silently. Since `communityValue` is already in Redux state, we could read it from the store instead of re-reading it from the URL. This also makes the component easier to test.

src/ducks/event-schemas/index.js
```diff
@@ -23,57 +28,62 @@ export const FETCH_EVENT_TYPE_SCHEMA_V2_SUCCESS = 'FETCH_EVENT_TYPE_SCHEMA_V2_SU
...
-        type: FETCH_EVENT_TYPE_SCHEMA_V2_SUCCESS,
+      const response = await axios.get(url, {
+        params: { event_id: eventId, location: locationParam, pre_render: true },
+        ...(communityInputValue ? { skipAuth: true } : {}),
       });
+
+      const rawSchema = response.data;
+
+      if (rawSchema?.schema) {
+        // v2 endpoint returns v1-format schema for VERSION_1 event types (e.g. community)
```
luixlive: Comment is misleading. Community event types are all forced to version 2 in `fetchEventTypes`. The parenthetical "(e.g. community)" is wrong.

src/ReportManager/ReportDetailView/index.js
```diff
@@ -778,14 +797,17 @@ const ReportDetailView = ({
...
-        </QuickLinks.NavigationBar>
+        </QuickLinks.NavigationBar>}
 
         <div className={styles.content}>
           <QuickLinks.SectionsWrapper>
             <QuickLinks.Section anchorTitle={t('reportDetailView.quickLinks.detailsAnchor')}>
               <DetailsSection
                 eventId={reportId}
                 eventSchema={eventSchema}
+                hidePriority={hidePriority}
```
luixlive, line 807: `DetailsSection` doesn't read `hidePriority` and `hideReportedBy` props. It gates those fields entirely on `!isCommunity`. Let's remove these props from here.

luixlive (approved):
Thanks so much for the effort and pacience @jeslefcourt . This PR seems ready to me 💯  

I still left a bunch of comments, but this time I see no blocker. They're just recommendations to fix possible minor bugs and increase the code readability, integration with old patterns, etc. Please read them, it should be fast, and I'll leave to your consideration if you want to apply some of them. It should be faster, I'm not asking for big architectural changes like in the first pass 😄

JoshuaVulcan (approved):
Looking good! Please clean up merge conflicts before merging down.

jeslefcourt:
## Round 2 — addressed (pushed `8e632d07`)

Thanks @luixlive (and Copilot). Rundown:

**Fixed**
- **events.js**: community create/note/file writes now send `skipAuth: true` — the community endpoints are unauthenticated (confirmed in `das`: `permission_classes = []` / `CommunityInputEventPermission`). `updateEvent` keeps its param (with a comment): community is **create-only** — there's no community event-update endpoint and a community report never routes to `updateEvent` (new reports have no `id` → `createEvent`). Added unit tests for the community URL + `skipAuth`.
- **Header**: restored `reviewLabel` — it's looked up dynamically via `t(`${state}Label`)`, so it wasn't actually unused (my earlier literal grep missed it). Good catch.
- **DasIcon**: reads the community value from Redux (`state.data.community.value`) instead of parsing the path → route-prefix safe; and tracks the errored `iconId` so a later valid icon isn't suppressed.
- **CommunityPage**: clearer render branches; `onBack` clears the prior timeout; single-type `onBack` refreshes the form in place instead of bouncing through the auto-redirect; `Promise.allSettled` removes the empty-page flash.
- **DetailsSection**: added a comment on why we gate by schema shape rather than `eventType.version` (community types are forced to v2 client-side, but the backend can still return a v1-format schema).
- **useJumpToLocation**: dropped the now-unnecessary `async`. **event-schemas**: reworded the misleading comment. **map.js**: `buildGeoSpanFilter` now normalizes longitude order unless the span actually crosses the antimeridian (`|Δlon| > 180`), with tests.
- **i18n**: added **es/fr/pt** translations for the community + review-state keys. ne-NP/sw fall back to en-US for now and will go through the translation process (draft sheet prepared).

**Left open for your call** (replied inline): the `fetchEventTypes` dedup, `isCommunity` vs `communityInputValue`, keeping `hidePriority`/`hideReportedBy`, and `PickMapLocationButton` hide-vs-disable — I leaned toward keeping these and explained why, but happy to go your way.

The `ref`-on-function-component comments are React-19 behaviors (we're on 19.2.5 — `ref` is a regular prop now); replied inline.

src/ducks/event-types/index.js
jeslefcourt: Kept the dedup (Map keyed by `value`, v1 winning) rather than the plain concat: if the same event-type `value` comes back from both the v1 and v2 endpoints, a concat renders it twice (and yields duplicate React keys). Happy to switch to the simpler concat if we're confident the two endpoints never overlap.

src/PickMapLocationButton/index.js
jeslefcourt: I left this returning `null` rather than a disabled button. In the community form there is no map at all, so a permanently-disabled "Set on map" control — with no way for a public user to ever enable it — felt more confusing than helpful. The mount decision is effectively the parent's (the button only has a map in the authenticated app). Happy to switch to a disabled + `aria-description` button if you'd prefer it stay visible.

src/ReportManager/ReportDetailView/index.js
jeslefcourt: Removed `skipSchemaFetch` (it was never passed by any caller). I kept `hidePriority`/`hideReportedBy` though, and made them functional again (DetailsSection now gates those fields on them). The rationale (in a comment): they're intentionally generic visibility props in the component's own vocabulary, so a future caller can hide those fields for some reason other than community without this component accumulating more "why" flags. Open to collapsing them if you feel strongly.

src/ReportManager/index.js
jeslefcourt: Kept `isCommunity` as an explicit prop. It drives a fair amount of genuinely community-specific UI/behaviour (the Header menu, the nav bar, the save redirect, the DetailsSection title), so deriving it from `!!communityInputValue` everywhere read as less clear than an explicit flag. Happy to change if you'd prefer.

src/ReportManager/ReportDetailView/index.js
jeslefcourt: Rather than removing these, I made them functional again — `DetailsSection` now gates the priority / reported-by fields on `hidePriority` / `hideReportedBy` (not on `isCommunity`), and they're wired through from `CommunityPage`. Rationale (in a comment on the component): keep field visibility in the component's own vocabulary instead of threading the *reason* (community) into it. So they're no longer dead/misleading.

src/GetUserLocationButton/index.js
```diff
@@ -12,14 +12,18 @@ import LoadingOverlay from '../LoadingOverlay';
 
 import * as styles from './styles.module.scss';
 
-const GetUserLocationButton = ({ onClick = null, onGet, ref, renderContent = null, ...otherProps }) => {
+const GetUserLocationButton = ({ onClick = null, onError = null, onGet, ref, renderContent = null, ...otherProps }) => {
   const dispatch = useDispatch();
```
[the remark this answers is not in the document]
jeslefcourt: This is fine on React 19 (we're on 19.2.5): function components now receive `ref` as a regular prop (forwardRef is deprecated). `GetUserLocationButton` destructures `ref` and applies it to its `<button ref={ref}>`, so the focus trap does attach.

src/LocationPicker/MenuPopover/index.js
```diff
@@ -166,16 +167,18 @@ const MenuPopover = ({
...
-
-            {t('pickMapLocationButton')}
-          </>}
-        />
+        <Suspense fallback={null}>
+          <PickMapLocationButton
+            onClick={() => eventReportTracker.track('Click \'Set on map\'')}
+            onPick={onMapLocationPick}
+            ref={!showUserLocation ? lastFocusableElementRef : undefined}
+            renderContent={() => <>
```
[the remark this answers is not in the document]
jeslefcourt: Same as the `GetUserLocationButton` thread: on React 19 (19.2.5) `ref` is passed as a normal prop, and both `PickMapLocationButton` and `GetUserLocationButton` forward it to their underlying element, so `lastFocusableElementRef.current` does get set.

src/ducks/event-types/index.js
jeslefcourt: Kept the dedup (Map keyed by `value`, v1 winning) rather than the plain concat: if the same event-type `value` comes back from both the v1 and v2 endpoints, a concat renders it twice (and yields duplicate React keys). Happy to switch to the simpler concat if we're confident the two endpoints never overlap.

src/PickMapLocationButton/index.js
jeslefcourt: I left this returning `null` rather than a disabled button. In the community form there is no map at all, so a permanently-disabled "Set on map" control — with no way for a public user to ever enable it — felt more confusing than helpful. The mount decision is effectively the parent's (the button only has a map in the authenticated app). Happy to switch to a disabled + `aria-description` button if you'd prefer it stay visible.

src/ReportManager/ReportDetailView/index.js
jeslefcourt: Removed `skipSchemaFetch` (it was never passed by any caller). I kept `hidePriority`/`hideReportedBy` though, and made them functional again (DetailsSection now gates those fields on them). The rationale (in a comment): they're intentionally generic visibility props in the component's own vocabulary, so a future caller can hide those fields for some reason other than community without this component accumulating more "why" flags. Open to collapsing them if you feel strongly.

src/ReportManager/index.js
jeslefcourt: Kept `isCommunity` as an explicit prop. It drives a fair amount of genuinely community-specific UI/behaviour (the Header menu, the nav bar, the save redirect, the DetailsSection title), so deriving it from `!!communityInputValue` everywhere read as less clear than an explicit flag. Happy to change if you'd prefer.

src/ReportManager/ReportDetailView/index.js
jeslefcourt: Rather than removing these, I made them functional again — `DetailsSection` now gates the priority / reported-by fields on `hidePriority` / `hideReportedBy` (not on `isCommunity`), and they're wired through from `CommunityPage`. Rationale (in a comment on the component): keep field visibility in the component's own vocabulary instead of threading the *reason* (community) into it. So they're no longer dead/misleading.

src/GetUserLocationButton/index.js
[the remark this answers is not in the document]
jeslefcourt: This is fine on React 19 (we're on 19.2.5): function components now receive `ref` as a regular prop (forwardRef is deprecated). `GetUserLocationButton` destructures `ref` and applies it to its `<button ref={ref}>`, so the focus trap does attach.

src/LocationPicker/MenuPopover/index.js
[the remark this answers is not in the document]
jeslefcourt: Same as the `GetUserLocationButton` thread: on React 19 (19.2.5) `ref` is passed as a normal prop, and both `PickMapLocationButton` and `GetUserLocationButton` forward it to their underlying element, so `lastFocusableElementRef.current` does get set.

luixlive:
It's good to go @jeslefcourt ! Some of them were just suggestions and code preferences, but from my side this PR is ready to be merged and I trust your instinct. Let's 🚢 it.

PR #1526, opened 2026-04-06
ERA-11250 Swap in new cookie manager by Osano

chrisj-er:
feat: update config in vite.
ix: A osanoPlugin() Vite plugin in vite.config.js that uses transformIndexHtml to inject the Osano snippet. It checks ctx.server — when
  present, the dev server is running (local dev), so it skips injection. During vite build (used by all CI/CD deployments), ctx.server is
  undefined, so the script gets injected.

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -2,31 +2,27 @@ import { defineConfig } from 'vite';
 import react from '@vitejs/plugin-react';
 import svgr from 'vite-plugin-svgr';
 
-function osanoPlugin() {
-  return {
-    name: 'inject-osano',
-    apply: 'build',
-    transformIndexHtml: {
-      order: 'pre',
-      handler() {
-        return [
-          {
-            tag: 'style',
-            children: '.osano-cm-widget{display: none;}',
-            injectTo: 'head',
-          },
-          {
-            tag: 'script',
-            attrs: {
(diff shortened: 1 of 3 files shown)


vite.config.js
```diff
@@ -2,6 +2,32 @@ import { defineConfig } from 'vite';
 import react from '@vitejs/plugin-react';
 import svgr from 'vite-plugin-svgr';
 
+function osanoPlugin() {
```
luixlive: nit: we prefer arrow function notation in the client repos
```Javascript
const osanoPlugin = () => ({  // Arrow function here
  name: 'inject-osano',
  apply: 'build',
  transformIndexHtml: {
    order: 'pre',
    handler: () => [  // And here
      {...},
      {...},
    ],
  },
});
```

vite.config.js
```diff
@@ -28,6 +54,7 @@ export default defineConfig({
   },
 
   plugins: [
+    osanoPlugin(),
```
luixlive: This plugin shouldn't interfere at all with React or SVG transformation (the other 2 plugins we list here), but given it's scope and simplicity, I would say it's better to add that plugin at the end of the list just to keep the hierarchy by relevance.
```Javascript
  plugins: [
    react(),

    // Match CRA SVG transformation.
    svgr({ ... }),

    osanoPlugin(),
  ],
```

luixlive (approved):
No blockers, the code seems fine. I just added 2 small recommendations to better align with our code standards.

PR #1529, opened 2026-04-08
ERA-13076: Remove approach to loading icons via one enormous sprite

jeslefcourt:
## Summary
- Removes the single large SVG sprite as the icon-loading mechanism (drops the preloaded `ReportTypeIconSprite` from `App.js` and the sprite-lookup logic from `DasIcon`)
- `DasIcon` now fetches each icon's SVG from its static URL, **sanitizes the markup, and inlines it** (via `dangerouslySetInnerHTML`) so the SVG can inherit `currentColor` and be recolored by CSS. Sanitization strips hardcoded `fill`/`stroke` colors, `<style>` blocks, and class attributes, rewriting them to `currentColor`. Fetched icons are cached in a module-level `svgCache`
- Falls back to a plain `` tag when the fetched resource is not SVG, and for `type === 'subjects'` icons; falls back to the generic icon when an icon fails to load
- Updates CSS selectors in patrol/report list and header styles from `svg` to `img`, and uses `filter: invert(1)` to render icons white on dark headers

## Jira
[ERA-13076](https://allenai.atlassian.net/browse/ERA-13076)

## Test plan
- [ ] Event icons render correctly in the report list
- [ ] Patrol icons render correctly in the patrol list and detail view
- [ ] Collection icons show the correct badge icon
- [ ] Generic fallback icon displays when an icon is missing
- [ ] Community-scoped routes load icons from the community API endpoint
- [ ] Icons recolor correctly via `currentColor` (e.g. white on dark headers)
- [ ] Cluster markers render and update correctly, including when an event has unsaved local edits

[ERA-13076]: https://allenai.atlassian.net/browse/ERA-13076?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/DasIcon/index.js
```diff
@@ -1,37 +1,54 @@
...
-export const calcIconUrl = (type, iconId) => {
-  const SPRITE_ID = spriteMappings[type].id;
-  const ICON_PREFIX = spriteMappings[type].prefix;
+const GENERIC_ICON_ID = 'generic_rep';
 
-  const iconIdMatch = document.querySelector(`#${SPRITE_ID} #${ICON_PREFIX}${iconId}`);
-  const iconIdRepMatch = document.querySelector(`#${SPRITE_ID} #${ICON_PREFIX}${iconId}_rep`);
-  if (iconIdMatch) {
-    return `#${ICON_PREFIX}${iconId}`;
+const DasIcon = ({ type, iconId, imageUrl, color: _color, dispatch: _dispatch, className, ...rest }) => {
```
luixlive: The code changes look great, but I find a couple places (nitpicks) that don't follow the conventions of this repo, like import spacing, unused props, inline scoped functions, "otherProps" instead of "rest". Here an updated version you could copy and paste:
```Javascript
import React from 'react';

import { API_V2_URL, DAS_HOST } from '../constants';

const GENERIC_ICON_ID = 'generic_rep';

const DasIcon = ({ type, iconId, imageUrl, className, ...otherProps }) => {
  const onImgError = (event) => {
    event.currentTarget.style.display = 'none';
  };

  if (type === 'subjects') {
    return <img alt="" className={className} onError={onImgError} src={imageUrl} {...otherProps} />;
  }

  if (window.location.pathname.startsWith('/community') && iconId) {
    const communityValue = window.location.pathname.split('/')[2];
    return <img
      alt=""
      className={className}
      onError={onImgError}
      src={`${API_V2_URL}community/${communityValue}/activity/events/eventtypes/icons/${iconId}`}
      {...otherProps}
    />;
  }

  const effectiveIconId = iconId || GENERIC_ICON_ID;
  const isGeneric = effectiveIconId.includes('generic');

  const onSpriteImgError = (event) => {
    const img = event.currentTarget;
    if (!img.src.includes(GENERIC_ICON_ID)) {
      img.src = `${DAS_HOST}/static/sprite-src/${GENERIC_ICON_ID}.svg`;
      img.classList.add('generic');
    } else {
      img.style.display = 'none';
    }
  };

  return (
    <img
      alt=""
      className={`${className} ${isGeneric ? 'generic' : ''}`.trim()}
      onError={onSpriteImgError}
      src={`${DAS_HOST}/static/sprite-src/${effectiveIconId}.svg`}
      {...otherProps}
    />
  );
};

export default DasIcon;

```

src/DasIcon/index.js
```diff
@@ -1,37 +1,54 @@
...
+  if (type === 'subjects') {
+    return (
+      <img
+        alt=""
```
luixlive: I see the agent added an empty alt, probably just to make the linter happy, but the real accessible practice is to provide a test here. Probably something like "`Icon for ${iconId}`".

src/EventIcon/index.js
```diff
@@ -1,53 +1,43 @@
...
+    const isPatrol = !!report?.patrol_segments?.length && isObject(report.patrol_segments[0]);
+    const type = isPatrol ? report?.patrol_segments?.[0]?.patrol_type : report.event_type;
+    const matchingEventType = eventTypes.find((eventType) => eventType.value === type);
+
+    return matchingEventType?.icon_id ?? null;
   }, [eventTypes, report.event_type, report.is_collection, report.patrol_segments]);
 
-  if (!report.is_collection) {
-    return <DasIcon iconId={iconId} type="events" title={report.event_type} {...rest} />;
+  const singleContainedIconId = useMemo(() => {
```
luixlive: I understand why the agent called this variable "singleContainedIconId" (it reads from `report.contains`, so it inferred a name based on that), but I don't think it's the right wording. Maybe something a bit more verbose but explicit like `iconIdForNonContainedEventType`?

src/EventIcon/index.js
```diff
@@ -1,53 +1,43 @@
...
+    if (!report.is_collection || report.contains?.length !== 1) return null;
+
+    const containedEventType = report.contains[0]?.related_event?.event_type;
+    return eventTypes.find((et) => et.value === containedEventType)?.icon_id ?? null;
+  }, [eventTypes, report.contains, report.is_collection]);
+
+  if (singleContainedIconId) {
+    return <span className={styles.wrapper} ref={ref}>
+      <DasIcon iconId={iconId} type="events" title={report.event_type} {...rest} />
```
luixlive, line 39: I see that we used to render the second `DasIcon` only if `topRatedReportAndType && topRatedReportAndType.event_type`, but that logic was removed. I guess it's intended, but why?

src/PatrolDetailView/Header/styles.module.scss
```diff
@@ -31,14 +31,10 @@
...
-  svg {
+  img {
     width: 1.5rem;
     height: 1.5rem;
-    fill: white;
-
-    &[class*=newPatrol] {
-      fill: $dark-gray;
-    }
+    filter: invert(1);
```
luixlive: I see that we used to have `fill: white;` when the node was a `svg`; now for `img` we use `filter: invert(1);`. But that's not the same. It's like applying a negative filter to the image. Are we certain this style will work properly with all our images?

src/App.js
```diff
@@ -9,10 +9,8 @@ import { useTranslation } from 'react-i18next';
 import ReactGA4 from 'react-ga4';
 
 import { createUserAnalyticsData } from './utils/analytics';
-import { SYSTEM_CONFIG_FLAGS } from './constants';
 
 import { ReactComponent as EarthRangerLogoSprite } from './common/images/sprites/logo-svg-sprite.svg';
```
luixlive, line 14: You forgot to delete the file from `/common/images/sprites/event-svg-sprite.svg` 😄

luixlive (changes requested):
A couple small suggestions and a couple questions. Nothing hard-blocking, but worth taking a minute to think them through.

src/DasIcon/index.js
jeslefcourt: Using your code and also enabled the Linter.  Didn't have that running, sorry.

src/EventIcon/index.js
jeslefcourt: Nice catch.  Shouldn't have been.

src/ClustersLayer/utils.js
```diff
@@ -2,6 +2,7 @@ import { centroid, featureCollection } from '@turf/turf';
 import mapboxgl from 'mapbox-gl';
 
 import { CLUSTER_CLICK_ZOOM_THRESHOLD, LAYER_IDS, SUBJECT_FEATURE_CONTENT_TYPE } from '../constants';
+import { calcSvgImageIconId } from '../MapImageFromSvgSpriteRenderer';
```
JoshuaVulcan, line 5: It seems like `MapImageFromSvgSpriteRenderer` should be removed entirely, and the utility function to calculate image URLS lifted/shifted elsewhere.

src/ClustersLayer/utils.js
```diff
@@ -23,8 +24,25 @@ const FEATURE_ICON_HTML_STYLES = { maxWidth: '24px', minWidth: '18px', height: '
 const FEATURE_SS_ICON_HTML_STYLES = { filter: 'brightness(0)' };
 const FEATURE_COUNT_HTML_STYLES = { fontSize: '16px', fontWeight: '500', paddingLeft: '4px', margin: '0' };
 
-const getFeatureIcon = (feature, mapImages) =>
-  mapImages[`${feature.properties.icon_id}-${feature.properties.priority}`]?.image;
+const getFeatureIcon = (feature, mapImages, locallyEditedEvent, eventStore) => {
+  const isLocally = locallyEditedEvent?.id === feature.properties.id;
```
JoshuaVulcan, line 28: variable naming nit: `isLocally` 😢

src/DasIcon/index.test.js
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
```
JoshuaVulcan, line 1: Love the changes here. Seems like we'd want to rename away from `DasIcon` as well?

JoshuaVulcan (changes requested):
Looks and feels pretty good so far! Will keep manually testing. My feedback on this review pass is around finishing off the refactor by making sure files and names are cleaned up.

src/ClustersLayer/index.js
```diff
@@ -97,6 +105,18 @@ const ClustersLayer = ({ onShowClusterSelectPopup }) => {
 
   useMapEventBinding('sourcedata', onSourceData);
 
+  useEffect(() => {
```
luixlive, line 108: I think these `useEffect`s can have a performance impact. Their dependency arrays include the callback itself (`updateClusterMarkersCallback`), this means that everytime the callback gets redefined, they will be called. In other words, whenever any of the following variables changes, we will be running the markers update:
addClusterPolygon, eventStore, locallyEditedEvent, map, onShowClusterSelectPopup, removeClusterPolygon.

Probably we could stabilize this with a `useRef` or a more complex `useEffect` that only triggers the callback specifically when `locallyEditedEvent` and `mapImages` change.

src/ClustersLayer/utils.js
```diff
@@ -23,8 +24,25 @@ const FEATURE_ICON_HTML_STYLES = { maxWidth: '24px', minWidth: '18px', height: '
...
+  if (isLocally) {
```
luixlive, line 29: This code block gives me the impression that the agent thought locally edited events could update their icons, but they can't. All the user can changes that results in an update to the icon is priority. I feel this method can be simplified, something like:
```Javascript
const getFeatureIcon = (feature, mapImages) => {
  const priority = locallyEditedEvent?.id === feature.properties.id
    ? locallyEditedEvent.priority
    : feature.properties.priority

  return mapImages[calcSvgImageIconId({ icon_id: feature.properties.icon_id, priority })]?.image;
}
```

Unless of course I'm missing something 🤔

src/DasIcon/index.js
```diff
@@ -1,38 +1,175 @@
...
-    id: 'reportTypeIconSprite',
-    prefix: 'das--activity--static--sprite-src--',
-  },
-};
+const GENERIC_ICON_ID = 'generic_rep';
+export const svgCache = new Map();
+
+const CONTAINER_SELECTOR = 'svg,g,defs,symbol,marker,clipPath,mask,pattern';
+
+const sanitizeSvg = (text) => {
```
luixlive, line 10: Feels like a brute-force approach to clean the icons. I guess it's better to keep this file than the sprites file. Probably would try a last prompt to Claude like: "Refactor this file and apply clean code practices plus explanatory comments to improve its readability and maintainability". And I would say this requires some thorough manual testing in a couple envs just to make sure all icons render correctly 👍

luixlive (commented):
Just a couple more comments and questions. I see now we are also updating the clusters when we have a locally edited event, which is a nice addition, but it's increasing the size of this PR considerably.

Once these comments are solved / responded, I'll be ready to approve. Still think we will need a special treatment from QA to make sure all icons render correctly and that clusters work and update correctly when we have local changes in an event that haven't been saved yet 👍 @AlanCalvillo

src/ClustersLayer/utils.js
luixlive: Agreed. `isLocally` is ambiguous. The code that defines the variable isn't, but I'd rather have something like `isLocallyEditedEventFeature`.

src/SvgIcon/index.js
```diff
@@ -1,38 +1,133 @@
...
   }
+
+  return (
+    <span
+      dangerouslySetInnerHTML={{ __html: injectClass(cached.svg, className) }}
+      style={{ display: 'contents' }}
+      {...rest}
+    />
+  );
 };
```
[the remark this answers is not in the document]
jeslefcourt: Updated the PR description to describe the real approach: `SvgIcon` fetches each icon's SVG, sanitizes it, and inlines it so it can inherit `currentColor` (with an `` fallback for non-SVG/subject icons). The inlining is intentional — a plain `` can't inherit `currentColor` for recoloring, which is why we sanitize + inline.

src/DasIcon/index.js
jeslefcourt: Refactored `sanitizeSvg` into named per-pass helpers (`collectContainerDefaultFills`, `removeStyleBlocks`, `removeClassAttributes`, `stripHardcodedFills`, `rewriteStrokesToCurrentColor`, `stripInlineFillAndStroke`, `removeContainerDefaultFills`) so it reads as a clear sequence (commit b10b38e2). Behavior unchanged, covered by the SvgIcon tests. The cross-env manual icon-render verification you mentioned is flagged for QA.

src/SvgIcon/index.test.js
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
```
jeslefcourt: Renamed `DasIcon` → `SvgIcon` — component, directory, and all consumers (commit b10b38e2).

src/ClustersLayer/index.js
jeslefcourt: Good catch. Stabilized `updateClusterMarkersCallback` with refs so it no longer churns on every `eventStore` change — the two effects now depend only on `[locallyEditedEvent]` and `[mapImages]` (commit b10b38e2). The render-phase ref assignments carry a justified `react-hooks/refs` disable with rationale (commit 2c9e27f8).

src/ClustersLayer/utils.js
jeslefcourt: You were right — only priority is user-editable in a way that changes the icon. Simplified `getFeatureIcon` to override just priority for the locally-edited event, and removed the now-dead `eventStore` threading through `createClusterHTMLMarker`/`addNewClusterMarkers` (commit b10b38e2).

src/ClustersLayer/utils.js
jeslefcourt: Lifted `calcSvgImageIconId` into `src/utils/mapImages.js` and repointed all importers (commit b10b38e2). On removing `MapImageFromSvgSpriteRenderer` entirely: its default-export component is still load-bearing — it generates `state.view.mapImages`, which both EventsLayer's on-map symbol layer and the cluster markers consume — so fully removing it needs a separate migration of the on-map rendering path. I'd suggest tracking that as a follow-up rather than expanding this PR. Leaving this thread open for your call on the follow-up.

src/ClustersLayer/utils.js
jeslefcourt: Renamed `isLocally` → `isLocallyEdited` (commit b10b38e2). The other occurrence in `getFeatureIcon` was removed by the simplification in the same pass.

src/DasIcon/index.test.js
```diff
@@ -0,0 +1,53 @@
...
+    test('renders an img when the server returns a non-SVG content-type (e.g. PNG)', async () => {
+      global.fetch = mockFetch('image/png', '');
+
+      const { container } = render(<DasIcon type="events" iconId="confiscation_rep" />);
+
+      await waitFor(() => {
+        expect(container.querySelector('img')).toBeInTheDocument();
+      });
+      expect(container.querySelector('img').src).toContain('confiscation_rep');
+    });
```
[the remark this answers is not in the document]
jeslefcourt: This is already handled — the suite saves and restores fetch via `jest.spyOn(global, 'fetch')` in `beforeEach` and `fetchSpy.mockRestore()` in `afterEach`, so it doesn't leak into other suites. (The file is now `src/SvgIcon/index.test.js` after the rename.)

src/EventIcon/index.test.js
```diff
@@ -0,0 +1,74 @@
...
+  is_collection: true,
+  event_type: 'incident_collection',
+  contains: containedEvents.map((event) => ({ related_event: event })),
+});
+
+describe('EventIcon collection badge icon', () => {
+  beforeEach(() => {
+    global.fetch = mockSvgFetch();
+  });
+
```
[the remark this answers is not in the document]
jeslefcourt: Already handled — `beforeEach` uses `jest.spyOn(global, 'fetch')` and `afterEach` calls `fetchSpy.mockRestore()`, so fetch is restored between suites.

src/LayerSelectorPopup/index.js
```diff
@@ -48,24 +63,35 @@ const LayerSelectorPopup = ({ data, id }) => {
...
+      const displayTitle = layer.properties.display_title || layer.properties.name || layer.properties.title;
+      const title = isLocallyEdited ? `* ${displayTitle}` : displayTitle;
+
       return <li className={styles.listItem} key={layer.properties.id} onClick={(e) => handleClick(e, layer)}>
         <img
-          alt={layer.properties.display_title || layer.properties.name || layer.properties.title}
+          alt={displayTitle}
           src={imgSrc}
           style={subjectIsStatic(layer) ? { filter: 'brightness(0) opacity(60%)' } : {}}
         />
```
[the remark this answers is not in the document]
jeslefcourt: Gear items now render a dedicated indicator (`<span className={styles.listItemGearIcon} />`) instead of an ``, so there's no missing-source image for gear. This handling came in via the develop merge (93d21749) and is combined here with the event-icon logic, so gear no longer risks a broken ``.

src/ClustersLayer/index.js
```diff
@@ -65,29 +65,59 @@ const ClustersLayer = ({ onShowClusterSelectPopup }) => {
   const { addClusterPolygon, removeClusterPolygon } = useClusterPolygon();
 
   const mapImages = useSelector((state) => state.view.mapImages);
-
+  const locallyEditedEvent = useSelector((state) => state.data.locallyEditedEvent);
+
+  // Use refs so the callback doesn't need to be recreated when these frequently-changing
```
luixlive: nit: Lots of comments added in this file! I like properly documenting complex code with comments, but I think this is probably a bit too much.

src/ClustersLayer/index.js
```diff
@@ -65,29 +65,59 @@ const ClustersLayer = ({ onShowClusterSelectPopup }) => {
...
+  // eslint-disable-next-line react-hooks/refs
+  removeClusterPolygonRef.current = removeClusterPolygon;
+
+  const onShowClusterSelectPopupRef = useRef(onShowClusterSelectPopup);
+  // eslint-disable-next-line react-hooks/refs
+  onShowClusterSelectPopupRef.current = onShowClusterSelectPopup;
+
+  // The callback intentionally reads these values from refs so its identity stays stable
+  // across frequent locallyEditedEvent/mapImages/etc. changes; only `map` is a real dep.
   const updateClusterMarkersCallback = useCallback(async () => {
```
luixlive, line 105: Claude review:
Async `updateClusterMarkersCallback` has no in-flight guard; the new `mapImages` `useEffect` and the `sourcedata` handler run it concurrently, racing on `clusterMarkerHashMapRef`.

src/ClustersLayer/utils.js
luixlive: Claude review:
`getClusterLeaves` callback uses `!error && resolve(features)` — on error the Promise never resolves, hanging `Promise.all` indefinitely.

src/SvgIcon/index.js
```diff
@@ -0,0 +1,207 @@
...
+import { useSelector } from 'react-redux';
+import DOMPurify from 'dompurify';
+import { API_V2_URL, DAS_HOST } from '../constants';
+
+const GENERIC_ICON_ID = 'generic_rep';
+export const svgCache = new Map();
+
+const CONTAINER_SELECTOR = 'svg,g,defs,symbol,marker,clipPath,mask,pattern';
+
+// Identify which containers have fill="none" as a base default before we strip anything.
```
luixlive: nit: Same is in the other file, too many comments here. It's even distracting from the code itself. We should keep comments to the minimum necessarily and make them clear but concise.

src/SvgIcon/index.js
```diff
@@ -0,0 +1,207 @@
...
+// A container's fill="none" should be removed only if it was acting as a default that its
+// children override — either via explicit fill attributes or CSS classes in a <style> block.
+// Removing it unconditionally breaks stroke-based icons where fill="none" is the design intent.
+const collectContainerDefaultFills = (doc) => {
```
luixlive, line 35: Claude review:
`collectContainerDefaultFills` uses child class-attribute presence as a proxy for fill styling, incorrectly marking stroke-only icons for fill removal.

src/SvgIcon/index.js
```diff
@@ -0,0 +1,207 @@
...
+  doc.querySelectorAll('[stroke]:not([stroke="none"])').forEach((el) => {
+    el.setAttribute('stroke', 'currentColor');
+  });
+};
+
+// Strip inline fill/stroke from style attributes (presentation attributes above handle these).
+const stripInlineFillAndStroke = (doc) => {
+  doc.querySelectorAll('[style]').forEach((el) => {
+    const cleaned = el.getAttribute('style')
+      .replace(/\bfill\s*:[^;]*(;?)/g, '')
```
luixlive: Claude review:
`stripInlineFillAndStroke` regex `[^;]*` stops at the first semicolon inside a data URI value, leaving a mangled style attribute.

src/SvgIcon/index.js
```diff
@@ -0,0 +1,207 @@
...
+  rewriteStrokesToCurrentColor(doc);
+  stripInlineFillAndStroke(doc);
+  removeContainerDefaultFills(doc, containersToUnfill);
+
+  return new XMLSerializer().serializeToString(doc.documentElement) || null;
+};
+
+const injectClass = (markup, className) => {
+  if (!className) return markup;
+  if (/<svg[^>]* class="/.test(markup)) {
```
luixlive: Claude review:
`injectClass` replace regex is unanchored and matches the first `<svg class=...>` in the string, which may be an inner nested `<svg>` rather than the root.

src/SvgIcon/index.js
```diff
@@ -0,0 +1,207 @@
...
+      const entry = svgCache.get(src);
+      Promise.resolve().then(() => { if (current) setCached(entry); });
+      return () => { current = false; };
+    }
+
+    const controller = new AbortController();
+    const { signal } = controller;
+
+    const fetchIcon = (url) =>
+      fetch(url, { signal }).then(res => {
```
luixlive: Claude review:
Native `fetch()` omits the `Authorization` header; community icon API endpoint returns 401 and every event icon is blank in community deployments.

src/SvgIcon/index.js
```diff
@@ -0,0 +1,207 @@
...
+  }, [src, fallbackSrc]);
+
+  if (!cached) return null;
+
+  if (cached.imgSrc) {
+    return <img alt="" className={className} src={cached.imgSrc} style={style} {...rest} />;
+  }
+
+  return (
+    <span
```
luixlive, line 244: Claude review:
`title` prop is spread onto the `<span>` wrapper, not into the inlined SVG — icon names are invisible to screen readers and browser tooltips.

src/SvgIcon/index.js
```diff
@@ -0,0 +1,207 @@
...
+
+const SvgIcon = ({ type, iconId, imageUrl, className, color, style, ...rest }) => {
+  const communityValue = useSelector((state) => state.data.community?.value);
+
+  const onImgError = (event) => {
+    event.currentTarget.style.display = 'none';
+  };
+
+  if (type === 'subjects') {
+    return <img alt={`${type} icon`} className={className} onError={onImgError} src={imageUrl} style={style} {...rest} />;
```
luixlive: Claude review:
`alt` text `'subjects icon'` is non-descriptive and not internationalized — every subject icon is announced identically by screen readers.

src/SvgIcon/index.js
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  const effectiveIconId = iconId || GENERIC_ICON_ID;
+  const isGeneric = effectiveIconId.includes('generic');
+
+  let iconSrc;
+  if (communityValue && type === 'events') {
```
luixlive: Claude review:
All patrol-icon callers pass `type='events'`; in community mode this routes patrol icon IDs to the events/eventtypes/icons endpoint, which does not serve patrol-type icons.

src/SvgIcon/index.js
```diff
@@ -0,0 +1,207 @@
...
+
+  // Map legacy color prop and style.fill to CSS color so fill:currentColor in the SVG inherits it.
+  const { fill: styleFill, ...restStyle } = style || {};
+  const effectiveColor = color || styleFill;
+  const svgStyle = effectiveColor ? { ...restStyle, color: effectiveColor } : restStyle;
+
+  return (
+    <InlineSvg
+      className={`${className || ''} ${isGeneric ? 'generic' : ''}`.trim()}
+      fallbackSrc={`${DAS_HOST}/static/sprite-src/${GENERIC_ICON_ID}.svg`}
```
luixlive: Claude review:
`fallbackSrc` hardcodes DAS_HOST/static/sprite-src — the same unauthenticated fetch path — so the fallback silently fails for any deployment where that path requires auth.

luixlive (approved):
Added a couple nits about too much comments, plus a few comments I found interesting from a review I ran locally with Claude. Didn't validate all of them since that would take me more time, but pretty sure you can feed those to your agent and see if it finds any possible improvement to commit before merging 👍 

From my end, approved.

src/ClustersLayer/utils.js
jeslefcourt: `calcSvgImageIconId` was lifted into `src/utils/mapImages.js` earlier (b10b38e2), and after merging develop it now carries develop's corrected key order. On removing `MapImageFromSvgSpriteRenderer` entirely: develop just invested two production hotfixes in that component (#1643, #1644 — icon fallbacks and cluster icon readiness), which this branch has now merged in. Removing it in this PR would mean re-implementing those brand-new fixes, so I'm keeping the full removal as a follow-up ticket. Resolving to keep this PR scoped.

src/SvgIcon/index.js
```diff
@@ -0,0 +1,207 @@
...
+          setCached(svgCache.get(fallbackSrc));
+          return;
+        }
+        fetchIcon(fallbackSrc)
+          .then(entry => {
+            svgCache.set(fallbackSrc, entry);
+            if (current) setCached(entry);
+          })
+          .catch(() => {});
+      });
```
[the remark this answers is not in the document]
jeslefcourt: Fixed in 60ff7b6b — when the fallback resolves, its entry is now also cached under the failing `src` key, so subsequent mounts skip the dead URL. Trade-off (a transient failure is pinned until reload) is noted in a comment. Test added.

src/utils/mapImages.js
```diff
@@ -0,0 +1,11 @@
...
+  let string = `${icon_id}`;
+
+  [priority, height, width]
+    .filter(item => item === 0 || !!item)
+    .forEach((item) => {
+      string+=`-${item}`;
+    });
+
+  return string;
+};
```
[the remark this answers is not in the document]
jeslefcourt: Superseded by the develop merge (216c4bc0): `calcSvgImageIconId` is now develop's join-based implementation (no `string+=`, no lint violation), and all callers guard `icon_id` before building keys, so no `"undefined"` IDs are produced.

src/ClustersLayer/utils.js
```diff
@@ -142,11 +148,26 @@ export const getRenderedClustersData = async (clustersSource, map) => {
...
-    )
+    (clusterFeatures) => hashCode(clusterFeatures.map((clusterFeature) => {
+      const isLocallyEdited = locallyEditedEvent?.id === clusterFeature.properties.id;
+      const priority = clusterFeature.properties.priority ?? 0;
+      // Include whether the icon image is loaded so the hash changes when mapImages gains the entry,
+      // forcing the marker to be recreated with the correct icon.
+      const iconKey = calcSvgImageIconId({
+        icon_id: clusterFeature.properties.icon_id,
+        priority: clusterFeature.properties.priority,
+      });
```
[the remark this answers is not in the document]
jeslefcourt: Obsolete after the develop merge (216c4bc0): the `iconKey`/`imageLoaded` mechanism was removed entirely, superseded by develop's `iconsReady` gating from #1644.

src/ClustersLayer/index.js
jeslefcourt: Trimmed in 60ff7b6b — the multi-line ref-pattern explanations are now one-liners; each `eslint-disable` keeps a short single-line justification.

src/ClustersLayer/index.js
jeslefcourt: Addressed by merging develop (216c4bc0): the callback now has develop's `latestClusterUpdateRunIdRef` stale-run guard — overlapping runs are expected, and any run that finishes after a newer one started discards its results instead of racing on `clusterMarkerHashMapRef`.

src/ClustersLayer/utils.js
jeslefcourt: Fixed in 60ff7b6b — the callback now resolves `[]` on error instead of never resolving, and empty clusters are filtered out afterwards (filtering the parallel ids/features arrays together and deriving hashes from the filtered set to keep them index-aligned). Note this pattern also ships in develop's current version, so this fixes production behavior too. Test added for the error path.

src/SvgIcon/index.js
jeslefcourt: Trimmed in 60ff7b6b — comments rewritten as concise one-liners; only the non-obvious constraints (e.g. sanitize ordering) kept.

src/SvgIcon/index.js
jeslefcourt: Fixed in 60ff7b6b — a child class only counts as fill evidence when some `<style>` block in the document actually declares `fill:`. Stroke-only icons with classed children no longer get their `fill="none"` stripped.

src/SvgIcon/index.js
jeslefcourt: Fixed in 60ff7b6b — replaced the regex with `el.style.removeProperty('fill'/'stroke')` (removing the `style` attribute when left empty), so data-URI values with embedded semicolons survive intact.

src/SvgIcon/index.js
jeslefcourt: Fixed in 60ff7b6b — `injectClass` now isolates the first `<svg …>` root tag and merges/adds the class only within it; nested `<svg>` elements are untouched.

src/SvgIcon/index.js
jeslefcourt: Fixed in 60ff7b6b — icon fetches go through the global axios instance (fetch adapter), so the RequestConfigManager interceptors attach the bearer token and USER-PROFILE header. Community icon endpoints now authenticate. Tests migrated to axios mocks.

src/SvgIcon/index.js
jeslefcourt: Fixed in 60ff7b6b — when a `title` is provided the wrapper renders `role="img"` + `aria-label` (title kept for mouse tooltips); without one it renders `aria-hidden="true"` so decorative icons are skipped. The `` branch uses `alt={title || ''}`.

src/SvgIcon/index.js
jeslefcourt: Fixed in 60ff7b6b — the literal `${type} icon` alt is gone; the subjects branch now uses `alt={title || ''}` (decorative when no accessible name is supplied). No live component call sites pass `type="subjects"` today, so no new i18n strings were needed.

src/SvgIcon/index.js
jeslefcourt: Fixed in 60ff7b6b — all patrol call sites (PatrolFilter FiltersPopover ×2, PatrolDetailView Header, PatrolListItem, PatrolTrackLegend) now pass `type="patrols"`, which routes to the sprite-src path in both community and standard modes instead of the events icon endpoint.

src/SvgIcon/index.js
jeslefcourt: Fixed in 60ff7b6b — sprite URLs are built by a shared community-aware helper, so the generic fallback uses `community/{value}/static/sprite-src/...` when a community is active and `DAS_HOST/static/sprite-src/...` otherwise.

src/ClustersLayer/utils.js
jeslefcourt: Update: we went ahead and did the full removal in this PR after all (e1336bed). `MapImageFromSvgSpriteRenderer` is gone — event map icons are now generated lazily by a plain registry module (`src/utils/eventMapIcons.js`) hooked into Mapbox's `styleimagemissing` event, so GL layers request icons on demand and re-render automatically when they resolve. Cluster HTML markers and the layer-selector popup read the same registry (with a subscription for refresh), and the develop hotfix semantics (`_rep` retry, 4xx fallback chain, transient retry, dedup) are preserved. `state.view.mapImages`/`MapImagesLayer` remain only for the subject and static-sensor pipelines. No follow-up ticket needed.

src/utils/mapImages.js
```diff
@@ -0,0 +1,14 @@
...
+export const EVENT_ICON_ID_PREFIX = 'event-icon|';
+
+// Builds the map-image cache key for an event's icon. Must match the format
+// emitted by the Mapbox icon-image expressions: the `event-icon|` prefix
+// followed by icon_id, priority, width, and height, pipe-separated, with absent
+// parts omitted (e.g. "event-icon|fire|200|24|32").
+export const calcSvgImageIconId = ({ icon_id, priority, width, height }) => {
+  const parts = [icon_id, priority, width, height].filter((value) => value === 0 || Boolean(value));
+  return `${EVENT_ICON_ID_PREFIX}${parts.join('|')}`;
+};
```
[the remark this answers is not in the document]
jeslefcourt: Verified but intentionally not changing the code: the reverse-parser is only fed by styleimagemissing ids, and all three GL icon-image expressions always emit the icon_id slot (a missing property yields `event-icon||200`, which parses to an empty icon_id and falls back to `generic` at fetch time). The JS callers that could omit icon_id either guard on it before building keys or only use the key for cache lookups, where a miss falls through to the image-URL fallback. Defaulting icon_id inside `calcSvgImageIconId` alone would desync JS-built keys from expression-built ids and break resolution — it would have to be paired with `coalesce` defaults in all three expressions. Documented the invariant on the function instead (a49f3270).

src/utils/eventMapIcons.js
```diff
@@ -0,0 +1,242 @@
...
+  svgEl.style.fill = `${color} !important`;
+  svgEl.setAttribute('fill', color);
+
+  svgEl.querySelectorAll('*').forEach((node) => {
+    ['class', 'style', 'fill', 'stroke'].forEach((attribute) => node.removeAttribute(attribute));
+  });
+
+  const serializedSvg = new XMLSerializer()
+    .serializeToString(svgEl)
+    .replace(/<style>.*?<\/style>/g, '');
```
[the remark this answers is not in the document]
jeslefcourt: Fixed in a49f3270 — `<style>` elements are now removed from the parsed SVG DOM before serializing (mirroring SvgIcon's `removeStyleBlocks`), and the regex is gone. This also covers `<style type="text/css">`, which the regex missed too. Regression test added with a multi-line attribute-bearing style block. Worth noting the regex was inherited verbatim from the old `MapImageFromSvgSpriteRenderer` on develop, so this fixes a latent production issue as well.

src/SvgIcon/index.js
```diff
@@ -0,0 +1,278 @@
...
+  const [cached, setCached] = useState(() => svgCache.get(src) ?? null);
+  const [imgFailed, setImgFailed] = useState(false);
+  const [renderedSrc, setRenderedSrc] = useState(src);
+
+  // Reset from the cache (or null) when src changes, so a stale icon never keeps rendering.
+  if (renderedSrc !== src) {
+    setRenderedSrc(src);
+    setCached(svgCache.get(src) ?? null);
+    setImgFailed(false);
+  }
```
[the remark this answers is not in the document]
jeslefcourt: This is React's documented "adjusting state when a prop changes" pattern (conditional, same-component setState during render): React re-renders immediately before commit, the `renderedSrc !== src` guard self-clears so no loop is possible, and it's StrictMode-safe. An effect-based reset would commit a frame with the stale icon first and also trips this repo's `react-hooks/set-state-in-effect` lint rule; a `key` remount would discard the whole subtree unnecessarily. Two independent review passes verified the implementation. Keeping as-is.

src/utils/mapImages.js
```diff
@@ -0,0 +1,17 @@
...
+// emitted by the Mapbox icon-image expressions: the `event-icon|` prefix
+// followed by icon_id, priority, width, and height, pipe-separated, with absent
+// parts omitted (e.g. "event-icon|fire|200|24|32").
+// Invariant: callers must pass an `icon_id` (or guard before calling) — absent
+// leading parts make the pipe-format non-reversible, and the GL icon-image
+// expressions always emit the icon_id slot.
+export const calcSvgImageIconId = ({ icon_id, priority, width, height }) => {
+  const parts = [icon_id, priority, width, height].filter((value) => value === 0 || Boolean(value));
+  return `${EVENT_ICON_ID_PREFIX}${parts.join('|')}`;
+};
```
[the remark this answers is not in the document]
jeslefcourt: Same rationale as the earlier thread on this function: defaulting `icon_id` inside `calcSvgImageIconId` alone would desynchronize JS-built keys from the GL icon-image expression output (`event-icon|generic|200` vs `event-icon||200`) and break icon resolution unless all three expressions gain matching coalesce defaults. No live caller produces the malformed shape (callers guard `icon_id`; the styleimagemissing parser is only fed expression-built ids, which always emit the slot), and the handler now registers images under the exact requested id, so even an unexpected id shape resolves. The invariant is documented on the function. Declining the change.

PR #1533, opened 2026-04-13
Add download patrol track

jeslefcourt:
## What does this PR do?

This pull request adds a new "Download Patrol Track" feature to the patrol options menu, allowing users to download the patrol track as a GeoJSON file. The feature is implemented with localization support across multiple languages and integrates with analytics tracking.

**Feature: Download Patrol Track**

* Added a new "Download Patrol Track" option to the `KebabMenu` in the `PatrolMenu` component, which allows users to download the patrol track as a `.geojson` file for the patrol leader. The download uses the patrol's serial number and leader's name for the filename, and includes analytics tracking. (`src/PatrolMenu/index.js`) [[1]](diffhunk://#diff-b48b4cf86b2cbe86d2d82f414cb0ca130c293649e71a3b16dfc17722e4e33de0R106-R118) [[2]](diffhunk://#diff-b48b4cf86b2cbe86d2d82f414cb0ca130c293649e71a3b16dfc17722e4e33de0R164-R170)
* Imported the necessary `DownloadArrowIcon`, `TRACKS_API_URL`, and `downloadFileFromUrl` utilities to support the download functionality. (`src/PatrolMenu/index.js`) [[1]](diffhunk://#diff-b48b4cf86b2cbe86d2d82f414cb0ca130c293649e71a3b16dfc17722e4e33de0R5) [[2]](diffhunk://#diff-b48b4cf86b2cbe86d2d82f414cb0ca130c293649e71a3b16dfc17722e4e33de0R14-R19)

**Localization Updates**

* Added the "downloadTrackButton" translation key for English, Spanish, French, Nepali, Portuguese, and Swahili locales to support the new menu option. (`public/locales/en-US/patrols.json`, `public/locales/es/patrols.json`, `public/locales/fr/patrols.json`, `public/locales/ne-NP/patrols.json`, `public/locales/pt/patrols.json`, `public/locales/sw/patrols.json`) [[1]](diffhunk://#diff-b60f03a7462c05e307e607ac2f6b4d8caadadfb1d2575450ad087111a8ad7092R67) [[2]](diffhunk://#diff-a44df95d00640a065bd7e8a03cdf6f5c93e3463788b72546dfc45b8c257192cbR65) [[3]](diffhunk://#diff-6a70ca17864c5f89ca48f0ba9ba98883af41efb70ce7fc34ebc1ec5f33ab64aaR65) [[4]](diffhunk://#diff-e3c0b49f7b98c23d2ce4c11ebe5838a03e6e5922194364dd42a33d69c9da74c2R65) [[5]](diffhunk://#diff-a9bb5871e460c6fe3a8b93f027d43c8a066b4f5b83f23cc09c26265e594c4836R65) [[6]](diffhunk://#diff-2f4efd3b56b0f8757f06ada4c04811c5197dbb833831958e5beb36f3d8bdd2afR67)
### Relevant link(s)
- [ERA-13106](https://allenai.atlassian.net/browse/ERA-13106)
- [Branch Environment](https://era-13106.dev.pamdas.org)

## Notes

[ERA-13106]: https://allenai.atlassian.net/browse/ERA-13106?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/public/locales/ne-NP/patrols.json
+++ b/public/locales/ne-NP/patrols.json
@@ -62,6 +62,7 @@
         "copyButton": "गस्तीको लिकं कपि गर्नुहोस्",
         "copyButtonMessage": "लिकं कपि भयो",
         "endPatrol": "गस्ती अन्त्य गर्नुहोस्",
+        "downloadTrackButton": "गस्तीको ट्र्याक डाउनलोड गर्नुहोस्",
         "printPatrolButton": "गस्ती प्रिन्ट गर्नुहोस्",
         "restorePatrol": "गस्ती पुर्नस्थापना गर्नुहोस्",
         "startPatrol": "गस्ती सुरु गर्नुहोस्",
--- a/public/locales/ne-NP/patrols.json
+++ b/public/locales/ne-NP/patrols.json
@@ -1,85 +1,86 @@
 {
-    "addToPatrolModal": {
-        "title": "गस्तीमा थप्नुहोस्",
-        "cancel": "रद्द गर्नुहोस्",
-        "noMorePatrols": "प्रदर्शनीका लागि कुनै गस्ती छैन ।",
-        "loading": "लोड हुँदैछ…"
+  "addToPatrolModal": {
+    "title": "गस्तीमा थप्नुहोस्",
+    "cancel": "रद्द गर्नुहोस्",
+    "noMorePatrols": "प्रदर्शनीका लागि कुनै गस्ती छैन ।",
+    "loading": "लोड हुँदैछ…"
+  },
+  "detailView": {
+    "addEmptyNoteAlert": "नयाँ नोट सुरक्षित गर्न सकिएन: पुरानै नोट सुरक्षित हुन बाँकी छ",
+    "formButtons": {
+      "save": "सुरक्षित गर्नुहोस्",
+      "cancel": "रद्द गर्नुहोस्"
     },
-    "detailView": {
-        "addEmptyNoteAlert": "नयाँ नोट सुरक्षित गर्न सकिएन: पुरानै नोट सुरक्षित हुन बाँकी छ",
-        "formButtons": {
-            "save": "सुरक्षित गर्नुहोस्",
-            "cancel": "रद्द गर्नुहोस्"
-        },
-        "header": {
-            "patrolSchedule": "{{scheduledStartTime}} निर्धारण गरिएको",
-            "patrolStartButton": "सुरु",
-            "patrolRestoreButton": "पुर्नस्थापना",
-            "uiStateTitles": {
-                "active": "सकृय",
-                "cancelled": "रद्द गरियो",
-                "done": "भयो",
-                "invalid": "स्वरुप मिलेन",
-                "readyToStart": "सुरु गर्न तयार अवस्थामा",
-                "scheduled": "निर्धारित",
-                "startOverdue": "बाँकी रहेको काम सुरु गर्नुहोस्"
-            }
-        },
-        "loadingMessage": "सुरक्षित हुँदैछ…",
-        "planSection": {
-            "autoStartCheckboxLargeLabel": "यसै समय अर्थरेन्जरमा स्वतः गस्ती सुरु गर्नुहोस्",
-            "autoStartCheckboxSmallLabel": "अर्थरेन्जरमा स्वतः गस्ती सुरु गर्नुहोस्",
-            "autoEndCheckboxLargeLabel": "यसै समय अर्थरेन्जरमा स्वतः गस्ती अन्त्य गर्नुहोस्",
-            "autoEndCheckboxSmallLabel": "अर्थरेन्जरमा स्वतः गस्ती अन्त्य गर्नुहोस्",
-            "endDateLabel": "अन्त्य मिति",
-            "endTimeLabel": "अन्त्य समय",
-            "endLocationLargeLabel": "अन्त्य स्थान",
-            "endLocationSmallLabel": "स्थान",
-            "locationSelectorPlaceholder": "स्थान सेट गर्नुहोस्",
-            "objectiveLabel": "उद्देश्य",
-            "objectivePlaceholder": "गस्तीको उद्देश्य वर्णन गर्नुहोस्…",
-            "startDateLabel": "सुरु मिति",
-            "startTimeLabel": "सुरु समय",
-            "startLocationLargeLabel": "सुरु स्थान",
-            "startLocationSmallLabel": "स्थान",
-            "title": "योजना",
-            "trackedByLabel": "द्वारा ट्रयाक गरिएको",
-            "trackedByPlaceholder": "उपकरण छान्नुहोस्…"
-        },
-        "quickLinksTitles": {
-            "activity": "गतिविधि",
-            "history": "इतिहास",
-            "plan": "योजना"
-        }
+    "header": {
+      "patrolSchedule": "{{scheduledStartTime}} निर्धारण गरिएको",
+      "patrolStartButton": "सुरु",
+      "patrolRestoreButton": "पुर्नस्थापना",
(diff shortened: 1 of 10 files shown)


src/PatrolMenu/index.js
```diff
@@ -100,6 +103,19 @@ const PatrolMenu = ({
     }
   }, [canEnd, onPatrolChange, patrolStartStopTitle]);
 
+  const patrolLeader = patrol.patrol_segments[0]?.leader;
```
luixlive: Extremely nit: we usually keep the code inside components body sorted more or less like this:
1- Hooks, store selectors, refs, state variables
2- Calculated and memoized variables
3- Callbacks
4- Component rendering (`return`)

It's silly feedback, definitely not a blocker, it just keeps the code organized and easy to scan. That said, these two variables `patrolLeader` and `patrolTimeRange` should be placed below line 43 `const { hasPatrolsUpdatePermission } = usePatrolsPermissions();`.

luixlive (commented):
Translations seem fine. The main changes in `PatrolMenu` are minimal and elegant. No blocker. Just before approving, I have a couple things for you to think about.

1- Should we include a new button in our "Exports" section of the hamburger menu?
Image

That would definitely increase the scope since we may need to add a modal, to select the patrol and the filter dates. Not sure if that makes sense.

2- The only condition we have now to show or not the new "Download" button is if the patrol has a leader, even if the leader has no tracks within the patrol bounds. This means that we can download GeoJSON files with no meaningful data. I'd suggest to improve the current condition to only allow downloading if there are tracks, what do you think?
Image

public/locales/ne-NP/patrols.json
```diff
@@ -1,84 +1,86 @@
 {
-    "addToPatrolModal": {
-        "title": "गस्तीमा थप्नुहोस्",
-        "cancel": "रद्द गर्नुहोस्",
-        "noMorePatrols": "प्रदर्शनीका लागि कुनै गस्ती छैन ।",
-        "loading": "लोड हुँदैछ…"
+  "addToPatrolModal": {
```
luixlive, line 2: This and the following translation files have lots of changes. I think the agent just matched the indentation of the others. But worth making sure that these texts still translate correctly.

src/PatrolMenu/index.js
```diff
@@ -38,6 +42,19 @@ const PatrolMenu = ({
   const patrolState = calcPatrolState(patrol);
 
   const { hasPatrolsUpdatePermission } = usePatrolsPermissions();
+  const patrolLeader = patrol.patrol_segments[0]?.leader;
+  const patrolTimeRange = patrol.patrol_segments[0]?.time_range;
+  const patrolLeaderTrackHasPoints = useSelector((state) => {
```
luixlive: Redux selectors run every time there is a store update, even if it's totally unrelated to what is being read here. For example, if our socket (real time) sends updates of new event types in the feed, this code-block would be executed. This of course can affect performance. In order to preserve memoization and avoid unnecessary processing the selector body should stay as trivial as possible.

This one is actually pretty simple to optimize. All we are really reading from the store is the patrol leader track times. The rest of the processing can stay outside of the selector. Here a quick refactorization that keeps the selector trivial and stable, and the code block very easy to read:
```Javascript
const patrolLeaderTrackTimes = useSelector((state) =>
  state.data.tracks[patrolLeader?.id]?.track?.features?.[0]?.properties?.coordinateProperties?.times);

const doesPatrolLeaderHaveTracksWithinPatrolTimeRange = useMemo(() => {
  if (patrolLeaderTrackTimes?.length > 0) {
    const patrolStartTimeStamp = patrolTimeRange?.start_time
      ? new Date(patrolTimeRange.start_time).getTime()
      : null;
    const patrolEndTimeStamp = patrolTimeRange?.end_time
      ? new Date(patrolTimeRange.end_time).getTime()
      : null;
    if (!patrolStartTimeStamp && !patrolEndTimeStamp) {
      // Patrol time range is not bounded and the leader has tracks.
      return true;
    }

    // If the patrol leader has any tracks within the patrol time range, return true.
    return patrolLeaderTrackTimes.some((trackTime) => {
      const trackTimeStamp = new Date(trackTime).getTime();
      return (!patrolStartTimeStamp || trackTimeStamp >= patrolStartTimeStamp)
        && (!patrolEndTimeStamp || trackTimeStamp <= patrolEndTimeStamp);
    });
  }

  return false;
}, [patrolLeaderTrackTimes, patrolTimeRange]);
```

src/PatrolMenu/index.js
```diff
@@ -145,6 +172,16 @@ const PatrolMenu = ({
         {t('printPatrolButton')}
       </KebabMenu.Option>
     }
+
+    <span title={(!patrolLeader || !patrolLeaderTrackHasPoints) ? t('noTrackDataTooltip') : ''}>
```
luixlive: I see that we decided to move from conditionally rendering the button to keep it rendered and just disable it. I'm ok with this approach, and actually feel it's a better UX. I just add this comment because that's not the approach we follow in the other `KebabMenu.Option`s above. I'm ok keeping this new approach, but if we want to stay consistent, we could also move back to the conditional render.

src/PatrolMenu/index.test.js
```diff
@@ -133,6 +140,66 @@ describe('PatrolMenu', () => {
     });
   });
 
+  describe('Download Patrol Track button', () => {
```
luixlive, line 148: Nice! 💯

luixlive (approved):
Approved 🎉  

I still added a couple recommendations worth taking a look at. If they make sense and you have 5 minutes, go for them! But this is ready to be merged.

src/PatrolMenu/index.js
```diff
@@ -38,6 +44,32 @@ const PatrolMenu = ({
...
+      : null;
+
+    if (!patrolStartTimestamp && !patrolEndTimestamp) return true;
+
+    // Track times are sorted newest-first; check range overlap using only the first/last entries
+    const trackNewest = new Date(patrolLeaderTrackTimes[0]).getTime();
+    const trackOldest = new Date(patrolLeaderTrackTimes[patrolLeaderTrackTimes.length - 1]).getTime();
+
+    return (!patrolStartTimestamp || trackNewest >= patrolStartTimestamp)
+      && (!patrolEndTimestamp || trackOldest <= patrolEndTimestamp);
```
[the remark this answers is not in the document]
jeslefcourt: Patrol tracks don't come from start/stop - This is okay

src/PatrolMenu/index.js
```diff
@@ -10,10 +14,12 @@ import { ReactComponent as CloseIcon } from '../common/images/icons/close-icon.s
 import { ReactComponent as RestoreIcon } from '../common/images/icons/restore.svg';
 
 import { DAS_HOST, PATROL_UI_STATES, PATROL_API_STATES } from '../constants';
+import { TRACKS_API_URL } from '../ducks/tracks';
 import { usePatrolsPermissions } from '../hooks/usePermissions';
```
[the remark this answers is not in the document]
jeslefcourt: This is a common pattern through the project.

src/PatrolMenu/index.js
```diff
@@ -145,6 +195,24 @@ const PatrolMenu = ({
...
     }
+
+    {isDownloadDisabled
+      ? <OverlayTrigger
+          placement="top"
+          overlay={<Tooltip id="download-track-tooltip">{t('noTrackDataTooltip')}</Tooltip>}
+        >
+        <span>
+          <KebabMenu.Option disabled onClick={handleDownloadTrack}>
+            <DownloadArrowIcon data-testid="download-arrow-icon" />
```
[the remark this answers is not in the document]
jeslefcourt: This is okay.

PR #1536, opened 2026-04-16
Decouple Auth0 config from build-time env files

StephenWithPH:
## Summary

Auth0 credentials (domain, client ID, audience) were previously baked into the JS bundle at build time via `.env.production` / `.env.development`. This made it impossible to deploy the same Docker image to multiple environments with different Auth0 tenants — staging built from a `release-*` branch always got production Auth0 values.

This PR introduces a runtime configuration layer (`src/auth0-config.js`) that resolves Auth0 values via a priority chain:

1. `window.__AUTH0_CONFIG__` — set by a Kubernetes ConfigMap mounted at `/config.js`
2. Hard-coded production defaults in `src/auth0-config.js`

## Approach: "fail open" to production

The key design decision is that **production Auth0 values are the hard-coded defaults**. If no ConfigMap is present (i.e. `/config.js` returns a 404), the app silently falls back to production credentials. This means:

- **Production requires zero configuration** — no ConfigMap needed, no new failure modes, no flicker. The `<script src="/config.js">` tag gets a 404 (a no-op), and the app uses the baked-in defaults.
- **Non-production environments opt in** by mounting a ConfigMap with dev/staging Auth0 values.
- **If someone forgets the ConfigMap on staging**, the app points at the production Auth0 tenant. Auth0's own allowed-origins configuration will reject the login attempt from a non-production URL, so this fails safely rather than silently misbehaving.

## Trade-offs considered

| Approach | Pros | Cons |
|---|---|---|
| **This PR: `config.js` via ConfigMap** | No Dockerfile changes, no entrypoint script, no blocking network request, no flicker. Single authoritative source for Auth0 config. | Requires K8s ConfigMap per non-prod environment. Developers must copy `public/config.js.example` for local dev. |
| Docker entrypoint generating `config.js` | Self-contained in the image, no K8s config needed | More moving parts (entrypoint script, Dockerfile changes, `chmod`). Harder to debug. |
| Fetch from `/api/v1.0/status` endpoint | Backend is single source of truth | Chicken-and-egg: Auth0Provider must be configured before any API call. Adds blocking network request before first render. App can't load at all if backend is down. |
| Build separate images per environment | No app changes at all | Longer CI, image proliferation, still baking config at build time. |
| Nginx `sub_filter` replacement | No app code changes | Fragile with minified JS, breaks content hashes, adds per-response overhead. |

## What changed

- **`src/auth0-config.js`** (new) — single source of truth for Auth0 config, resolves `window.__AUTH0_CONFIG__` then production defaults
- **`public/config.js.example`** (new) — template for local development with dev Auth0 values
- **`index.html`** — loads `/config.js` synchronously before the app bundle
- **`src/index.js`**, **`src/Login/index.js`**, **`src/Auth0TokenManager/index.js`** — import from `auth0-config` instead of `import.meta.env`
- **`.env.development`**, **`.env.production`** — `REACT_APP_AUTH0_*` vars removed entirely
- **`.gitignore`** — ignores `public/config.js`

## Infra follow-up (not in this PR)

Each non-production environment needs a ConfigMap mounted at `/usr/share/nginx/html/config.js`:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: das-web-auth0-config
data:
  config.js: |
    window.__AUTH0_CONFIG__ = {
      domain: 'auth-dev.pamdas.org',
      clientId: 'hLoPCTgBCrlLAVzqg74YSmOftaSfb5Uf',
      audience: 'https://dev.pamdas.org/api',
    };
```

## Local development

```sh
cp public/config.js.example public/config.js
yarn start
```


public/config.js.example
```diff
@@ -0,0 +1,10 @@
+// Copy this file to public/config.js for local development.
+// public/config.js is gitignored and will not be committed.
+//
+// Without this file, the app uses production Auth0 values
+// from src/auth0-config.js.
+window.__AUTH0_CONFIG__ = {
```
luixlive, line 6: Right now the configuration you want to inject into the app is exclusively about Auth0, so this name makes sense. But since we are implementing a pattern that will probably scale to other areas of the app in the future, let's make the entire implementation more generic (agnostic) of the technologies it will configure.

I'd rename this variable to something like `window.__APP_CONFIG__`. And the comments above too: "Without this file, the app used production values from src/config.js.".

src/Auth0TokenManager/index.js
```diff
@@ -4,6 +4,7 @@ import { useDispatch, useSelector } from 'react-redux';
 import { useLocation } from 'react-router';
 import { POST_AUTH_SUCCESS } from '../ducks/auth';
 import useNavigate from '../hooks/useNavigate';
+import auth0Config from '../auth0-config';
```
luixlive, line 7: Same thing applies, let's rename this file to "config.js" or "app-config.js", and we can keep the configurations clean by using objects per subject like:
```Javascript
import appConfig from '../app-config';

const token = await getAccessTokenSilently({
  authorizationParams: {
    audience: appConfig.auth0.audience,
  },
});
```

src/Auth0TokenManager/index.test.js
```diff
@@ -74,7 +74,7 @@ describe('Auth0TokenManager', () => {
       await waitFor(() => {
         expect(mockGetAccessTokenSilently).toHaveBeenCalledWith({
           authorizationParams: {
-            audience: undefined, // process.env.REACT_APP_AUTH0_AUDIENCE
+            audience: 'https://pamdas.org/api',
```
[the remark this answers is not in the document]
luixlive: Mocking the app config sounds great. We can even do it in `src/setupTests.js` so they keep the same mocked values while the tests execute.

src/auth0-config.js
```diff
@@ -0,0 +1,22 @@
+// Auth0 configuration that varies by deployment environment.
```
luixlive, line 1: As it was already mentioned, let's rename this and update its content so it's a global config 👍 Claude provided this example updated:
```Javascript
// Runtime configuration injected via /config.js
// Production values are the hard-coded defaults.
// Non-production environments override via window.__APP_CONFIG__.

const PRODUCTION_DEFAULTS = {
  auth0: {
    domain: 'auth.pamdas.org',
    clientId: 'FHoeQpdko5EMFU8JjjCzjPWT7k1sqm20',
    audience: 'https://pamdas.org/api',
  },
};

const overrides = window.__APP_CONFIG__ ?? {};

const config = {
  auth0: {
    ...PRODUCTION_DEFAULTS.auth0,
    ...overrides.auth0,
  },
};

export default config;
```

index.html
```diff
@@ -18,6 +18,7 @@
     You need to enable JavaScript to run this app.
   </noscript>
   <div id="root"></div>
+  <script src="/config.js"></script>
```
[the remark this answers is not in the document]
luixlive: I think we don't want to add defer since a file so small should load extremely fast, and we want the configuration to be available ASAP.

luixlive (changes requested):
Addressing the feedback, I'd say we are ready to remove the draft tag and actually test this implementation 👍

StephenWithPH:
superseded by #1542

PR #1548, opened 2026-04-22
ERA-13041: Support nested property names 

luixlive:
## What does this PR do?
Update `transformSchemaToFormElements` utility and `SchemaForm` component to properly handle schemas with the new convention of field ids in the `ui` object in order to support fields with same names when placed in different JSON schema objects.

## Evidence
The branch environment contains an event type with several fields with the same name called "Same field names in different JSON schema objects supported".

### Relevant link(s)
- [ERA-13041](https://allenai.atlassian.net/browse/ERA-13041)
- [Branch Environment](https://era-13041.dev.pamdas.org)

## Notes
There has been a long time semantic issue I've been slowly trying to fix in both repos: using "fields" instead of "form elements". Not all form elements are fields, for example sections and headers. I incorrectly encompassed all elements as fields in several places, and I've been slowly correcting it. There were still a bunch of places in this repo where the fix had to be done, and I did it in this PR. That's why you see 81 file changes. A bunch of them have nothing to review, they are just variable or file name updates.

[ERA-13041]: https://allenai.atlassian.net/browse/ERA-13041?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/ReportManager/DetailsSection/SchemaForm/fields/Collection/SortableList/Item/FormModal/index.js
```diff
@@ -72,28 +72,36 @@ const FormModal = ({
...
-          {leftColumn.map((fieldId) => renderField(
-            fieldId,
-            formData[fieldId],
-            onFieldChange,
-            errors?.[fieldId],
-            focusLocationMarker,
-            [...breadcrumbs, { display: title, id: fieldId }]
-          ))}
+          {leftColumn.map((leftColumnChildId) => {
+            const leftColumnChildName = leftColumnChildId.split('.').pop();
```
luixlive: `formData` and `errors` follow the JSON schema structure, so we need to read from them by field name, not id:
```Javascript
// Example formData
const formData = {
  name: 'Luis',
  children: [{
    name: 'Luis II',
  }],
};

// Example errors
const errors: {
  children: {
    0: {
      name: {
        message: 'This field is required',
      },
    },
    message: 'Some items of this collection have errors in their inner forms.',
  },
};
```
In that example the field id of the inner "name" field would be `children.name`, but it's field name would be `name`, which we can easily get with:
```Javascript
// "children.name" --> "name"
const fieldName = fieldId.split('.').pop();
```

You'll see that pattern in a few places of this PR.

src/ReportManager/DetailsSection/SchemaForm/formElements/Collection/SortableList/Item/FormModal/index.js
```diff
@@ -23,7 +23,7 @@ const FormModal = ({
   onDone,
   onFieldChange,
   readOnly,
-  renderField,
+  renderFormElement,
```
luixlive, line 27: `Collection/SortabaleList/Item/FormModal` is the modal that renders the inner form of a collection item:
Image

It's internals needed to be updated to call the render prop `renderFormElement` for each collection item form field, properly distinguishing between the field name and field id.

src/ReportManager/DetailsSection/SchemaForm/formElements/Collection/SortableList/Item/FormPreview/index.js
```diff
@@ -34,40 +34,44 @@ const FormPreview = ({
       className={`${styles.formPreview} ${isDragOverlay ? styles.dragOverlay : ''} ${hasError ? styles.error : ''}`}
       data-testid="schema-form-collection-item-form-preview"
     >
-    {fieldIds.map((fieldId) => <li className={styles.fieldSummary} key={fieldId}>
-      <div>
-        <p className={`${styles.label} ${errors?.[fieldId] ? styles.error : ''}`}>
-          {formElements[fieldId].details.label}
-        </p>
+    {fieldIds.map((fieldId) => {
```
luixlive, line 38: `Collection/SortableList/Item/FormPreview` is the small preview of a collection item form without opening its modal:
Image

It's internals needed to calculate the field names in order to properly render the summary, humanized field values, and resolve location markers.

src/ReportManager/DetailsSection/SchemaForm/formElements/Collection/SortableList/Item/index.js
```diff
@@ -37,7 +37,7 @@ const Item = ({
   onDelete = null,
   readOnly,
   ref,
-  renderField = null,
+  renderFormElement = null,
```
luixlive, line 40: `Collection/SortableList/Item`s are the collection items. They work as a parent component for the form preview and form modal, and handle certain callbacks that needed to be updated to properly use field ids and field names:
Image

src/ReportManager/DetailsSection/SchemaForm/formElements/Collection/index.js
```diff
@@ -21,7 +21,7 @@ const Collection = ({
   id,
   onFieldChange,
   readOnly,
-  renderField,
+  renderFormElement,
```
luixlive, line 24: The collection field is the only field type that renders other form elements, inside the collection item forms. For this reason, it required some tweaks to properly calculate the field names from the ids in order to properly set form data, errors, ids, and calculate location marker ids.

src/ReportManager/DetailsSection/SchemaForm/fields/Location/index.js
```diff
@@ -39,7 +40,7 @@ const Location = ({
       jumpToLocationButtonZoom={JUMP_TO_LOCATION_BUTTON_ZOOM}
       onBlur={() => blurLocationMarker()}
       onChange={(newLocation) => onFieldChange(id, newLocation || undefined)}
-      onFocus={() => focusLocationMarker(id)}
+      onFocus={() => focusLocationMarker(name)}
```
luixlive: Location markers ids are built recursively, a location field can be nested several levels deep inside collections. Now we need to extract the names, we can't simply send down the ids, since the id ignores the index of the collection items.

A marker id will look like this: `collection_1[2].collection_2[1].location_field`. Which means the location marker belongs to the "location field inside the second item of the collection_2 that is in the third item of the collection 1". It's the only way to be able to tell which marker belongs to which field:
Image

src/ReportManager/DetailsSection/SchemaForm/formElements/Section/index.js
```diff
@@ -16,14 +16,16 @@ const Section = ({
   id,
   onFieldChange,
   onFieldErrorsChange,
-  renderField,
+  renderFormElement,
```
luixlive, line 19: Sections are the entry-point of form element rendering. This file didn't even need a change at all, since for section children, ids and names are the same. We only prefix the parent id for fields inside collections. However, for code readability, I thought it would be better to add redundancies like:
```Javascript
const fieldName = fieldId;
```

This way the code is loud about the difference between the field id and the field name, even if for this case they happen to be the same string.

src/ReportManager/DetailsSection/SchemaForm/utils/useSchemaValidations/index.js
```diff
@@ -9,27 +9,27 @@ const ajv = new Ajv2020({ allErrors: true });
 addFormats(ajv);
 ajv.addKeyword({ keyword: 'x-section', schemaType: 'string' });
 
-const insertErrorRecursively = (fieldId, message, errorPath, errors, t) => {
+const insertErrorRecursively = (fieldName, message, errorPath, errors, t) => {
```
luixlive, line 12: No logical change, just variable renaming to keep the semantic difference between ids and names.

src/ReportManager/DetailsSection/SchemaForm/index.js
```diff
@@ -11,8 +11,8 @@ import useMapLocationMarkers from './utils/useMapLocationMarkers';
 import useSchemaValidations from './utils/useSchemaValidations';
 
 import Boolean from './fields/Boolean';
-import Collection from './fields/Collection';
 import ChoiceList from './fields/ChoiceList';
+import Collection from './fields/Collection';
```
luixlive: `SchemaForm` is the main component for rendering v2 schemas as forms. Most of the changes are done in its children, but there are a couple things that needed to be updated here.

Marker ids now use "[" and "]" for indexes, like: `collection_1[0].location_1` so calculating the id of a collectionItem just consists of:
```Javascript
// "collection_1[0].location_1" --> "collection_1[0]"
const collectionItemId = `${markerIdPathParts[0]}`;
```

Also, location marker names are calculated here recursively, and now we need to distinguish between the field id and the field path:
- Field id: "collection_1.field_1".
- Field path: "collection_1[2].field_1". This one includes the specific index.

src/utils/v2-event-schemas/transformSchemaToFormElements/transformField/transformAttachmentField/index.js
```diff
@@ -1,10 +1,13 @@
 const transformAttachmentField = (
   attachmentFieldId,
+  _attachmentFieldName,
```
luixlive: From here onwards you already know what's coming, updating `transformSchemaToFormElements`. We already did this in the form builder, so this isn't new. It follows the same approach.

src/utils/img.js
```diff
@@ -2,13 +2,27 @@ import { DAS_HOST } from '../constants';
 
 const urlContainsOwnHost = url => url.includes('http');
 const imgIsDataUrl = url => url.includes('data:image');
-const imgIsFromStaticMedia = url => /^(\/static\/media)/.test(url);
+
+const imgIsAppBundledAsset = url => {
```
luixlive, line 6: Some map images weren't loading correctly in localhost because this method was outdated. It tried to find the Webpack default dev directory for assets, but Vite stores them in a different place.

src/utils/v2-event-schemas/transformSchemaToFormElements/UndefinedFormElementError/index.js
```diff
@@ -1,7 +1,7 @@
 class UndefinedFormElementError extends Error {
   constructor(formElementId, parentId) {
     super(
-      `Form element "${formElementId}" referenced in "${parentId}" is not defined in the UI schema.`,
+      `Form element "${formElementId}" referenced in "${parentId}" is not defined in the schema.`,
     );
```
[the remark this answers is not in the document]
luixlive: Good feedback, but I feel like this already give a good hint. Also the improvement would require touching several files, which I'd argue it's out of scope of an already too big PR.

src/ReportManager/DetailsSection/SchemaForm/fields/Collection/SortableList/Item/FormModal/index.js
luixlive: I'm no longer doing that. It had a smell, we shouldn't have to calculate the name like that if we have it in the form element details object as the "value" (as stored by the user in the form builder). Better to have a source of truth than to try to "calculate the truth".

src/ReportManager/DetailsSection/SchemaForm/formElements/Section/index.js
luixlive: I'm no longer calculating the field name. Now I just read it from the source of truth: the value property from the form element details object.

JoshuaVulcan (approved):

PR #1550, opened 2026-04-23
RF-1224: Use deployment dates instead of update dates in gear popup

jeslefcourt:
## Summary
- Gear popup header date now uses the most recent `last_deployed` across all devices, falling back to `gear.last_updated`
- Device list sorted ascending by `last_deployed` (undated devices last, stable sort for ties)
- Gear popup width increased 25% (`16.25rem`)
- Clicking the gear name in the sidebar list now jumps to location (same as the jump icon)

Fixes: [RF-1224](https://allenai.atlassian.net/browse/RF-1224)

## Test plan
- [ ] Open a gear popup on the map and verify the header date reflects the most recent device deployment date
- [ ] Verify devices with no `last_deployed` appear after dated devices in the popup list
- [ ] Verify clicking a gear name in the sidebar jumps the map to that gear
- [ ] Run `npx jest src/GearPopup/index.test.js` — all 11 tests pass

[RF-1224]: https://allenai.atlassian.net/browse/RF-1224?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/GearPopup/index.js
```diff
@@ -22,16 +22,21 @@ const GearPopup = ({ data }) => {
     : primaryLabel;
   const { coordinates } = data;
 
+  const mostRecentDate = (gear.devices || []).reduce((best, device) => {
+    if (!device.last_deployed) return best;
+    return !best || device.last_deployed > best ? device.last_deployed : best;
+  }, null) || gear.last_updated;
```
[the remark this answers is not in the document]
jeslefcourt: @copilot apply changes based on [this feedback](https://github.com/PADAS/das-web-react/pull/1550#discussion_r3134201751)

src/GearPopup/index.test.js
```diff
@@ -0,0 +1,130 @@
...
+};
+
+const isBefore = (a, b) =>
+  !!(a.compareDocumentPosition(b) & Node.DOCUMENT_POSITION_FOLLOWING);
+
+describe('GearPopup', () => {
+  test('renders nothing when gear is not found in the store', () => {
+    renderGearPopup(null);
+    expect(screen.queryByTestId('gear-popup-title')).toBeNull();
+  });
```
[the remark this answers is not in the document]
jeslefcourt: @copilot apply changes based on [this feedback](https://github.com/PADAS/das-web-react/pull/1550#discussion_r3134201763)

chrisj-er:
@claude review

src/GearPopup/index.js
```diff
@@ -22,16 +22,21 @@ const GearPopup = ({ data }) => {
     : primaryLabel;
   const { coordinates } = data;
 
+  const mostRecentDate = (gear.devices || []).reduce((best, device) => {
```
luixlive: We may want to memoize this so we don't run a loop on every render:
```Javascript
const lastDeployedDeviceDate = useMemo(() => (gear.devices || []).reduce((best, device) => {
  if (!device.last_deployed) {
    return best;
  }
  return !best || device.last_deployed > best ? device.last_deployed : best;
}, null), [gear.devices]);

const mostRecentUpdateDate = lastDeployedDeviceDate || gear.last_updated;
```

Just consider that in that case we need to put the `useMemo` hook over the line `if (!gear) return null;`.

src/GearPopup/index.js
```diff
@@ -46,13 +51,19 @@ const GearPopup = ({ data }) => {
         <dt>{t('typeLabel')}</dt>
         <dd>{gear.type}</dd>
 
-        {(gear.devices || []).map((device) => <React.Fragment key={device.device_id}>
+        {[...(gear.devices || [])].sort((a, b) => {
```
luixlive: A little bit harsh to read all these conditionals in the middle of the JSX. We can create a helper at the top of the file:
```Javascript
const sortByLastDeployed = (a, b) => {
  if (!a.last_deployed && !b.last_deployed) return 0;
  if (!a.last_deployed) return 1;
  if (!b.last_deployed) return -1;
  if (a.last_deployed === b.last_deployed) return 0;
  return a.last_deployed > b.last_deployed ? 1 : -1;
};
```

And here just call it:
```Javascript
{(gear.devices ?? []).sort(sortByLastDeployed).map((device) => <React.Fragment key={device.device_id}>
```

src/GearPopup/styles.module.scss
```diff
@@ -1,5 +1,11 @@
 @import '../common/styles/vars/colors';
 
+.popup {
+  [class*=mapboxgl-popup-content] {
+    width: 16.25rem;
```
luixlive, line 5: Not an issue here, just wanted to know if there's a good reason to override the default popup width. Will this new width look good in different device sizes? Mobile ones?

src/hooks/useJumpToLocation/index.js
```diff
@@ -65,7 +65,7 @@ const useJumpToLocation = () => {
       const mapBoundaries = coords.reduce(buildLocationJumpBounds, new LngLatBounds());
       map.fitBounds(mapBoundaries, { linear: true, speed: 200, padding });
     } else {
-      map.easeTo({ center: isArrayCoords ? coords[0] : coords, zoom, padding, speed: 200 });
+      map.easeTo({ center: isArrayCoords ? coords[0] : coords, zoom, padding, speed: 200, ...(offset ? { offset } : {}) });
```
luixlive: This way of receiving the third (options) parameter and spreading it feels a bit weird. What do you think of something like this:
```Javascript
return (coords, zoom = 15, options = {}) => {
  const isArrayCoords = Array.isArray(coords[0]);

  const padding = calcPadding(currentTab, isArrayCoords, itemId, isMediumLayoutOrLarger);

    const mapBoundaries = coords.reduce(buildLocationJumpBounds, new LngLatBounds());
    map.fitBounds(mapBoundaries, { linear: true, speed: 200, padding, ...options });
  } else {
    map.easeTo({ center: isArrayCoords ? coords[0] : coords, zoom, padding, speed: 200, ...options });
  }
};
```

The options parameter is still optional and you can pass the offset there, which would be spreaded to both `fitBounds` and `easeTo`.

src/SideBar/GearTab/index.js
```diff
@@ -60,7 +61,13 @@ const GearListItem = memo(({ ...gear }) => {
 
   return <>
     <p className={mapLayersStyles.itemTitle} data-testid="gear-item-name">
-      <span className={styles.displayId}>{rowTitle}</span>
+      <span
```
luixlive: I see agents are causing noise around this change, and for a good reason. This is not the most semantic HTML nor the accessible way of hearing a click. It would be more appropriate to render a button instead of a span, and add stiles to make it look just like the span. Something like:
```Javascript
{coordinates ? (
  <button
    className={`${styles.displayId} ${styles.displayIdClickable}`}
    onClick={onJumpClick}
    type="button"
  >
    {rowTitle}
  </button>
) : (
  <span className={styles.displayId}>
    {rowTitle}
  </span>
)}
``` 

Notice the <p> is gone since intractable elements shouldn't live inside a <p>.

src/SideBar/GearTab/styles.module.scss
```diff
@@ -93,6 +93,20 @@
   word-break: break-word;
 }
 
+.displayIdClickable {
```
luixlive, line 104: These styles would be updated if you switch to a button as suggested.

luixlive (changes requested):
Almost done, some small suggestions on improvements for code readability and HTML semantics 👍

src/GearPopup/styles.module.scss
jeslefcourt: Good question. The override is intentional: the gear popup now lists multiple devices, each with a `last_deployed` date, and the default Mapbox popup width made those rows wrap awkwardly — the extra ~25% gives the device list enough horizontal room.

On mobile it's safe: `16.25rem` ≈ 260px, which is narrower than the ~320px smallest-phone viewport, and `.body`/`.details` already carry `max-width: 100%` plus `word-break`, so nothing overflows on small screens. Happy to switch the fixed `width` to a `max-width` if you'd rather it shrink further on very narrow devices, but I don't think it's necessary.

src/GearPopup/index.js
jeslefcourt: Done in 1e6f95ac — wrapped in `useMemo([gear])` and moved above the `if (!gear) return null;` guard (with optional chaining since `gear` can be null there).

src/GearPopup/index.js
jeslefcourt: Done in 1e6f95ac — extracted to a top-level `sortByLastDeployed` helper. One tweak: I kept the `[...(gear.devices || [])]` spread before `.sort(...)`, since sorting `gear.devices` directly would mutate the Redux store array in place.

src/hooks/useJumpToLocation/index.js
jeslefcourt: Done in 1e6f95ac — changed the signature to `(coords, zoom = 15, options = {})` and spread `...options` into both `fitBounds` and `easeTo`. The existing `{ offset }` caller works unchanged.

src/SideBar/GearTab/index.js
jeslefcourt: Done in 1e6f95ac — renders a real `<button type="button">` when coordinates exist (plain `<span>` otherwise), dropped the wrapping `<p>`, and removed the `role`/`tabIndex`/`onKeyDown` workarounds. This also resolves the Space-key-scroll a11y nit the agents flagged.

src/SideBar/GearTab/styles.module.scss
jeslefcourt: Done in 1e6f95ac — added button-default resets (`background`/`border`/`font`/`padding`/`margin`/`color`/`text-align`/`width`) to `.displayId` so the button looks identical to the old span.

src/hooks/useJumpToLocation/index.js
```diff
@@ -56,16 +56,23 @@ const useJumpToLocation = () => {
...
   }), [routerLocation.pathname]);
 
-  return (coords, zoom = 15) => {
+  /**
+   * @param {Array} coords - Coordinate(s) to jump to.
+   * @param {number} [zoom=15] - Target zoom level.
+   * @param {Object} [options={}] - Spread into the underlying mapbox `fitBounds`/`easeTo` call,
+   *   so accepted keys are any mapbox camera/animation options (e.g. `offset`). Keys that collide
+   *   with the computed `padding`/`speed`/`linear`/`zoom` will override them.
+   */
```
luixlive: nit: we don't use JSDocs in this repository, we can clean this comment.

src/GearPopup/index.js
```diff
@@ -9,11 +9,26 @@ import { gearHumanReadableLabel } from '../utils/gear';
...
   const { t } = useTranslation('map-popups', { keyPrefix: 'gearPopup' });
   const gearId = data?.properties?.id;
   const gear = useSelector((state) => (gearId ? state.data.gear.byId[gearId] : null));
 
+  const mostRecentDate = useMemo(() => {
+    const lastDeployedDeviceDate = (gear?.devices || []).reduce((best, device) => {
+      if (!device.last_deployed) return best;
+      return !best || new Date(device.last_deployed) > new Date(best) ? device.last_deployed : best;
+    }, null);
+    return lastDeployedDeviceDate || gear?.last_updated;
```
luixlive, line 29: This falls back to `gear.last_updated` only when no device has `last_deployed`. Any deployed device suppresses a recent `gear.last_updated` from the header. Even if the device was deployed years ago, and the last update was a few hours ago. Do we want that behavior?

src/GearPopup/index.js
```diff
@@ -9,11 +9,26 @@ import { gearHumanReadableLabel } from '../utils/gear';
 
 import * as styles from './styles.module.scss';
 
+const sortByLastDeployed = (a, b) => {
+  if (!a.last_deployed && !b.last_deployed) return 0;
+  if (!a.last_deployed) return 1;
+  if (!b.last_deployed) return -1;
+  return new Date(a.last_deployed) - new Date(b.last_deployed);
+};
```
luixlive, line 17: nit: `sortByLastDeployed` and `mostRecentDate` construct Date objects for ISO 8601 strings. That string format can be compared directly as a string because of the way the string is built: `a.last_deployed > b.last_deployed`. Just a tiny processing and memory optimization.

luixlive (approved):
Nice work! Still left 2 nits and one question, but no blockers.

JoshuaVulcan (approved):

src/hooks/useJumpToLocation/index.js
jeslefcourt: Done in 9b32d450 — removed the JSDoc block.

src/SideBar/GearTab/index.js
```diff
@@ -39,6 +39,10 @@ const manufacturerStateKey = (manufacturerKey) => (manufacturerKey === '' ? '__o
 const gearGroupTracker = trackEventFactory(MAP_LAYERS_CATEGORY);
 
 const GEAR_JUMP_ZOOM = 14;
+// Approximate downward screen-space offset (px) applied when jumping, so the bottom-anchored gear
+// popup has room to open below the gear without clipping the top of the viewport. The value is
+// approximate; tall popups (many devices) may still clip.
```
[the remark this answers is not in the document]
jeslefcourt: Done in 9b32d450 — reworded: with `anchor: 'bottom'` the popup opens above the gear, so the comment now says "above".

src/SideBar/GearTab/index.js
```diff
@@ -59,9 +63,20 @@ const GearListItem = memo(({ ...gear }) => {
...
-    <p className={mapLayersStyles.itemTitle} data-testid="gear-item-name">
-      <span className={styles.displayId}>{rowTitle}</span>
-    </p>
+    {coordinates ? (
+      <button
+        className={`${mapLayersStyles.itemTitle} ${styles.displayId} ${styles.displayIdClickable}`}
+        data-testid="gear-item-name"
+        onClick={onJumpClick}
+        type="button"
+      >
```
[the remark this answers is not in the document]
jeslefcourt: Done in 9b32d450 — the gear-name click now fires the same analytics event as the jump icon and navigates to `/` below the medium breakpoint so the sidebar closes on small screens. Added a test covering the small-screen navigation.

src/GearPopup/index.js
jeslefcourt: Keeping the `Date`-based comparison intentionally: lexicographic comparison of ISO 8601 strings is only safe when every timestamp carries the same UTC offset. ER API timestamps can come back with tenant-local offsets (e.g. `2024-06-15T12:00:00+03:00`), and string ordering silently breaks across mixed offsets. `new Date(...)` normalizes to the instant regardless of offset, and since this runs over a handful of devices inside a `useMemo`, the extra allocation is negligible.

src/GearPopup/index.js
jeslefcourt: Intentional. The header is meant to answer "when was this gear deployed" — an edit to the gear record doesn't change the fact of when it was deployed, so a recent `last_updated` shouldn't mask the deployment date. `last_updated` is only used as a fallback when no device has ever been deployed, so there's still something meaningful to show. (This is the core of RF-1224 — update dates were being read as field activity.)

PR #1554, opened 2026-04-28
Upgrade to Vite 8

luixlive:
## What does this PR do?
- Upgrade all repository dependencies
- Migrate to Vite 8
- Refactoring of our Log In page

## Evidence
Kapture 2026-04-28 at 10 49 23

### Relevant link(s)
- [Branch Environment](https://upgrade-dependencies-04-27.dev.pamdas.org/)

## Note

Do not worry about the amount of files changed! It's because of the Eslint auto-fix command. Reviewing like 50 files takes like 1 second each: removing a whitespace, or a redundant double negation.


.yarn/releases/yarn-4.14.1.cjs
luixlive, line 1: Upgrade Yarn to 4.14.

public/locales/en-US/login.json
```diff
@@ -9,9 +9,16 @@
     "signInFailed": "Sign-in failed. Please try again.",
     "unknownErrorMessage": "An error has occurred. Please try again."
   },
+  "errors": {
```
luixlive, line 12: I refactored our Log In component. It was using React Bootstrap fields we weren't using anymore in our forms, so the designs felt off with the rest of the app. Also, I increased our a11y scope so in these i18n files you'll see some new error messages, headings and aria labels.

src/__test-helpers/mocks.js
```diff
@@ -37,7 +37,7 @@ export const createMapMock = (override = {}) => {
     },
     on: jest.fn((name, ...rest) => {
       const layerId = isString(rest[0]) ? rest[0] : '';
-      const callback = !!layerId ? rest[1] : rest[0];
+      const callback = layerId ? rest[1] : rest[0];
```
luixlive, line 40: I ran the Eslint auto-fix command, which produced changes in lots of files, but they are as simple as this. You can just quick scroll them, the logic is unaffected.

src/Login/index.js
```diff
@@ -1,159 +1,251 @@
-import React, { memo, useCallback, useEffect, useState } from 'react';
-import Alert from 'react-bootstrap/Alert';
-import Button from 'react-bootstrap/Button';
-import Form from 'react-bootstrap/Form';
+import React, { useCallback, useEffect, useRef, useState } from 'react';
```
luixlive, line 1: The Log In page changes. More accessibility, consistent UI with the app, better error messages and UX (autofocusing fields when there are errors), added a loading indicator in our button... 🥇

src/i18n.js
```diff
@@ -60,5 +60,4 @@ i18n
         queryStringParams: { v: I18N_FILES_VERSION },
       }],
     },
-    showSupportNotice: false,
```
luixlive, line 63: `i18next` released a major release. There's no breaking changes for us, but we no longer need to disable the show support notice manually 😄

eslint.config.js
```diff
@@ -19,10 +19,15 @@ export default defineConfig([
     'terraform/',
   ]),
 
-  js.configs.recommended,
-
   {
+    name: 'src/javascript',
```
luixlive, line 23: Eslint 10 is out, and as part of the upgrade I also cleaned a bit our flat config to be easier to read.

vite.config.js
```diff
@@ -1,7 +1,45 @@
-import { defineConfig } from 'vite';
+import { defineConfig, transformWithOxc } from 'vite';
```
luixlive, line 1: Vite 8 ⚡ In my computer the build time was decreased from ~8 seconds (already ridiculously fast in comparison to Webpack) to less than 2 seconds 😆 

Since they moved away from esbuild to oxc, we needed a new custom plugin that enabled us to treat `.js` files as `.jsx`. It's a practice the entire industry just dropped and they are no longer supporting it now providing configuration to do so. The best long-term fix would be to rename all our files in the app, but in the meanwhile, the plugin is enough.

src/Login/index.js
```diff
@@ -1,159 +1,251 @@
...
 import useNavigate from '../hooks/useNavigate';
 
 import * as styles from './styles.module.scss';
 
+const SUBMIT_LOADER_SIZE = 18;
+
 const LoginPage = () => {
   const dispatch = useDispatch();
   const location = useLocation();
+  const {
```
luixlive, line 25: @StephenWithPH can you do me a big favor? Just make sure locally that the Auth0 Log In flow still works as expected. I also updated its button, alerts (actually fixed a bug there since the alert was being shown twice), and refactored a bit the code. I tested it myself and seems fine, but I'd rather ask for 3 mins of your time to pull this branch and test it to make sure I didn't break anything 🙏

StephenWithPH: Will do. Until other PRs land, I'll do this _after_ the other reviewers.

StephenWithPH (approved):
I validated that I was able to log in using Auth0 in the deployed env:

Image

src/Login/index.js
StephenWithPH: https://github.com/PADAS/das-web-react/pull/1554#pullrequestreview-4193128145

src/DetailViewComponents/ActivitySection/ContainedReportListItem/index.js
```diff
@@ -75,7 +75,7 @@ const ContainedReportListItem = ({ cardsExpanded, onCollapse, onExpand, report }
       in={isOpen}
     >
       <div>
-        {!!reportFromEventStore
```
JoshuaVulcan, line 78: Were these boolean coercions causing issues, or is this picking at nits?

JoshuaVulcan (approved):

src/DetailViewComponents/ActivitySection/ContainedReportListItem/index.js
luixlive: There weren't any issues, it surprised me too. As part of the upgrades, we moved to Eslint 10 and a couple plugin packages were updated too. We use the recommended practices for most of them, so my guess is they just decided it's a bad practice to use them if you are not holding a boolean variable.

```Javascript
// Here it's meaningful to actually coerce as a boolean:
const hasReportFromEventStore = !!reportFromEventStore;

// Here, it isn't. It's exactly the same if we make it a boolean or just leave it to
// the run time execution to validate if the value is truthy.
if (reportFromEventStore) {
  ...
}
```

PR #1557, opened 2026-05-01
ERA-13228 Fix basemap tile layer maxZoom and maxNativeZoom having no effect

jeslefcourt:
## Summary
- Each `SourceComponent` now derives its own `sourceConfig` from its own `layer` prop, so per-layer `maxNativeZoom` is correctly applied to each raster source. Previously all sources shared a single config derived from `currentBaseLayer`, and `useMapSources` silently ignored config changes for already-existing sources.
- `mapConfig` (which drives `map.setMaxZoom`/`map.setMinZoom`) is now computed from `activeLayer` (always fresh from the API via `baseLayers`) with fallback to `currentBaseLayer`, preventing stale `redux-persist` state from capping interactive zoom at the wrong value on initial load.
- Zoom limits are re-asserted on the Mapbox GL `idle` event as a defensive measure against async source processing overriding `transform.maxZoom`.

Jira: [ERA-13228](https://allenai.atlassian.net/browse/ERA-13228)

## Test plan
- [ ] 13 new automated tests added in `TileLayerRenderer.test.js` — all passing
- [ ] Verify `maxZoom` sets the interactive zoom ceiling correctly on initial load
- [ ] Verify `maxNativeZoom` causes tile overzooming above its value without capping interactive zoom
- [ ] Verify both settings work correctly on first load without needing to switch layers and back

[ERA-13228]: https://allenai.atlassian.net/browse/ERA-13228?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/BaseLayerRenderer/TileLayerRenderer.js
+++ b/src/BaseLayerRenderer/TileLayerRenderer.js
@@ -17,14 +17,15 @@ const RASTER_SOURCE_OPTIONS = {
 
 const RenderFunction = ({ children }) => <>{children}</>;
 
-const SourceComponent = ({ id, tileUrl, sourceConfig }) => {
-  const config = useMemo(() => ({
-    ...RASTER_SOURCE_OPTIONS,
-    tiles: [
-      tileUrl,
-    ],
-    ...sourceConfig,
-  }), [sourceConfig, tileUrl]);
+const SourceComponent = ({ id, layer }) => {
+  const config = useMemo(() => {
+    const { sourceConfig } = calcConfigForMapAndSourceFromLayer(layer);
+    return {
+      ...RASTER_SOURCE_OPTIONS,
+      tiles: [layer.attributes.url],
+      ...sourceConfig,
+    };
+  }, [layer]);
 
   useMapSources([{ id }], config);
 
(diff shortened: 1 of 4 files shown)


jeslefcourt:
@claude review once

src/BaseLayerRenderer/TileLayerRenderer.js
```diff
@@ -17,14 +17,15 @@ const RASTER_SOURCE_OPTIONS = {
...
-  const config = useMemo(() => ({
-    ...RASTER_SOURCE_OPTIONS,
-    tiles: [
-      tileUrl,
-    ],
-    ...sourceConfig,
-  }), [sourceConfig, tileUrl]);
+const SourceComponent = ({ id, layer }) => {
+  const config = useMemo(() => {
+    const { sourceConfig } = calcConfigForMapAndSourceFromLayer(layer);
```
luixlive: I see that `calcConfigForMapAndSourceFromLayer` is a utility only used here. We used to call it once, but this PR now separates the call to fetch `sourceConfig` and the one to fetch `mapConfig`. I think it would be cleaner to break that utility into two simpler ones: `calculateSourceConfigurationFromLayer` and `calculateMapConfigurationFromLayer` (I prefer verbose variable names, but that's optional):
```Javascript
const sourceConfiguration = calculateSourceConfigurationFromLayer(layer);
...
const mapConfiguration = calculateMapConfigurationFromLayer(layer);
```

src/BaseLayerRenderer/TileLayerRenderer.js
```diff
@@ -40,15 +41,24 @@ const TileLayerRenderer = (props) => {
     layers.find(({ id }) => id === currentBaseLayer?.id)
   , [currentBaseLayer?.id, layers]);
 
-  const { mapConfig, sourceConfig } = useMemo(() =>
-    calcConfigForMapAndSourceFromLayer(currentBaseLayer)
-  , [currentBaseLayer]);
+  const { mapConfig } = useMemo(() =>
```
luixlive: The calculation of the configurations is just a couple conditionals and accessing object properties, nothing complex or inefficient. I'm pretty sure memoization here are harming more than helping. Same applies in the call above to get source configuration. We should get rid of the `useMemo` 👍

luixlive (approved):
Nice 💯  I added a couple nit comments that are non-blockers if you want to improve this PR a bit more before merging!

PR #1561, opened 2026-05-06
ERA-13244: Fix HTTP errors showing "unknown error" instead of human-readable message on event submit

jeslefcourt:
## Summary
- Fixes `generateErrorMessageForRequest` in `src/utils/request.js` which was checking `error.response?.status?.code` — Axios puts the HTTP status code at `error.response.status` (a number), so `.code` was always `undefined` and the switch never matched any case
- Updates `generateErrorListForApiResponseDetails` catch block in `ReportDetailView` to call `generateErrorMessageForRequest` instead of always falling back to "unknown error"
- A 429 on event submit now shows: *"Too many requests. Please try again later, and contact your administrator if this problem persists (429)."*

## Test plan
- [ ] Submit an event while rate-limited (HTTP 429) and verify the human-readable message appears
- [ ] Verify other HTTP error codes (400, 403, 500, etc.) also show their localized messages
- [ ] Verify that structured API error detail responses still parse and display correctly
- [ ] Verify the event feed error path (FEED_FETCH_ERROR) still works correctly

Jira: [ERA-13244](https://allenai.atlassian.net/browse/ERA-13244)

[ERA-13244]: https://allenai.atlassian.net/browse/ERA-13244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/utils/request.js
```diff
@@ -7,8 +7,8 @@ export const generateErrorMessageForRequest = (error) => {
     return t('noResponse');
   }
 
-  if (error.response?.status?.code) {
-    switch (error.response.status.code) {
+  if (error.response?.status) {
```
luixlive: Hmm this code has been there for a while. It seems risky to change it suddenly. I just tested an erroneous call on our Open API:
Image

I see the code under `status.code`. Maybe we need to add some sort of intelligent condition here?
```Javascript
const statusCode = error.response?.status?.code ?? error.response?.status;
switch (statusCode) {
```

luixlive (changes requested):
Trivial and sound. I just think we need to double check if that change in `generateErrorMessageForRequest` is correct.

src/utils/request.js
jeslefcourt: Good catch, and you're right that dropping the `status.code` path was risky. Addressed in 8e3716ef by reading the code defensively from all three shapes, body-first:

```js
const statusCode = error.response?.data?.status?.code
  ?? error.response?.status?.code
  ?? error.response?.status;
```

That covers the `status.code` you saw, the DAS error-body location `data.status.code` (the same `status` object `generateErrorListForApiResponseDetails` reads `detail` from), and a plain numeric HTTP `status` — so the human-readable message surfaces regardless of which shape comes back.

Separately, I also reordered this function so the `error.response.status` switch runs **before** the `error.request` no-response check: an axios error that received a response (e.g. a real 429) carries *both* `.request` and `.response`, so the old request-first order short-circuited to "No response from the server" and the status switch was never reached.

Added `src/utils/request.test.js` (11 cases) covering all three shapes, body-code precedence over numeric status, the no-response path, and the no-code fallback. Full suite green.

src/utils/request.js
```diff
@@ -3,12 +3,12 @@ import i18next from 'i18next';
 export const generateErrorMessageForRequest = (error) => {
   const t = i18next.getFixedT(null, 'utils', 'generateErrorMessageForRequest');
 
-  if (error.request) {
-    return t('noResponse');
-  }
+  const statusCode = error.response?.data?.status?.code
+    ?? error.response?.status?.code
```
luixlive: Do we ever get a property `code` directly under `status`? This seems like possible dead code.

luixlive (approved):
One question, but this is ready 💯

src/utils/request.js
jeslefcourt: Good instinct — it was indeed dead code. For axios errors `error.response.status` is always the numeric HTTP status, so `.code` on it can never be set; the `status.code` in your screenshot lives in the response **body**, i.e. `data.status.code`, which the first clause covers. (It's also the original pre-fix condition — its being unreachable is exactly why this function never produced a status-specific message before.) Dropped the middle clause and its two test cases in 2f104bfd; suite still green.

PR #1598, opened 2026-06-12
ERA-13404: Add common-DB "Sign in with email" login path

StephenWithPH:
## What does this PR do?

Adds the common-DB Auth0 login path to the web client's login page. On a site with `require_idp = true` but **no** `idp_org_id` configured, the page now renders a **"Sign in with email"** button that initiates the common-DB Auth0 PKCE redirect with no `organization` param (Auth0 falls back to the tenant's Default Directory) — replacing the old behavior, which rendered an empty form and tripped a `missingOrg` error toast.

- New `loginButtonEmail` ("Sign in with email") string in all six locales.
- `buildAuth0AuthorizationParams` (in `utils/auth0.js`) forwards `organization` only when a non-blank `idp_org_id` is configured (trimmed); omits it otherwise.
- The login page renders the Auth0 button for both cases, labelled "Sign in" for the rcuksa org path and "Sign in with email" for the common-DB path.
- Removes the now-dead `missingOrg` effect and its orphaned locale strings.

**Unchanged / out of scope:**
- rcuksa org-based path (`require_idp && idp_org_id`) and legacy ROPC username/password path (`require_idp = false`) are untouched.
- In-memory token custody (`cacheLocation: 'memory'`, refresh tokens) is a separate, deferred ticket — common-DB tokens are custodied exactly as rcuksa's are today.
- RFC 9728 discovery lands in Phase 2B (ERA-13429).

## Evidence

Unit tests (`yarn test src/Login src/utils/auth0`) — **44 passing**, covering:
- the "Sign in with email" button renders and no `missingOrg` alert fires when `require_idp && !idp_org_id`;
- clicking it calls `loginWithRedirect` with `{ authorizationParams: { audience } }` and **no** `organization`;
- `buildAuth0AuthorizationParams` forwards a configured org, and omits it for null / empty / whitespace-only (and trims surrounding whitespace);
- no regression on the rcuksa ("Sign in" + `organization`) or ROPC paths.

> ⚠️ The live Universal Login round-trip can't be exercised here yet — it depends on backend tickets **ERA-13338** and **ERA-13402** being live.

### Relevant link(s)
- [ERA-13404](https://allenai.atlassian.net/browse/ERA-13404)
- [Branch Environment](https://era-13404.dev.pamdas.org)

## Notes

- Three small, independently-green commits (i18n string → conditional org param + tests → render branch + cleanup).
- Absorbs the flow-selection / render-branch work from the closed ERA-13403, per the ticket.

[ERA-13404]: https://allenai.atlassian.net/browse/ERA-13404?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

This is a site using the new, common DB:

https://github.com/user-attachments/assets/6c24d52a-2b56-4e2c-8e46-b961381e6ef4

---

... and this is our rcuksa-fixture site, which mimics rcuksa in terms of Auth0 organizations. Note that I deployed this branch into that env.

https://github.com/user-attachments/assets/2f1cf449-2eea-49db-a14c-32c0a9744c3c

... the verification suite also passed https://github.com/PADAS/authentication-test-harness/actions/runs/27439100021/job/81108439980#step:5:37


chrisj-er (approved):

src/utils/auth0.js
```diff
@@ -2,3 +2,15 @@ export const hasAuth0CallbackParams = (searchParams) => {
   const urlParams = new URLSearchParams(searchParams);
   return urlParams.has('code') && (urlParams.has('state') || urlParams.has('error'));
 };
+
+// Build the authorizationParams for an Auth0 redirect login. The IdP
+// organization is forwarded only when one is configured (a non-blank value
+// after trimming); common-DB sites have no org and must omit the param
+// entirely so Auth0 falls back to the tenant's Default Directory.
+export const buildAuth0AuthorizationParams = (audience, idpOrgId) => {
```
luixlive, line 10: nit (personal opinion honestly): It seems unnecessary to create a helper method just to build an object as simples as this one. Specially that the method is called only once, we are not even reusing it.
```javascript
const idpOrgIdTrimmed = idpOrgId?.trim();
await auth0LoginWithRedirect({
  authorizationParams: {
    audience: appConfig.auth0.audience,
    ...(idpOrgIdTrimmed ? { organization: idpOrgIdTrimmed } : {})
  },
});
```

src/Login/index.test.js
```diff
@@ -136,19 +136,36 @@ describe('Login', () => {
     expect(button).toBeDisabled();
   });
 
-  test('hides the sign-in button and shows a configuration error when IDP is required without an organization ID', () => {
+  test('shows the common-DB "Sign in with email" button and no configuration error when IDP is required without an organization ID', () => {
```
luixlive, line 139: nit: `No configuration error` sounds like we are protecting ourselves from a regression that isn't likely to happen. A simpler and honest name would be: `shows the common-DB username-and-password form when IDP login is required without an organization ID`

luixlive (approved):
A couple nits, but no blockers. Nice and clean 💯

src/utils/auth0.js
StephenWithPH: I was on the fence about this. Since we will soon need to selectively specify a different Auth0 connection based on the user's choice, I figured this helper would grow a little more in the future.

src/Login/index.test.js
StephenWithPH: ah... that was to cover the negative assertion:

https://github.com/PADAS/das-web-react/blob/4b2201570a4e055a534fc99e936e178ae14fa22a/src/Login/index.test.js#L151

PR #1621, opened 2026-06-29
ERA-13529: Force symbol re-layout so cleared subjects disappear immediately

jeslefcourt:
## Summary

Fixes [ERA-13529](https://allenai.atlassian.net/browse/ERA-13529): when subject clustering is **off**, clearing subjects from the map (**Clear All** or clearing a subject group) left the icons on screen until the user panned.

## Root cause

`setData` empties the `subject-symbol-source` GeoJSON source immediately (verified: `features: 0`), but Mapbox does **not** rebuild a symbol layer's rendered bucket when its source data shrinks unless the layer is forced to re-lay-out. `triggerRepaint()`, `_update(true)`, and `fadeDuration: 0` all leave the stale bucket drawn — only a camera change (pan/zoom) or a layer mutation rebuilds it. That's why panning made the icons disappear.

## Fix

In `SubjectsLayer`, when the unclustered subject set shrinks, wait for the source to finish reloading (its `sourcedata` / `isSourceLoaded` event) and toggle the unclustered subject layers' `visibility` off→on to force the rebuild — **without moving the camera** and without waiting on global `idle`. The layer's existing visibility is preserved, so a user's name-labels-off setting is respected. Clustered subjects render as DOM markers and update immediately, so this only applies when clustering is off.

## Tests

- New `src/SubjectsLayer/index.test.js` (7 tests). The regression test asserts the visibility toggle fires only after the source reloads; it was confirmed to **fail without the fix**. Guard tests cover: clustering on, no shrink, wrong source id, source not yet loaded, and an already-hidden layer being skipped.
- Added `getLayoutProperty` / `isSourceLoaded` defaults to `createMapMock`.
- `yarn test src/SubjectsLayer`: 7/7 pass. Lint clean on changed files.

[ERA-13529]: https://allenai.atlassian.net/browse/ERA-13529?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/SubjectsLayer/index.js
+++ b/src/SubjectsLayer/index.js
@@ -57,20 +57,14 @@ const SubjectsLayer = ({ mapImages = {}, onSubjectClick }) => {
     setMapSubjectFeatures({ ...subjectFeatureCollection });
   }, [mapImages, subjectFeatureCollection]);
 
-  // When an unclustered symbol source's data shrinks (e.g. "Clear All" or clearing a
-  // subject group), Mapbox keeps the previously rendered symbol bucket on screen — the
-  // removed icons linger until something forces the layer to re-lay-out, which is why a
-  // pan/zoom makes them disappear. Updating the source data, repainting, and forcing a
-  // placement pass all fail to rebuild the bucket; toggling the layer's visibility does.
-  // So, once the source has finished reloading the smaller dataset, briefly toggle the
-  // unclustered subject layers' visibility to force the rebuild without moving the camera.
-  // Clustered subjects render as DOM markers, which update immediately, so this only
-  // applies when clustering is off.
+  // When an unclustered symbol source's data shrinks Mapbox keeps the
+  // previously rendered symbol bucket on screen. Toggling the layer's
+  // visibility rebuilds the bucket.
   const previousSubjectCountRef = useRef(0);
+  const currentSubjectCount = mapSubjectFeatures.features?.length ?? 0;
   useEffect(() => {
-    const currentCount = mapSubjectFeatures.features?.length ?? 0;
-    const didShrink = currentCount < previousSubjectCountRef.current;
-    previousSubjectCountRef.current = currentCount;
+    const didShrink = currentSubjectCount < previousSubjectCountRef.current;
+    previousSubjectCountRef.current = currentSubjectCount;
 
     if (!didShrink || !map || shouldSubjectsBeClustered) return;
 
@@ -91,7 +85,7 @@ const SubjectsLayer = ({ mapImages = {}, onSubjectClick }) => {
     };
     map.on('sourcedata', onSourceData);
     return () => map.off('sourcedata', onSourceData);
-  }, [map, mapSubjectFeatures, shouldSubjectsBeClustered]);
+  }, [currentSubjectCount, map, shouldSubjectsBeClustered]);
 
   const onSubjectSymbolClick = useMemo(() => withMultiLayerHandlerAwareness(
     map,
--- a/src/SubjectsLayer/index.js
+++ b/src/SubjectsLayer/index.js
@@ -1,4 +1,4 @@
-import React, { memo, useCallback, useContext, useEffect, useMemo, useState } from 'react';
+import React, { memo, useCallback, useContext, useEffect, useMemo, useRef, useState } from 'react';
 import { featureCollection } from '@turf/turf';
 import { useSelector } from 'react-redux';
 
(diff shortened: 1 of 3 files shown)


src/SubjectsLayer/index.js
```diff
@@ -57,6 +57,42 @@ const SubjectsLayer = ({ mapImages = {}, onSubjectClick }) => {
...
 
+  // When an unclustered symbol source's data shrinks (e.g. "Clear All" or clearing a
+  // subject group), Mapbox keeps the previously rendered symbol bucket on screen — the
+  // removed icons linger until something forces the layer to re-lay-out, which is why a
+  // pan/zoom makes them disappear. Updating the source data, repainting, and forcing a
+  // placement pass all fail to rebuild the bucket; toggling the layer's visibility does.
+  // So, once the source has finished reloading the smaller dataset, briefly toggle the
+  // unclustered subject layers' visibility to force the rebuild without moving the camera.
+  // Clustered subjects render as DOM markers, which update immediately, so this only
+  // applies when clustering is off.
```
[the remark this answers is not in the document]
jeslefcourt: Good catch on the mismatch — this was against the original PR description. The `triggerRepaint` approach was abandoned after testing confirmed it (and `_update(true)` and `fadeDuration: 0`) does not clear the stale symbol bucket; only forcing a layer re-layout does. The PR title/body have since been updated to describe the actual visibility-toggle fix and its scope (subjects, clustering-off). Resolving.

src/__test-helpers/mocks.js
```diff
@@ -54,6 +54,8 @@ export const createMapMock = (override = {}) => {
     }),
     setLayerZoomRange: jest.fn(),
     getLayer: jest.fn(),
+    getLayoutProperty: jest.fn((layerId, property) => (property === 'visibility' ? 'visible' : undefined)),
+    isSourceLoaded: jest.fn(() => true),
```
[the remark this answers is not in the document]
jeslefcourt: Done — `createMapMock.getLayoutProperty` now throws for layers that don't exist (via `this.getLayer(layerId)`), mirroring Mapbox GL. Production already guards the real call with `map.getLayer(layerId) && ...`, and `getLayoutProperty` is only used by SubjectsLayer, so no other suites are affected. SubjectsLayer tests remain green (7/7).

jeslefcourt:
@luixlive  This solution feels like a bit of a hack, but I tried a half-dozen other approaches that either didn't work or added side effects.  This seems to work smoothly.  The load of subject icons when clustering is turned off seems to be a lot smoother too (sometimes you had to pan the map or zoom in/out to get things to load properly).

src/__test-helpers/mocks.js
```diff
@@ -54,6 +54,15 @@ export const createMapMock = (override = {}) => {
     }),
     setLayerZoomRange: jest.fn(),
     getLayer: jest.fn(),
+    getLayoutProperty: jest.fn(function (layerId, property) {
+      // Mapbox GL throws for layers that don't exist; mirror that so tests can't pass
```
luixlive: I like the deep mock even throwing the error, but the comment is unnecessary, just over-explaining something easy to understand from the code.

src/SubjectsLayer/index.js
```diff
@@ -57,6 +57,42 @@ const SubjectsLayer = ({ mapImages = {}, onSubjectClick }) => {
     setMapSubjectFeatures({ ...subjectFeatureCollection });
   }, [mapImages, subjectFeatureCollection]);
 
+  // When an unclustered symbol source's data shrinks (e.g. "Clear All" or clearing a
```
luixlive: I think Claude is getting crazier with these comments 😅 Some times it delivers a 5 line comment for a 2 line change haha. The reasoning behind this code is not easy to understand, so the comment is good, but can be reduced to something like:
```
  // When an unclustered symbol source's data shrinks Mapbox keeps the
  // previously rendered symbol bucket on screen. Toggling the layer's
  // visibility rebuilds the bucket.
```

src/SubjectsLayer/index.js
```diff
@@ -57,6 +57,42 @@ const SubjectsLayer = ({ mapImages = {}, onSubjectClick }) => {
...
+
+    const onSourceData = (event) => {
+      if (event.sourceId === UNCLUSTERED_SOURCE_ID && event.isSourceLoaded) {
+        map.off('sourcedata', onSourceData);
+        rebuildLayers();
+      }
+    };
+    map.on('sourcedata', onSourceData);
+    return () => map.off('sourcedata', onSourceData);
+  }, [map, mapSubjectFeatures, shouldSubjectsBeClustered]);
```
luixlive: Race condition here.

The effect's dependency array includes `mapSubjectFeatures`, a new object reference produced every time `mapImages` or `subjectFeatureCollection` changes. `subjectFeatureCollection` recomputes on any subject store change, i.e. on every real-time WebSocket position update for any subject, cleared or not.

Sequence:
1. Subjects are cleared. Effect runs: `didShrink` is `true`, sets `previousSubjectCountRef.current = 0`, subscribes a sourcedata listener waiting for the reload.
2. Before that async sourcedata event fires, an unrelated WebSocket position update for some other subject produces a new `mapSubjectFeatures` reference. Effect cleanup tears down the still-pending listener. Effect re-runs.
3. On re-run, count is still 0 and the ref was already set to 0 in step 1, so `didShrink` is now false. It returns early without re-subscribing.
4. The rebuild is silently dropped.

Suggested fix: key the effect on something that only changes when the count changes.

luixlive (approved):
More long comments from Claude that could be summarized, and a possible race condition. Both are nitpicks, so I'll approve, both worth feeding these comments to Claude.

src/SubjectsLayer/index.js
jeslefcourt: Good catch. Fixed in 1c07f34f by keying the effect on the derived subject count (a primitive) instead of the `mapSubjectFeatures` object reference. Now an unrelated WebSocket position update no longer re-runs the effect or tears down the pending `sourcedata` listener — the effect only re-runs when the count actually changes, so the shrink detection and rebuild survive.

src/SubjectsLayer/index.js
jeslefcourt: Trimmed to your three-line version in 1c07f34f 😄

src/__test-helpers/mocks.js
jeslefcourt: Removed the comment in 1c07f34f — agreed, the code speaks for itself here.

PR #1623, opened 2026-06-30
ERA-13534: Update subject historical details table with real-time observations

jeslefcourt:
## Summary

The subject historical details table (`SubjectHistoricalDataModal`, opened from the history button in the map subject popup) fetched observations once on open and never reacted to WebSocket updates. New observations only appeared after closing/reopening the modal or paginating.

This subscribes to the subject's `last_position_date` in the store (updated by `SOCKET_SUBJECT_STATUS`) and, on page 1, **prepends** the new observation — or **replaces** an existing row matched by `recorded_at` instant — built from the subject store entry. No refetch, no loading flicker.

## Changes

- **Realtime updates:** new effect prepends/replaces observation rows on page 1 when `subjectStore[subjectId].last_position_date` changes.
- **Crash fix:** coerce `device_status_properties` to an array in `buildObservationFromSubject` — a position-only socket frame can leave the value object-shaped (`{0:…,1:…}`), which previously threw `TypeError: .find is not a function` in `ObservationRow`.
- **Dedupe/count fix:** match observations by parsed `recorded_at` instant (handles API `…-08:00` vs socket `…Z` formats) instead of a synthesized id, so re-delivered observations replace in place rather than duplicating and inflating the pagination count.
- Ignore `.claude/settings.local.json` (contains local credentials).

## Testing

`yarn jest src/SubjectHistoricalDataModal` → **15/15 pass**. New coverage: page-1 prepend, instant-based replace-in-place (no count bump), re-delivered fetched row, object-shaped device props (crash regression), null `last_position`, and static sensors.

## Notes

- Scoped to `SubjectHistoricalDataModal`; `getDeviceStatusPropertiesForSubject` and the subjects reducer were intentionally left untouched.
- The lone `react-hooks/set-state-in-effect` lint is a repo-wide migration artifact (also on the pre-existing fetch effect), left unsuppressed for symmetry.

Jira: [ERA-13534](https://allenai.atlassian.net/browse/ERA-13534)

[ERA-13534]: https://allenai.atlassian.net/browse/ERA-13534?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/SubjectHistoricalDataModal/index.js
+++ b/src/SubjectHistoricalDataModal/index.js
@@ -1,4 +1,4 @@
-import React, { memo, useEffect, useId, useState } from 'react';
+import React, { memo, useEffect, useId, useRef, useState } from 'react';
 import flatten from 'lodash/flatten';
 import Modal from 'react-bootstrap/Modal';
 import startCase from 'lodash/startCase';
@@ -9,6 +9,7 @@ import { useDispatch, useSelector } from 'react-redux';
 import { useTranslation } from 'react-i18next';
 
 import { fetchObservationsForSubject } from '../ducks/observations';
+import { getDeviceStatusPropertiesForSubject } from '../utils/subjects';
 import { selectCoordinatesRepresentation } from '../selectors/location';
 import useStringifyCoordinates from '../hooks/useStringifyCoordinates';
 
@@ -26,6 +27,21 @@ export const getObservationUniqProperties = (observations) => {
   const observationsDeviceProperties = observations.map((result) => result?.device_status_properties ?? []);
   const uniqPropertiesByLabel = unionBy(flatten(observationsDeviceProperties), 'label');
   return uniqPropertiesByLabel.map((property) => property.label);
+};
+
+const buildObservationFromSubject = (subject) => {
+  const recordedAt = subject.last_position?.properties?.coordinateProperties?.time ?? subject.last_position_date;
+  const coordinates = subject.last_position?.geometry?.coordinates;
+  const deviceStatusProperties = getDeviceStatusPropertiesForSubject(subject);
+
+  return {
+    id: recordedAt,
+    recorded_at: recordedAt,
+    location: coordinates ? { longitude: coordinates[0], latitude: coordinates[1] } : null,
+    device_status_properties: Array.isArray(deviceStatusProperties)
+      ? deviceStatusProperties
+      : Object.values(deviceStatusProperties ?? {}),
+  };
 };
 
 const getPageItemNumbers = (pagesCount, activePage) => {
@@ -107,6 +123,11 @@ const SubjectHistoricalDataModal = ({ subjectId, subjectIsStatic, title }) => {
   const [observationProperties, setObservationProperties] = useState([]);
   const [subjectObservations, setSubjectObservations] = useState([]);
 
+  const subject = useSelector((state) => state.data?.subjectStore?.[subjectId]);
+  const subjectLastPositionDate = subject?.last_position_date;
+
+  const lastProcessedPositionDate = useRef(subjectLastPositionDate);
+
   useEffect(() => {
     setLoadState(true);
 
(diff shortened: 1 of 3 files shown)


src/SubjectHistoricalDataModal/index.js
```diff
@@ -124,6 +145,33 @@ const SubjectHistoricalDataModal = ({ subjectId, subjectIsStatic, title }) => {
...
+    if (existingIndex !== -1) {
+      setSubjectObservations((current) => current.map((observation, index) =>
+        index === existingIndex ? { ...newObservation, id: observation.id } : observation));
+      return;
+    }
+
+    setSubjectObservations((current) => [newObservation, ...current].slice(0, ITEMS_PER_PAGE));
+    setObservationsCount((count) => count + 1);
+    setObservationProperties((current) => current.length ? current : getObservationUniqProperties([newObservation]));
+  }, [subject, subjectLastPositionDate, activePage, subjectObservations]);
```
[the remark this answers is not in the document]
jeslefcourt: Good catch — fixed in e5fcaa86. Row replacement now matches by the `recorded_at` instant inside the functional `setState` updater (by value, not a precomputed index), so it can no longer target the wrong row if `current` drifts from the render-time snapshot. The render-time check is retained only to gate the count/column updates.

src/SubjectHistoricalDataModal/index.js
```diff
@@ -124,6 +145,33 @@ const SubjectHistoricalDataModal = ({ subjectId, subjectIsStatic, title }) => {
...
 
+  useEffect(() => {
+    if (lastProcessedPositionDate.current === subjectLastPositionDate) {
+      return;
+    }
+    lastProcessedPositionDate.current = subjectLastPositionDate;
+
+    if (activePage !== 1 || !subject) {
+      return;
+    }
```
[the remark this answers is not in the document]
jeslefcourt: Intentional, leaving as-is. Updating the ref before the page guard is deliberate: when the user returns to page 1, `activePage` changes and re-triggers the fetch effect (deps `[activePage, dispatch, subjectId]`), which refetches page 1 from the API including any observations that arrived while on page > 1 — so the update is not lost, it comes back via the refetch rather than the prepend. Moving the ref update after the guard would cause double-handling (prepend + refetch → duplicate row) on return to page 1. The only residual gap is a narrow race where the socket beats backend persistence, which self-heals on the next update or page change.

src/SubjectHistoricalDataModal/index.js
```diff
@@ -107,6 +123,11 @@ const SubjectHistoricalDataModal = ({ subjectId, subjectIsStatic, title }) => {
   const [observationProperties, setObservationProperties] = useState([]);
   const [subjectObservations, setSubjectObservations] = useState([]);
 
+  const subject = useSelector((state) => state.data?.subjectStore?.[subjectId]);
```
luixlive: nit: we usually keep the content of our components organized by first having third party hooks, then selectors, then React state:
```Javascript
const SubjectHistoricalDataModal = ({ subjectId, subjectIsStatic, title }) => {
  const { t } = useTranslation('subjects', { keyPrefix: 'subjectHistoricalDataModal' });

  const subject = useSelector((state) => state.data?.subjectStore?.[subjectId]);

  const [activePage, setActivePage] = useState(1);
  const [loading, setLoadState] = useState(true);
  const [observationsCount, setObservationsCount] = useState(1);
  const [observationProperties, setObservationProperties] = useState([]);
  const [subjectObservations, setSubjectObservations] = useState([]);

  const subjectLastPositionDate = subject?.last_position_date;

  const lastProcessedPositionDate = useRef(subjectLastPositionDate);

  ...
```

src/SubjectHistoricalDataModal/index.test.js
```diff
@@ -8,11 +8,30 @@ import { fetchObservationsForSubject } from '../ducks/observations';
 import { mockStore } from '../__test-helpers/MockStore';
 import { GPS_FORMATS } from '../utils/location';
 import mockedObservationsData from '../__test-helpers/fixtures/observations';
-import { render, waitFor, screen } from '../test-utils';
+import { render, waitFor, screen, act } from '../test-utils';
 import { epsg5367 } from '../__test-helpers/fixtures/location';
 
 import SubjectHistoricalDataModal, { ITEMS_PER_PAGE, getObservationUniqProperties, SORT_BY } from './';
 
+// A store whose state can be mutated in place without swapping the store
```
luixlive, line 16: We can simplify this block to:
```Javascript
// A store whose state can be mutated in place without swapping the store
// instance. Keeps dispatch stable and useSelector re-reads.
```

luixlive (approved):
A couple nitpicks.

src/SubjectHistoricalDataModal/index.js
jeslefcourt: Good call — moved the `useSelector` above the `useState` block so it follows the hooks → selectors → state ordering. Done in 3c1268f5.

src/SubjectHistoricalDataModal/index.test.js
jeslefcourt: Simplified to the two-line version. Done in 3c1268f5.

PR #1624, opened 2026-06-30
fix: isolate per-analyzer failures so one bad analyzer can't empty the Analyzers list (ERA-13536)

jeslefcourt:
## What

Fault-isolates per-analyzer processing in `fetchAnalyzers` so a single failing analyzer can no longer wipe out the entire Analyzers list.

Jira: [ERA-13536](https://allenai.atlassian.net/browse/ERA-13536)

## Why

`fetchAnalyzers` builds the analyzer list inside one `Promise.all(data.map(async (analyzer) => {...}))`. If processing **any one** analyzer throws, the whole `Promise.all` rejects, `FETCH_ANALYZERS_SUCCESS` is never dispatched, and the Analyzers sidebar list stays at its empty initial state — with no user-facing error.

This is happening on **wildlifeactdev.pamdas.org**: a proximity analyzer ("Tourist Road Proximity - Line") is configured with `threshold_dist_meters: -1.0` over a line geometry. The proximity branch buffers the feature by `threshold / 1000` km; negative-buffering a line returns `undefined` from turf, so `proximityPoly.geometry` throws a `TypeError` — and that single bad analyzer made *every* analyzer (KPR Geofence Analyzer included) disappear from the list.

## Change

- Wrap each analyzer's async processing in `try/catch`. On error, `console.warn` (name + id + error) and return `null` instead of rejecting the shared `Promise.all`.
- Filter out the `null` (failed) analyzers before the existing empty-features check.
- Proximity buffer logic is intentionally **unchanged** — guarding invalid `threshold_dist_meters` is a tracked follow-up, as is backend/admin validation.

## Tests

Added `src/ducks/analyzers.test.js` (MSW + `setupServer`, matching the repo's duck-test pattern). It reproduces the real-world failure (a valid geofence, a valid proximity, and a negative-buffer line proximity) and asserts:
- the two valid analyzers are still dispatched and the failing one is excluded;
- exactly one `console.warn` fires naming the failed analyzer.

```
Test Suites: 1 passed, 1 total
Tests:       2 passed, 2 total
```

[ERA-13536]: https://allenai.atlassian.net/browse/ERA-13536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ducks/analyzers.js
+++ b/src/ducks/analyzers.js
@@ -20,41 +20,48 @@ export const fetchAnalyzers = () => async (dispatch) => {
   const { data: { data } } = await axios.get(ANALYZERS_API_URL, { params: { active: true } });
 
   const analyzers = await Promise.all(data.map(async (analyzer) => {
-    const spatialEntries = Object.entries(analyzer.spatial_groups);
-    const nonNullLinks = spatialEntries.filter(entry => entry[1] !== null);
-    const fetchedFeatures = await Promise.all(nonNullLinks.map(async (link) => {
-      const { data: result } = await axios.get(link[1]);
-      const concatFeatures = [];
-      result.data.features.forEach((analyzerFeature) => {
-        const feature = analyzerFeature.features[0];
-        feature.id = featureLayerIdentifier++;
-        if (analyzer.analyzer_category === 'proximity') {
+    // isolate per-analyzer processing so one bad analyzer can't reject the
+    // whole Promise.all and wipe out the entire analyzer list
+    try {
+      const spatialEntries = Object.entries(analyzer.spatial_groups);
+      const nonNullLinks = spatialEntries.filter(entry => entry[1] !== null);
+      const fetchedFeatures = await Promise.all(nonNullLinks.map(async (link) => {
+        const { data: result } = await axios.get(link[1]);
+        const concatFeatures = [];
+        result.data.features.forEach((analyzerFeature) => {
+          const feature = analyzerFeature.features[0];
+          feature.id = featureLayerIdentifier++;
+          if (analyzer.analyzer_category === 'proximity') {
 
-          const proximityPoly = buffer(
-            feature,
-            (analyzer?.threshold_dist_meters ?? DEFAULT_PROXIMITY_ANALYZER_DISPLAY_RADIUS) / 1000,
-            {
-              steps: 32,
-              units: 'kilometers',
-            }
-          );
-          feature.geometry = proximityPoly.geometry;
-        }
-        feature.properties.admin_href = analyzer.admin_href;
-        feature.properties.title = analyzer.name;
-        feature.properties.analyzer_type = analyzer.analyzer_category;
-        feature.properties.spatial_group = feature.geometry.type + '.' + link[0];
-        feature.properties.id = feature.properties.pk;
-        concatFeatures.push(feature);
-      });
-      return concatFeatures;
-    }));
-    // flatten the feature array -
-    const features = [].concat(...fetchedFeatures);
-    return { id: analyzer.id, name: analyzer.name, type: analyzer.analyzer_category, geojson: featureCollection(features) };
+            const proximityPoly = buffer(
+              feature,
+              (analyzer?.threshold_dist_meters ?? DEFAULT_PROXIMITY_ANALYZER_DISPLAY_RADIUS) / 1000,
+              {
+                steps: 32,
+                units: 'kilometers',
+              }
+            );
+            feature.geometry = proximityPoly.geometry;
+          }
+          feature.properties.admin_href = analyzer.admin_href;
+          feature.properties.title = analyzer.name;
+          feature.properties.analyzer_type = analyzer.analyzer_category;
+          feature.properties.spatial_group = feature.geometry.type + '.' + link[0];
+          feature.properties.id = feature.properties.pk;
+          concatFeatures.push(feature);
+        });
+        return concatFeatures;
+      }));
+      // flatten the feature array -
+      const features = [].concat(...fetchedFeatures);
+      return { id: analyzer.id, name: analyzer.name, type: analyzer.analyzer_category, geojson: featureCollection(features) };
+    } catch (error) {
+      console.warn(`error processing analyzer "${analyzer.name}" (${analyzer.id})`, error);
+      return null;
+    }
   }));
 
-  const results = analyzers.filter(item => !!item.geojson.features.length);
+  const results = analyzers.filter(item => !!item && !!item.geojson.features.length);
 
   dispatch({
     type: FETCH_ANALYZERS_SUCCESS,
--- a/src/ducks/analyzers.test.js
+++ b/src/ducks/analyzers.test.js
@@ -1,0 +1,117 @@
+import { http, HttpResponse } from 'msw';
+import { setupServer } from 'msw/node';
+
+import { fetchAnalyzers, FETCH_ANALYZERS_SUCCESS } from './analyzers';
+
+const GOOD_GEOFENCE_LINK = 'http://localhost/spatial-group/good-geofence';
+const BAD_PROXIMITY_LINK = 'http://localhost/spatial-group/bad-proximity';
+const GOOD_PROXIMITY_LINK = 'http://localhost/spatial-group/good-proximity';
+
+const makeSpatialFeature = (geometry) => ({
+  features: [{
+    type: 'Feature',
+    geometry,
+    properties: { pk: 'feature-pk' },
+  }],
+});
+
+const goodGeofenceAnalyzer = {
+  id: 'geofence-id',
+  name: 'Good Geofence',
+  analyzer_category: 'geofence',
+  admin_href: 'http://localhost/admin/geofence',
+  spatial_groups: { warning: GOOD_GEOFENCE_LINK },
+};
+
+// negative threshold over a line geometry: buffer() returns undefined, so
+// reading proximityPoly.geometry throws a TypeError for this analyzer only
+const badProximityAnalyzer = {
+  id: 'bad-proximity-id',
+  name: 'Tourist Road Proximity - Line',
+  analyzer_category: 'proximity',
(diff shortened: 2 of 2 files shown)


src/ducks/analyzers.js
```diff
@@ -20,41 +20,48 @@ export const fetchAnalyzers = () => async (dispatch) => {
...
-    const fetchedFeatures = await Promise.all(nonNullLinks.map(async (link) => {
-      const { data: result } = await axios.get(link[1]);
-      const concatFeatures = [];
-      result.data.features.forEach((analyzerFeature) => {
-        const feature = analyzerFeature.features[0];
-        feature.id = featureLayerIdentifier++;
-        if (analyzer.analyzer_category === 'proximity') {
+    // isolate per-analyzer processing so one bad analyzer can't reject the
+    // whole Promise.all and wipe out the entire analyzer list
+    try {
```
luixlive, line 75: Claude:
Try/catch isolates failures per-analyzer, not per spatial-group link, so one bad link still discards sibling links' valid geometry for that analyzer.

src/ducks/analyzers.js
```diff
@@ -20,41 +20,48 @@ export const fetchAnalyzers = () => async (dispatch) => {
...
+          feature.properties.spatial_group = feature.geometry.type + '.' + link[0];
+          feature.properties.id = feature.properties.pk;
+          concatFeatures.push(feature);
+        });
+        return concatFeatures;
+      }));
+      // flatten the feature array -
+      const features = [].concat(...fetchedFeatures);
+      return { id: analyzer.id, name: analyzer.name, type: analyzer.analyzer_category, geojson: featureCollection(features) };
+    } catch (error) {
```
luixlive, line 77: Claude:
The new catch-all has no `isCancel()` guard, so a benign axios request cancellation is misreported as an analyzer failure.

src/ducks/analyzers.js
```diff
@@ -20,41 +20,48 @@ export const fetchAnalyzers = () => async (dispatch) => {
   const { data: { data } } = await axios.get(ANALYZERS_API_URL, { params: { active: true } });
 
   const analyzers = await Promise.all(data.map(async (analyzer) => {
```
luixlive: This approach can be simplified using `Promise.allSettled` instead of a `Promise.all` with a manual implementation of error handling 😄

src/ducks/analyzers.test.js
```diff
@@ -0,0 +1,117 @@
...
+};
+
+// negative threshold over a line geometry: buffer() returns undefined, so
+// reading proximityPoly.geometry throws a TypeError for this analyzer only
+const badProximityAnalyzer = {
+  id: 'bad-proximity-id',
+  name: 'Tourist Road Proximity - Line',
+  analyzer_category: 'proximity',
+  admin_href: 'http://localhost/admin/bad-proximity',
+  threshold_dist_meters: -1.0,
```
luixlive, line 50: Claude:
The 'bad proximity' test fixture relies on an unasserted, version-specific `turf.buffer()` quirk (returning undefined for a negative buffer on a `LineString`) instead of an explicit, documented failure contract.

src/ducks/analyzers.js
jeslefcourt: Good catch — went one level finer than link-level, since the actual ERA-13536 failure is per-feature (a bad buffer), so link-level alone would still drop good sibling features in the same group.

There are now three tiers of isolation, each a named helper: `fetchSpatialGroupFeatures` (one unusable feature doesn't discard its group), `fetchAnalyzerFeatures` (one unusable group doesn't discard its siblings), `fetchAnalyzers` (one unusable analyzer doesn't discard the list). Covered by the new `a failed spatial group does not discard its sibling groups` and `an unusable feature does not discard the rest of its spatial group` tests.

Fixed in d983872.

src/ducks/analyzers.js
jeslefcourt: Fixed in d983872, following the `ducks/tracks` precedent. Cancellations are rethrown past the per-group handler and the thunk returns early without dispatching.

This also caught a second bug the original version introduced: the blanket catch would have swallowed a cancel and dispatched a *partial* list, clobbering the store on teardown. Now the existing list is left alone. New test: `a cancelled request leaves the existing list alone and is not reported as a failure`.

src/ducks/analyzers.js
jeslefcourt: Agreed 😄 — `Promise.allSettled` at both fan-out points now, replacing the try/catch → `return null` → `.filter(Boolean)` dance. Fixed in d983872.

Also took the Boy Scout pass: `fetchAnalyzers` went from one ~40-line nested block to four named helpers, `link[0]`/`link[1]` destructured to `[spatialGroupName, url]`, and the magic numbers named (`PROXIMITY_BUFFER_STEPS`, `METERS_PER_KILOMETER`).

src/ducks/analyzers.test.js
jeslefcourt: Fixed by giving the production code an explicit contract instead of leaning on turf's return value: `proximityFootprintGeometry` now throws ``could not buffer ${type} geometry by ${n}m`` when turf yields no geometry, rather than letting a TypeError fall out downstream. The test asserts that exact message, so a turf upgrade that changes negative-buffer behaviour fails loudly instead of silently testing nothing.

One caveat worth naming: the fixture still depends on turf returning nothing for the negative-buffer case. It's now load-bearing in an assertion rather than unasserted, which is the part that was actually risky. (Related: turf also returns nothing for a negative-buffered *point*, so the `bufferable sibling` fixture uses a polygon — a negative buffer just shrinks it.)

Fixed in d983872.

jeslefcourt:
@luixlive Requested changes made.

jeslefcourt:
Thanks — validated all of these against the code, and every one held up. Fixed in 52d41e31c, plus a `develop` merge.

**1. Per-feature isolation doesn't isolate** — right, and the blast radius is bigger than the group. I reproduced it: one `features: null` entry sitting between two good ones drops the *whole analyzer from the list*, because the group rejects, the analyzer ends up with zero features, and the `features.length` filter then removes it. `toAnalyzerLayerFeature` now takes the raw entry and validates inside the `try`.

One refinement: `features: []` was already isolated correctly — destructuring `[feature]` from `[]` yields `undefined` and throws *inside* the try. It's specifically `null`/missing, where the array pattern itself throws in the parameter list.

**2. `?? DEFAULT` doesn't cover `0` or negative** — confirmed against turf:

| geometry | 0 m | −1 m |
| --- | --- | --- |
| Point | `undefined` | `undefined` |
| LineString | `undefined` | `undefined` |
| Polygon | `Polygon` | `Polygon` (shrunken) |

Kept these as reported failures rather than defaulting the radius, so the ticket's real-world case stays a failure. `proximityFootprintGeometry` now rejects anything that isn't a finite positive number up front, which closes the polygon hole — `-1m` over a polygon fails the same way it does over a line, instead of drawing a footprint smaller than the geometry it covers. `null`/`undefined` still falls back to 500 m, and there's a test for that.

**3. An empty result now persists** — fixed, and it turned out to interact with item 7. Suppressing the dispatch outright meant a total failure could never *report* itself. So the thunk always dispatches, carrying `hasFailures`, and the reducer owns the rule: keep the existing list when a load produced nothing but failures. An unreachable or non-array list response now dispatches `FETCH_ANALYZERS_ERROR` instead of rejecting into the uncaught `dispatch(fetchAnalyzers())` at `App.js:145`.

**4. Pre-existing `findAnalyzerIdByChildFeatureId` bug** — confirmed end to end: `selectors/index.js:23` returns `.data`, `Map/index.js:125` consumes that selector, and the util then did `analyzerFeatures?.data ?? []` on an array. Always `[]`, always `null`. Now searches what the caller passes, with a new test file.

**5. Fixtures use spatial-group keys that don't exist** — renamed to `warning_group` / `critical_group` / `proximity_group`, which enabled exactly the assertion you pointed at: a buffered proximity feature comes out as `Polygon.proximity_group`.

**6. Dispatched payload shape is unasserted** — confirmed by mutation, then closed. All seven now fail:

| Deleted / blanked | Before | After |
| --- | --- | --- |
| `feature.id` | 5/5 green | 1 failure |
| `admin_href` | 5/5 green | 1 failure |
| `title` | 5/5 green | 1 failure |
| `analyzer_type` | 5/5 green | 1 failure |
| `properties.id` | 5/5 green | 2 failures |
| `name` | 5/5 green | 1 failure |
| `type` | 5/5 green | 1 failure |

**7. No user-facing error state** — added an error banner with `role="alert"` and a retry button, following the GearTab pattern, distinguishing total from partial failure. Three keys across all six locales, translated against each locale's existing wording for "analyzers" and reusing its GearTab retry label.

**Minor**
- Ticket number and narration removed; the test file is rewritten.
- Lazy `import('@turf/buffer')` dropped — the chain `App.js:32 → Map/index.js:68 → ClustersLayer/index.js:13 → useClusterPolygon/index.js:2 → @turf/turf → @turf/buffer → @turf/jsts` is fully static with no `lazy()` on it, so it bought nothing. Removing it also deletes the unguarded `.some()`, which was the line that actually threw first on a nullish list entry — `warnAnalyzerFailure` never got the chance. It's guarded now too.

**One tradeoff worth naming:** `hasFailures` lives in the persisted slice, so after a failed session it rehydrates as `true` and the banner can flash before the on-mount fetch resolves. I left it that way since the list being shown really is the incomplete one, but a persist transform would drop it on rehydrate if you'd prefer.

Also merged `develop`: the branch was 33 commits behind and pinned `I18N_FILES_VERSION` at `1.52` against develop's `1.54`, so merging as-is would have rolled it backwards. Resolved to `1.55`.

Full suite green (333 suites / 3675 tests). ESLint clean on every changed file; stylelint unchanged — the one error in `MapLayersTab/styles.module.scss` is a pre-existing parse error on a commented-out `z-index`.

PR #1628, opened 2026-07-01
feat: show observation details on track point popup (ERA-13537)

jeslefcourt:
## Summary

Resolves [ERA-13537](https://allenai.atlassian.net/browse/ERA-13537) — customers want to see per-position telemetry (e.g. speed) when clicking a point along a subject's historic track ("snail trail").

Clicking a track timepoint now fetches the underlying observation for that subject + time and exposes its `device_status_properties` via the same **"more details"** toggle already used by the subject popup.

## Changes

- **`TimepointPopup`** now fetches the observation by `subject_id` over a small time window bracketing the point, then selects the matching record. Matching is done by **instant**, not string: the track endpoint serializes times in UTC (`...+00:00`) while observation `recorded_at` uses the tenant-local offset (`...-07:00`) — the same instant, different string. A closest-in-window fallback handles sub-second rounding (track times are whole-second; `recorded_at` carries milliseconds).
- **New shared `AdditionalDeviceProperties` component** extracts the toggle + `device_status_properties` list + `localStorage` persistence that previously lived inline in `SubjectPopup`. Both popups now consume it, so the control stays consistent and the preference is shared.
- **`SubjectPopup`** refactored to use the shared component. Its extra behaviors (static-subject gating, time-slider active/empty rendering) are preserved via optional props that default off, so `TimepointPopup` is unaffected.
- 6px of spacing between the location and the details in the track point popup (collapses when there are no details to show).

## Verification

- Unit tests: `TimepointPopup` (7) + `SubjectPopup` (5) pass. New tests lock in the instant-vs-string matching and the non-zero query window as regression guards (both fail against the pre-fix code).
- **Verified against a live tenant (easterisland):** across sampled tracked subjects, the popup resolves the correct observation for a clicked point 8/8, with device properties shown where the observation has them. This live check is what surfaced the two matching bugs (UTC-vs-local string mismatch, and a zero-width `since==until` query returning nothing), which are fixed here.
- ESLint clean on all touched files.

## Notes for reviewers

- The observation query uses a ±1s window; the nearest observation to a track point was always within 999ms in live data. If a source ever emits denser-than-1s data, the exact-instant match still disambiguates.
- `AdditionalDeviceProperties` returns `null` when there are no device properties, so no empty control or stray spacing renders.

[ERA-13537]: https://allenai.atlassian.net/browse/ERA-13537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/TimepointPopup/styles.module.scss
```diff
@@ -0,0 +1,3 @@
+.additionalProperties {
+  margin-top: 6px;
+}
```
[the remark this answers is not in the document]
jeslefcourt: Intentional — the spacing was finalized at 6px and the PR description has been updated to match. Resolving.

src/AdditionalDeviceProperties/index.js
```diff
@@ -0,0 +1,67 @@
...
+  const additionalPropsShouldBeToggleable = hasAdditionalDeviceProps
+    && deviceStatusProperties.length > 2
+    && !staticSubject;
+  const showAdditionalProps = hasAdditionalDeviceProps
+    && (additionalPropsShouldBeToggleable ? additionalPropsToggledOn : true);
+
+  const toggleShowAdditionalProperties = useCallback(() => {
+    toggleAdditionalPropsVisibility(!additionalPropsToggledOn);
+
+    window.localStorage.setItem(STORAGE_KEY, !additionalPropsToggledOn);
```
luixlive: nit / question: Do we really want a persistent local storage variable? It seems to me like a React state variable per popup could be enough. Persisting something so simple sounds like an overkill.

src/AdditionalDeviceProperties/index.js
```diff
@@ -0,0 +1,67 @@
...
+      className={`${styles.additionalProperties} ${timeSliderActive ? styles.disabled : ''}`}
+      data-testid="additional-props"
+      >
+      {deviceStatusProperties.map((deviceStatusProperty, index) => <li
+        key={`${deviceStatusProperty.label}-${index}`}
+      >
+        <strong>{deviceStatusProperty.label}</strong>
+
+        {timeSliderActive ? <span>{t('noHistoricalDataSpan')}</span> : <span data-testid="additional-props-value">
+          {deviceStatusProperty.value.toString()}
```
luixlive: `deviceStatusProperty.value.toString()` will throw an error when a device has a null value.

src/AdditionalDeviceProperties/index.js
```diff
@@ -0,0 +1,67 @@
...
+
+          <span> {deviceStatusProperty.units}</span>
+        </span>}
+      </li>)}
+    </ul>}
+
+    {additionalPropsShouldBeToggleable && <Button
```
luixlive: One nit: I've been trying to use native <button> instead of Bootstrap <Buttons> to have more control over the styling and rendering.

One request: add `aria-expanded` and `aria-controls` props to the button, so screen reader get a signal that the item expanded or collapsed.

src/AdditionalDeviceProperties/styles.module.scss
```diff
@@ -0,0 +1,32 @@
...
+    opacity: 0.6;
+    pointer-events: none;
+  }
+
+  li {
+    display: flex;
+    justify-content: space-between;
+
+    &:not(:last-child) {
+      border-bottom: 0.06px solid $gray-border;
```
luixlive: Better to make this `1px` to avoid an invisible border in small screens and keep the UI consistent.

src/AdditionalDeviceProperties/index.js
luixlive, line 1: This component is missing its unit tests file 😄

src/SubjectPopup/index.js
```diff
@@ -133,35 +113,11 @@ const SubjectPopup = ({ data }) => {
...
-        size="sm"
-        type="button"
-        variant="link"
-        >
-        {t(`additionalPropsButton.${additionalPropsToggledOn ? 'fewer' : 'more'}`)}
-      </Button>}
-    </>}
+    <AdditionalDeviceProperties
+      deviceStatusProperties={device_status_properties}
+      staticSubject={isStatic}
```
luixlive: `staticSubject` and `timeSliderActive` seem to be boolean props. We should prefix them with `is`: `isStaticSubject` and `isTimeSliderActive` so the content and intent is readable.

src/TimepointPopup/index.js
```diff
@@ -1,35 +1,102 @@
...
-      },
-      reportedById: data.properties.id,
-      time: data.properties.time,
-    }}
-    showLabel={false}
-  />
-</>;
+import * as styles from './styles.module.scss';
+
+// The track point's timestamp and the observation's `recorded_at` describe the same
```
luixlive: Too verbose. Explaining the "why" of the variable is enough:
```Javascript
// The backend returns nothing for a zero-width since==until query, so bracket
// the track point's timestamp with a small window.
```

src/TimepointPopup/index.js
```diff
@@ -1,35 +1,102 @@
...
+
+    (async () => {
+      try {
+        const { results } = await dispatch(fetchObservationsForSubject({
+          subject_id: subjectId,
+          since: new Date(targetTime - TIME_WINDOW_MS).toISOString(),
+          until: new Date(targetTime + TIME_WINDOW_MS).toISOString(),
+          include_empty_location: true,
+          page_size: 10,
+        }));
```
luixlive, line 45: Two comments from Claude:

- `include_empty_location: true` makes location-less status-only pings eligible for the match, so a ping sharing the point's `recorded_at` can beat the actual position fix.

- The query passes no `sort_by` and caps at `page_size: 10` for a 2-second window, so a chatty device can push the target observation off the returned page.

src/TimepointPopup/index.js
```diff
@@ -1,35 +1,102 @@
...
+
+        if (ignore) return;
+
+        const observations = results ?? [];
+        const exactMatch = observations.find(
+          (observation) => new Date(observation.recorded_at).getTime() === targetTime
+        );
+        const closestMatch = observations.reduce((closest, observation) => {
```
luixlive: Claude:
The closest-match reduce accepts any observation within the ±1s window unconditionally, so a neighbouring observation's telemetry can be silently attributed to the clicked track point.

src/TimepointPopup/index.js
```diff
@@ -1,35 +1,102 @@
...
+          const distance = Math.abs(new Date(observation.recorded_at).getTime() - targetTime);
+          if (!closest || distance < closest.distance) return { observation, distance };
+          return closest;
+        }, null);
+
+        const matchedObservation = exactMatch ?? closestMatch?.observation;
+        setDeviceStatusProperties(matchedObservation?.device_status_properties ?? []);
+      } catch {
```
luixlive: Claude: 
The empty `catch {}` makes a failing `/observations` endpoint indistinguishable from a point that genuinely has no telemetry.

luixlive (changes requested):
A bunch of small comments. Blocking specially because of the missing tests file. Claude should be able to tackle all of them in an iteration 👍

src/AdditionalDeviceProperties/index.js
jeslefcourt: @luixlive Can you please update the agents with this direction so it picks up that difference in future work?

src/AdditionalDeviceProperties/styles.module.scss
jeslefcourt: @luixlive Another good instruction for the coding agent.

src/AdditionalDeviceProperties/index.js
luixlive: Will do. I recently updated the AGENTS.md in the form-builder repo with more granular development preferences. Hopefully that helps here too 👍

jeslefcourt:
@luixlive Fixes made.

jeslefcourt:
Validated all six against the code and fixed them in d78a80fd9, plus a `develop` merge. Notes where I found more or less than reported.

**1. Crash on null `device_status_properties`** — confirmed, and I reproduced the whole chain:

```
JSON.stringify(null)                    -> "null"     (ducks/popup.js:64)
getDeviceStatusPropertiesForSubject     -> null       (JSON.parse succeeds; the ?? [] never fires)
render(deviceStatusProperties={null})   -> TypeError: Cannot read properties of null (reading 'length')
```

Normalized inside the component with `Array.isArray`, and a property with no value now renders `—` instead of `Battery %`.

One correction: **TimepointPopup isn't on the crash path.** Its `?? []` already guarantees an array, so the null only arrives through SubjectPopup on socket updates. Normalizing in the component is still right.

**2. Wrong observation matched** — all three confirmed, one with a nuance.

The window is now 5s against a 1s tolerance, so a neighbour can actually be rejected rather than being the only thing the query could have returned. Candidates with an unparseable `recorded_at` are filtered before the reduce. And observations equally close to the point now match neither, which is the case that was letting a device reporting twice in a second pick up its neighbour's telemetry.

The NaN behaviour is order-dependent, which is worth recording:

| response order | result |
| --- | --- |
| unparseable first | picks it, `distance: NaN`, tolerance rejects → **every valid match dropped** |
| unparseable second | unaffected |

Once `NaN` becomes `closest`, every later `distance < NaN` is false, so nothing displaces it.

The socket point I confirmed at runtime before touching it:

```
socket carried time       : 2021-01-27T09:09:30+00:00
newest point .time        : 2021-01-27T09:04:25+00:00   <- previous point
track times[0]            : 2021-01-27T09:09:30+00:00   <- correct
```

`merge({}, points.features[0].properties, update.properties)` seeds from the previously newest point, and the payload carries its time under `coordinateProperties`, not as a flat `time`. Fixed there, with tests.

**3. Missing `isCancel` guard** — added, following the `ducks/tracks` convention, and the `console.warn` moved inside the ignore check.

**4. `aria-controls` dangles** — the list is now always mounted with `hidden` toggled, placed after the button. Note this moves the "more details" toggle above the list in **SubjectPopup** too, not just the track point popup. It reads fine as an accordion header, but it's a visible change on a screen this PR wasn't otherwise touching — worth a look before merge, and easy to revert to list-first if you'd rather (the dangling IDREF is fixed either way by always rendering).

**5. Three new tests can't fail** — one of the three, and I fixed all of them. I mutation-tested each rather than taking it on faith:

| Mutation | Result |
| --- | --- |
| impl adds `include_empty_location: true` | **1 failure** — the test does catch it |
| `page_size` 100 → 10 | **1 failure** — caught |
| tolerance check removed | **1 failure** — caught |
| `reportedById` + `time` deleted from `reportData` | **9/9 green** — genuinely uncovered |

So the `reportData` call is right: `systemConfig: {}` makes `AddItemButton` return `null` at line 65, and both new fields could be deleted with nothing failing. Fixed by enabling `events_enabled` in the test store and asserting the report data.

The other two assertions do fail when their code breaks, so "can't fail" isn't quite it. Your underlying point about the tolerance test stands though, and it's the sharper one: it was pinning a branch that **couldn't be reached in production**, because the window and the tolerance were both 1000ms. Fixing item 2 is what gives that test meaning. I also found the window width itself wasn't pinned at all — 1000 → 60000 left everything green — so there's now a test that the window exceeds the tolerance.

**6. Cleanups** — all three done. `useLocalStorage` is byte-compatible as you say; I also wrapped its `getItem` in a try/catch, since it `JSON.parse`s unguarded and would now throw during render on any pre-existing garbage under that key. Worth knowing it writes on mount through its effect, so opening a popup persists the default where previously nothing was written until the user toggled — I took that as acceptable, since it's how every other consumer of the hook already behaves.

For the button: `@mixin linkButton` already exists in `common/styles/_buttons.scss`, so I used it rather than adding one — 16 declarations down to 6, keeping only the ones carrying intent (link colour, size, radius, focus ring). And the list's negative margin is gone, so the popup's `margin-top: 6px` is now actually 6px.

**Two things outside the list.** A pre-existing smell in `SubjectPopup/index.test.js` surfaced: it renders a popup in `beforeEach` and again inside two tests, which was invisible while collapsed lists were unmounted but means `getByTestId` now finds two. Scoped those with `within(container)`.

And the branch was 160 commits behind `develop`, far enough that its tests wouldn't load at all against a current `node_modules` — `src/test-utils.jsx` imports `react-router`, which develop added to `transformIgnorePatterns`. CI here was green only because it last ran in August against the then-current lockfile. Merged develop; no locale changes, so no `I18N_FILES_VERSION` bump needed.

Full suite green (335 suites / 3707 tests). ESLint unchanged on the touched files — the 3 errors there are pre-existing. Stylelint went from 4 problems to 2 on the styles I touched, both the repo-wide kebab-case rule that contradicts the camelCase convention in AGENTS.md.

PR #1629, opened 2026-07-01
ERA-13445: Timeslider speed control

luixlive:
## What does this PR do?
Add playback speed options to TimeSlider component.
- Introduced a speed button to open playback speed options.
- Added playback speed options (0.5x, 0.75x, 1x, 1.25x, 1.5x, 2x) with corresponding labels in multiple languages.
- Implemented functionality to select and apply playback speed, including keyboard navigation for accessibility.
- Updated styles for the speed menu and button to enhance user experience.
- Added tests to verify the speed button and menu functionality.

## Evidence
image

### Relevant link(s)
- [ERA-13445](https://allenai.atlassian.net/browse/ERA-13445)
- [Branch Environment](https://era-13445.dev.pamdas.org)

[ERA-13445]: https://allenai.atlassian.net/browse/ERA-13445?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/public/locales/en-US/components.json
+++ b/public/locales/en-US/components.json
@@ -382,6 +382,10 @@
     "popoverHeader": "Date Range",
     "popoverResetButton": "Reset",
     "sliderLabel": "Timeslider",
+    "speedButtonLabel": "Open playback speed options",
+    "speedMenuHeader": "Playback Speed",
+    "speedMenuLabel": "Playback speed options",
+    "speedMenuOptionLabel": "Set playback speed to {{speed}}",
     "timeSliderMapControl": {
       "iconTitle": "Timeslider"
     },
--- a/public/locales/es/components.json
+++ b/public/locales/es/components.json
@@ -359,6 +359,10 @@
     "popoverHeader": "Rango de fechas",
     "popoverResetButton": "Reiniciar",
     "sliderLabel": "Control de tiempo",
+    "speedButtonLabel": "Abrir opciones de velocidad de reproducción",
+    "speedMenuHeader": "Velocidad de Reproducción",
+    "speedMenuLabel": "Opciones de velocidad de reproducción",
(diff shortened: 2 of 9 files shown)


AlanCalvillo:
just a small ux note (which is probably unrelated to this particular piece of work): resizing the subject popup creates a noticeable flickering effect. it looks like rendering the popup is faster than calculating the track length, so the popup gets resized a moment later. maybe we could show a skeleton while the calculation completes, or await for it to finish before rendering the popup.

https://github.com/user-attachments/assets/81499e6d-2eac-4da4-b597-f07964337013

luixlive:
> just a small ux note (which is probably unrelated to this particular piece of work): resizing the subject popup creates a noticeable flickering effect. it looks like rendering the popup is faster than calculating the track length, so the popup gets resized a moment later. maybe we could show a skeleton while the calculation completes, or await for it to finish before rendering the popup.

Thanks @AlanCalvillo , good catch.

The subject popup flickers because of two renders per frame, one before calculating a string and one after the string is there. This makes the popover jump. The bug is not really scoped by this PR, which only added the speed control. It is a two-renders issue that is now more evident when using the auto-play. We should fix it, but I think it's better to do it in its own ticket.

cc: @JoshuaVulcan

chrisj-er:
Agree, the flicker was already there I can see it on easterisland running 2.142. Write that up as a p2 bug to be fixed.

chrisj-er (approved):
This is a clean, well-tested change.

AlanCalvillo:
here is the bug ticket https://allenai.atlassian.net/browse/ERA-13542
cc: @luixlive @chrisj-er

PR #1634, opened 2026-07-02
ERA-13539: adopt Auth0 refresh-token rotation for ER Web token renewal

StephenWithPH:
**Jira:** https://allenai.atlassian.net/browse/ERA-13539

## What
ER Web (Auth0 / `require_idp` sites) renews access tokens via the rotating `refresh_token` grant instead of silent-iframe auth. Two commits:
- Enable `useRefreshTokens` on `Auth0Provider` (auto-requests `offline_access` → Auth0 issues a rotating refresh token).
- Renew on 401 in `RequestConfigManager`: on a retriable 401, `getAccessTokenSilently()` → persist (cookie + Redux) → replay the request once; on renewal failure or a repeat 401, sign out as before.

## Why
Prerequisite for server-side MFA enforcement (ERA-13387): once Server validates the MFA mirror claims on every access token, the client must renew silently (the mirror claims survive the refresh — verified orgless) rather than force a full re-login every ~2 days. Also cookie-independent / more robust than silent-iframe. Mirrors the mobile fix (EM-2286).

## Scope
Auth0 / `require_idp` deployments only. Legacy `grant_type=password` sites are unaffected (they never use Auth0 renewal; `getAccessTokenSilently` throws → same sign-out as today).

## Verification
- `/oauth/token` now returns a `refresh_token` (auth-test-harness recon / DevTools).
- Org-enforced-MFA refresh path unblocked by PADAS/auth0-platform-management#117 (merged) — the `security-policies` Action now no-ops on `oauth2-refresh-token` grants (without it, org refresh returned `403 mfa_required`).
- 6 `RequestConfigManager` unit tests (renew+replay, loop-guard round-trip, renewal-failure→sign-out, no-config, skipAuth, non-401). Full suite green.

## Reviewer notes (context intentionally kept out of the code)
- **Loop guard:** the replay carries a private, client-only `retriedAfterRefresh` config marker so a repeat 401 can't recurse into another refresh. It rides axios's `mergeConfig` — the same mechanism the existing `skipAuth` flag uses (verified in axios 1.17.0).
- **Don't hand-roll the refresh:** concurrent 401s are safe because `getAccessTokenSilently` is serialized/deduped inside auth0-spa-js (one rotation; others use the cached token). A manual mutex or a direct `/oauth/token` call would risk refresh-token reuse-detection revoking the family.
- **cancelToken:** the replay reuses the master cancel token; a concurrent sign-out could cancel an in-flight replay — benign (we're navigating to login anyway).
- **Storage:** the refresh token lives in localStorage (standard `auth0-spa-js` pattern; consistent with existing token storage).
- **`useRefreshTokensFallback` left `false`** deliberately — a failed refresh surfaces as a clean login redirect, not a silent iframe retry.

## Out of scope / follow-ups
- `mfa_max_age` vs the 2-day access-token lifetime (ERA-13387) — future work.
- Test hardening: an MSW real-axios integration test for the loop-guard round-trip + extracting `handle401Errors` into a pure function — follow-up (not filed).

### DIFF
--- a/src/RequestConfigManager/index.js
+++ b/src/RequestConfigManager/index.js
@@ -1,10 +1,11 @@
 import { memo, useCallback, useEffect } from 'react';
 import axios from 'axios';
-import { connect } from 'react-redux';
+import { connect, useDispatch } from 'react-redux';
 import { useLocation } from 'react-router';
+import { useAuth0 } from '@auth0/auth0-react';
 import { toast } from 'react-toastify';
 
-import { clearAuth, resetMasterCancelToken } from '../ducks/auth';
+import { clearAuth, POST_AUTH_SUCCESS, resetMasterCancelToken } from '../ducks/auth';
 
 import { REACT_APP_ROUTE_PREFIX } from '../constants';
 import { showToast } from '../utils/toast';
@@ -58,17 +59,38 @@ const RequestConfigManager = ({
   token,
   user,
 }) => {
-  const location = useLocation();
+  const { search } = useLocation();
   const navigate = useNavigate();
+  const dispatch = useDispatch();
+  const { getAccessTokenSilently } = useAuth0();
 
-  const handle401Errors = useCallback((error) => {
-    if (error && error.toString().includes('401') && !error.config?.skipAuth) {
+  const handle401Errors = useCallback(async (error) => {
+    const isAuthError = error?.response?.status === 401 && !error.config?.skipAuth;
+    const request = error?.config;
+
+    if (isAuthError && request && !request.retriedAfterRefresh) {
+      try {
+        const accessToken = await getAccessTokenSilently();
+        document.cookie = `token=${accessToken};path=/`;
+        dispatch({ type: POST_AUTH_SUCCESS, payload: { data: { access_token: accessToken } } });
+        return axios({
+          ...request,
+          retriedAfterRefresh: true,
+          headers: { ...request.headers, Authorization: `Bearer ${accessToken}` },
+        });
+      } catch (renewalError) {
+        console.warn('Token renewal failed; signing out', renewalError);
+      }
+    }
+
+    if (isAuthError) {
       resetMasterCancelToken();
       clearAuth().then(() => {
-        navigate({ pathname: `${REACT_APP_ROUTE_PREFIX}login`, search: location.search });
+        navigate({ pathname: `${REACT_APP_ROUTE_PREFIX}login`, search });
       });
     }
-  }, [clearAuth, location?.search, navigate, resetMasterCancelToken]);
+    return Promise.reject(error);
+  }, [clearAuth, dispatch, getAccessTokenSilently, navigate, resetMasterCancelToken, search]);
 
   const addMasterCancelTokenToRequests = useCallback((config) => {
     config.cancelToken = config.cancelToken || (masterRequestCancelToken && masterRequestCancelToken.token);
@@ -114,10 +136,7 @@ const RequestConfigManager = ({
         handleGeoPermWarningHeader(response, userLocationAccessGranted);
         return response;
       },
-      (error) => {
-        handle401Errors(error);
-        return Promise.reject(error);
-      }
+      (error) => handle401Errors(error)
     ];
 
     const interceptorId = axios.interceptors.response.use(...interceptorConfig);
--- a/src/RequestConfigManager/index.js
+++ b/src/RequestConfigManager/index.js
@@ -1,11 +1,11 @@
 import { memo, useCallback, useEffect } from 'react';
 import axios from 'axios';
-import { connect, useDispatch } from 'react-redux';
+import { connect } from 'react-redux';
 import { useLocation } from 'react-router';
 import { useAuth0 } from '@auth0/auth0-react';
 import { toast } from 'react-toastify';
 
-import { clearAuth, POST_AUTH_SUCCESS, resetMasterCancelToken } from '../ducks/auth';
+import { applyAccessToken, clearAuth, resetMasterCancelToken } from '../ducks/auth';
 
 import { REACT_APP_ROUTE_PREFIX } from '../constants';
 import { showToast } from '../utils/toast';
@@ -51,6 +51,7 @@ const handleGeoPermWarningHeader = (response, userLocationAccessGranted) => {
 
 
 const RequestConfigManager = ({
+  applyAccessToken,
   clearAuth,
   userLocationAccessGranted,
   masterRequestCancelToken,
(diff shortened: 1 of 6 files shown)


StephenWithPH:
@luixlive and @chrisj-er ... this can wait until next week. The env is deployed at https://era-13539.dev.pamdas.org, and we're properly requesting the scope.

I'm considering advocating for setting out Auth0 dev tenant's access token lifetime to something very short (5 minutes), so we can tease out any refresh token problems much faster.

src/RequestConfigManager/index.js
```diff
@@ -58,17 +59,38 @@ const RequestConfigManager = ({
...
+    if (isAuthError) {
       resetMasterCancelToken();
       clearAuth().then(() => {
-        navigate({ pathname: `${REACT_APP_ROUTE_PREFIX}login`, search: location.search });
+        navigate({ pathname: `${REACT_APP_ROUTE_PREFIX}login`, search });
       });
     }
-  }, [clearAuth, location?.search, navigate, resetMasterCancelToken]);
+    return Promise.reject(error);
+  }, [clearAuth, dispatch, getAccessTokenSilently, navigate, resetMasterCancelToken, search]);
```
luixlive: Now that REST 401s are recoverable via refresh, should the socket's `resp_authorization` 401 handler also attempt `getAccessTokenSilently()` + re-authorization emit before calling `clearAuth()`? It's not directly in the scope of this change, but it's related and could be a nice improvement.

src/RequestConfigManager/index.js
```diff
@@ -58,17 +59,38 @@ const RequestConfigManager = ({
...
+
+  const handle401Errors = useCallback(async (error) => {
+    const isAuthError = error?.response?.status === 401 && !error.config?.skipAuth;
+    const request = error?.config;
+
+    if (isAuthError && request && !request.retriedAfterRefresh) {
+      try {
+        const accessToken = await getAccessTokenSilently();
+        document.cookie = `token=${accessToken};path=/`;
+        dispatch({ type: POST_AUTH_SUCCESS, payload: { data: { access_token: accessToken } } });
```
luixlive: Could this reuse `postAuthSuccess` from `ducks/auth.js` instead of re-implementing the cookie-write + dispatch? It's now duplicated in three places, we could have a single utility.

luixlive (changes requested):
Not hard blockers, but a couple questions / suggestions I'd like you to read before approving. Most of the PR is good and well implemented 👍

src/RequestConfigManager/index.js
StephenWithPH: > Could this reuse `postAuthSuccess` from `ducks/auth.js` instead of re-implementing the cookie-write + dispatch? It's now duplicated in three places, we could have a single utility.

Claude kept suggesting it, and I kept telling Claude "just do what we're here to do"!. I'll have Claude extract.

src/RequestConfigManager/index.js
StephenWithPH: @luixlive Let's talk about the socket side of this at standup tomorrow. I want to make sure that will be graceful with auth step-up.

src/RequestConfigManager/index.js
StephenWithPH: fixed in 7ce9f9de07828636c05a720268302e9b15e7222c

src/RequestConfigManager/index.js
StephenWithPH: Leaning heavily on 🤖 here's a **not implemented** patch:

```patch
diff --git a/src/withSocketConnection/index.js b/src/withSocketConnection/index.js
index e3142622..87d022e3 100644
--- a/src/withSocketConnection/index.js
+++ b/src/withSocketConnection/index.js
@@ -1,4 +1,5 @@
-import React, { createContext, useEffect, useState } from 'react';
+import React, { createContext, useEffect, useRef, useState } from 'react';
+import { useAuth0 } from '@auth0/auth0-react';
 
 import store from '../store';
 import useRealTimeImplementation from './useRealTimeImplementation';
@@ -16,13 +17,19 @@ const WithSocketContext = (props) => {
     socketIO
   } = useRealTimeImplementation();
 
+  const { getAccessTokenSilently } = useAuth0();
+  const getAccessTokenSilentlyRef = useRef(getAccessTokenSilently);
+  useEffect(() => {
+    getAccessTokenSilentlyRef.current = getAccessTokenSilently;
+  }, [getAccessTokenSilently]);
+
   useEffect(() => {
     if (socketIO){
       let socketReconnectTimeout;
 
       const instantiate = () => {
         const socket = createSocket();
-        const socketWithEvents = bindSocketEvents(socket, store);
+        const socketWithEvents = bindSocketEvents(socket, store, getAccessTokenSilentlyRef);
 
         const teardown = () => {
           socketWithEvents.close();
diff --git a/src/withSocketConnection/useRealTimeImplementation/index.js b/src/withSocketConnection/useRealTimeImplementation/index.js
index 4c4bc2e8..07c4bf94 100644
--- a/src/withSocketConnection/useRealTimeImplementation/index.js
+++ b/src/withSocketConnection/useRealTimeImplementation/index.js
@@ -3,7 +3,7 @@ import React, { useCallback, useEffect, useState } from 'react';
 import { DAS_HOST } from '../../constants';
 import { resetSocketStateTracking } from './helpers';
 import { SOCKET_HEALTHY_STATUS } from '../../ducks/system-status';
-import { clearAuth } from '../../ducks/auth';
+import { applyAccessToken, clearAuth } from '../../ducks/auth';
 import { events } from './config';
 import { calcEventFilterForRequest } from '../../utils/event-filter';
 import { calcPatrolFilterForRequest } from '../../utils/patrol-filter';
@@ -44,9 +44,10 @@ const useRealTimeImplementation = () => {
     return [interval, timeout];
   }, []);
 
-  const bindSocketEvents = useCallback((socket, store) => {
+  const bindSocketEvents = useCallback((socket, store, getAccessTokenSilentlyRef) => {
     let eventsBound = false;
     let pingInterval, pingTimeout;
+    let authRetried = false;
 
     socket.on('connect', () => {
       store.dispatch({ type: SOCKET_HEALTHY_STATUS });
@@ -66,9 +67,20 @@ const useRealTimeImplementation = () => {
     socket.on('connect_error', (msg) => {
       console.log('realtime: connection error', msg);
     });
-    socket.on('resp_authorization', (msg) => {
+    socket.on('resp_authorization', async (msg) => {
       const { status } = msg;
       if (status.code === 401) {
+        if (!authRetried && getAccessTokenSilentlyRef?.current) {
+          authRetried = true;
+          try {
+            const accessToken = await getAccessTokenSilentlyRef.current();
+            store.dispatch(applyAccessToken(accessToken));
+            socket.emit('authorization', { type: 'authorization', id: 1, authorization: `Bearer ${accessToken}` });
+            return;
+          } catch (renewalError) {
+            console.warn('realtime token renewal failed; signing out', renewalError);
+          }
+        }
         console.warn('realtime auth rejected', msg);
         return store.dispatch(clearAuth());
       }
```

... but this seems like it may merit a future, follow-on PR by a real React dev!

luixlive (approved):
Thanks for addressing the feedback!

PR #1657, opened 2026-07-15
ERA-13685: Centralize ER Web 401 auth-recovery (shared unit) + websocket silent-renew

StephenWithPH:
## What & why

[ERA-13685](https://allenai.atlassian.net/browse/ERA-13685)

Makes ER Web's websocket auth-rejection handling congruent with the REST client's 401 renewal (shipped in ERA-13539), behind **one shared, single-flight auth-recovery unit** used by both transports. Scope here is **silent renewal only**; the interactive MFA step-up branch is left as an unwired seam for the step-up tickets.

## Commits

1. **Add shared single-flight `auth-recovery` unit** — `recoverAuth({ stepUp, challenge })`. Concurrent 401s from REST + websocket collapse into one token renewal that every caller awaits; the renewed token is applied once via `applyAccessToken`. Auth0 primitives are injected via `registerAuthRecovery` (from the `useAuthRecovery` hook) rather than imported, since they're React-hook-bound. `stepUp` is an intentional unwired seam.
2. **Bound silent renewal with a timeout** — a stalled `getAccessTokenSilently()` rejects after 30s (→ sign-out) instead of hanging, which would otherwise leave the socket connected-but-unauthorized and freeze REST recovery too via the shared single-flight. The interactive step-up path is deliberately **not** time-boxed (it waits on the user completing MFA).
3. **Route REST 401 renewal through the unit** — `RequestConfigManager.handle401Errors` delegates to `recoverAuth()`; no REST behavior change (renew → replay once → sign out on failure).
4. **Silently renew websocket auth on 401** — on a `resp_authorization` 401, renew once via `recoverAuth()` and re-authorize on the same open socket, with a per-cycle `authRetried` guard that resets on a successful (200) authorization. Renewal failure or a repeat 401 falls through to the existing `clearAuth()` sign-out.

## Dependencies / context

- **Foundation for the MFA step-up work.** The step-up handlers build on this unit: **ERA-13405** (REST) and **ERA-13733** (websocket), both under **ERA-13427**.
- **ERA-13733's server side isn't live yet.** The rt_api websocket step-up signal (**ERA-13732**) must ship before the socket can distinguish a step-up 401 from an ordinary one — until then the socket only ever sees a generic 401, which this PR's silent-renew handles correctly.
- A `TODO(ERA-13733)` marks the one interaction the step-up ticket must handle: single-flight must not collapse a step-up into an in-flight silent renewal.

## Testing

- New unit tests for `auth-recovery` (single-flight, apply-once, timeout scoping, empty-token guard, step-up seam).
- `RequestConfigManager` tests rewritten to assert delegation to `recoverAuth`; `ducks/auth` gains a test for `applyAccessToken`'s cookie write (incl. `path=/`) + `POST_AUTH_SUCCESS` dispatch.
- Real `resp_authorization` coverage added to `withSocketConnection` (renew+re-auth, renewal-fails→sign-out, double-401→sign-out, 200-resets-guard).
- Full suite green: 306 suites / 2871 tests.

## Deploy

`deploy` label added to spin up a feature environment for manual verification.

[ERA-13685]: https://allenai.atlassian.net/browse/ERA-13685?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/RequestConfigManager/index.js
+++ b/src/RequestConfigManager/index.js
@@ -2,13 +2,14 @@ import { memo, useCallback, useEffect } from 'react';
 import axios from 'axios';
 import { connect } from 'react-redux';
 import { useLocation } from 'react-router';
-import { useAuth0 } from '@auth0/auth0-react';
 import { toast } from 'react-toastify';
 
-import { applyAccessToken, clearAuth, resetMasterCancelToken } from '../ducks/auth';
+import { clearAuth, resetMasterCancelToken } from '../ducks/auth';
 
 import { REACT_APP_ROUTE_PREFIX } from '../constants';
+import { recoverAuth } from '../utils/auth-recovery';
 import { showToast } from '../utils/toast';
+import useAuthRecovery from '../hooks/useAuthRecovery';
 import useNavigate from '../hooks/useNavigate';
 
 const STARTUP_TIME = new Date();
@@ -51,7 +52,6 @@ const handleGeoPermWarningHeader = (response, userLocationAccessGranted) => {
 
 
 const RequestConfigManager = ({
-  applyAccessToken,
   clearAuth,
   userLocationAccessGranted,
   masterRequestCancelToken,
@@ -62,7 +62,8 @@ const RequestConfigManager = ({
 }) => {
   const { search } = useLocation();
   const navigate = useNavigate();
-  const { getAccessTokenSilently } = useAuth0();
+
+  useAuthRecovery();
 
   const handle401Errors = useCallback(async (error) => {
     const isAuthError = error?.response?.status === 401 && !error.config?.skipAuth;
@@ -70,8 +71,7 @@ const RequestConfigManager = ({
 
     if (isAuthError && request && !request.retriedAfterRefresh) {
       try {
-        const accessToken = await getAccessTokenSilently();
-        applyAccessToken(accessToken);
+        const accessToken = await recoverAuth();
         return axios({
           ...request,
           retriedAfterRefresh: true,
@@ -89,7 +89,7 @@ const RequestConfigManager = ({
       });
     }
     return Promise.reject(error);
-  }, [applyAccessToken, clearAuth, getAccessTokenSilently, navigate, resetMasterCancelToken, search]);
+  }, [clearAuth, navigate, resetMasterCancelToken, search]);
 
   const addMasterCancelTokenToRequests = useCallback((config) => {
     config.cancelToken = config.cancelToken || (masterRequestCancelToken && masterRequestCancelToken.token);
@@ -180,4 +180,4 @@ const mapStateToProps = ({ data: { selectedUserProfile, user, masterRequestCance
 });
 
 
-export default connect(mapStateToProps, { applyAccessToken, clearAuth, resetMasterCancelToken })(memo(RequestConfigManager));
+export default connect(mapStateToProps, { clearAuth, resetMasterCancelToken })(memo(RequestConfigManager));
--- a/src/hooks/useAuthRecovery.js
+++ b/src/hooks/useAuthRecovery.js
@@ -1,0 +1,18 @@
+import { useEffect } from 'react';
+import { useAuth0 } from '@auth0/auth0-react';
+
+import { registerAuthRecovery } from '../utils/auth-recovery';
+
+/**
+ * Registers the live @auth0/auth0-react primitives into the shared auth-recovery unit
+ * so the axios interceptor and the websocket handler can both drive token renewal.
+ */
+const useAuthRecovery = () => {
+  const { getAccessTokenSilently } = useAuth0();
+
+  useEffect(() => {
+    registerAuthRecovery({ silentRenew: () => getAccessTokenSilently() });
+  }, [getAccessTokenSilently]);
+};
+
+export default useAuthRecovery;
--- a/src/utils/auth-recovery.js
+++ b/src/utils/auth-recovery.js
@@ -1,0 +1,46 @@
+import store from '../store';
+import { applyAccessToken } from '../ducks/auth';
+
+/**
+ * Shared, single-flight auth recovery. Concurrent 401s collapse into one in-flight
+ * token renewal that every caller awaits, and the renewed token is applied once.
+ * Auth0 primitives are injected via `registerAuthRecovery` (they are React-hook-bound),
+ * not imported.
+ */
+
+let primitives = { silentRenew: null, stepUp: null };
+
+// The one shared in-flight recovery promise (null when idle).
+let inFlight = null;
+
+export const registerAuthRecovery = (next) => {
+  primitives = { ...primitives, ...next };
+};
+
+export const recoverAuth = ({ stepUp = false, challenge = null } = {}) => {
+  if (!inFlight) {
+    inFlight = (async () => {
+      // Only silent renewal is wired; the stepUp branch is a reserved seam for MFA step-up.
+      const recover = stepUp ? primitives.stepUp : primitives.silentRenew;
+      if (typeof recover !== 'function') {
+        throw new Error(`auth-recovery: no ${stepUp ? 'stepUp' : 'silentRenew'} primitive registered`);
+      }
+
+      const accessToken = await recover(challenge);
+      if (!accessToken) {
+        throw new Error('auth-recovery: renewal returned no token');
+      }
+      store.dispatch(applyAccessToken(accessToken));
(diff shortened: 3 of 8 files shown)


src/withSocketConnection/index.test.js
```diff
@@ -8,4 +31,102 @@ describe('recreating the web socket', () => {
...
+    on: jest.fn((event, fn) => { handlers[event] = fn; }),
+    emit: jest.fn(),
+    handlers,
+  };
+};
+
+const makeStore = (accessToken = 'fresh.token') => ({
+  dispatch: jest.fn(),
+  getState: () => ({ data: { token: { access_token: accessToken }, selectedUserProfile: null } }),
+});
```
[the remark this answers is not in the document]
StephenWithPH: Good call — the test store was an immutable snapshot, so renewal couldn't change the token mid-flow. Fixed in 717cd77c418ea0b3f503deffe60944295359ea68: `makeStore` now holds a mutable token with a `setToken` setter, and `bindSocket` accepts an initial token.

src/withSocketConnection/index.test.js
```diff
@@ -8,4 +31,102 @@ describe('recreating the web socket', () => {
...
+    jest.useRealTimers();
+  });
+
+  test('a 401 silently renews and re-authorizes on the same socket, without signing out', async () => {
+    recoverAuth.mockResolvedValue('fresh.token');
+
+    const { socket, store } = bindSocket();
+    await act(async () => {
+      await socket.handlers.resp_authorization({ status: { code: 401 } });
+    });
```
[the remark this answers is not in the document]
StephenWithPH: Agreed — this was effectively a tautology, since the store was seeded with the same token `recoverAuth` resolved to. Fixed in 717cd77c418ea0b3f503deffe60944295359ea68: the renew test now starts with a stale token and the `recoverAuth` mock calls `store.setToken('fresh.token')` (modeling `applyAccessToken` updating Redux), so asserting `Bearer fresh.token` proves re-auth used the renewed token — plus an explicit assertion that it did not emit `Bearer stale.token`.

src/utils/auth-recovery.js
```diff
@@ -0,0 +1,75 @@
...
+  let timeoutId;
+  const timeout = new Promise((_, reject) => {
+    timeoutId = setTimeout(() => reject(new Error('auth-recovery: renewal timed out')), ms);
+  });
+  try {
+    return await Promise.race([promise, timeout]);
+  } finally {
+    clearTimeout(timeoutId);
+  }
+};
```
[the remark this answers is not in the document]
StephenWithPH: Looked into this, but I don't think the unhandled rejection can occur here, so I'm leaving `withTimeout` as-is. `Promise.race([p, timeout])` calls `.then()` on every input promise as part of racing them, so the renewal promise already has a rejection reaction attached. If it rejects after the timeout has won, that reaction consumes the rejection (a no-op, since the race already settled) rather than letting it surface as unhandled. A defensive `.catch()` would be redundant here.

src/withSocketConnection/index.test.js
```diff
@@ -1,4 +1,27 @@
 import React from 'react';
+import { renderHook, act } from '@testing-library/react';
```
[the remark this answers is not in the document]
StephenWithPH: humans... this was already here, and did not trip anything in GitHub Actions.

src/hooks/useAuthRecovery.js
```diff
@@ -0,0 +1,22 @@
+import { useEffect } from 'react';
+import { useAuth0 } from '@auth0/auth0-react';
+
+import { registerAuthRecovery } from '../utils/auth-recovery';
+
+/**
+ * Registers the live @auth0/auth0-react primitives into the shared auth-recovery
```
luixlive: This comments is to verbose and gives implementation details from the parent and provides ticket numbers. Not idea, better to reduce it only to: `Registers the live @auth0/auth0-react primitives into the shared auth-recovery unit so the axios interceptor and the websocket handler can both drive token renewal.`

src/RequestConfigManager/index.js
```diff
@@ -180,4 +180,4 @@ const mapStateToProps = ({ data: { selectedUserProfile, user, masterRequestCance
 });
 
 
-export default connect(mapStateToProps, { applyAccessToken, clearAuth, resetMasterCancelToken })(memo(RequestConfigManager));
\ No newline at end of file
+export default connect(mapStateToProps, { clearAuth, resetMasterCancelToken })(memo(RequestConfigManager));
```
luixlive, line 183: nit and out of scope, but if you feel like helping a bit this old repo, you put Claude to remove the deprecated `connect` and `mapStateToProps` APIs in favor of `useSelector` and `useDispatch`. I'm sure it will do it in a minute 😄 

The changes seem good here 👍

src/utils/auth-recovery.js
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * Both the REST client (RequestConfigManager's axios interceptor) and the realtime
+ * websocket handler funnel their 401s here, so a burst of concurrent rejections
+ * collapses into a single token renewal that every caller awaits, and the renewed
+ * token is applied to Redux + cookie in exactly one place.
+ *
+ * The live @auth0/auth0-react primitives are injected by `useAuthRecovery` rather
+ * than imported, because they are React-hook-bound. Only silent renewal is wired
+ * today; `stepUp` is an intentional seam for the RFC 9470 MFA step-up work
+ * (ERA-13405 REST, ERA-13733 websocket) and is left unregistered here.
```
luixlive: Uff another giant comment that provides ticket numbers and implementation details from the caller sites. Keeping this comment updated will be a challenge, better to remove it and simply add inline comments describing what the methods do.

src/utils/auth-recovery.js
```diff
@@ -0,0 +1,75 @@
...
+
+// The one shared in-flight recovery promise (null when idle).
+let inFlight = null;
+
+export const registerAuthRecovery = (next) => {
+  primitives = { ...primitives, ...next };
+};
+
+export const recoverAuth = ({ stepUp = false, challenge = null } = {}) => {
+  // TODO(ERA-13733): single-flight is keyed only on `inFlight`. When step-up is wired,
```
luixlive: This one is a TODO, not sure if worth it but if you'd rather keep it, let's make sure we eventually remove it.

luixlive (commented):
Very well tested! But I see many giant blocks of comments with several implementation details from other files and pointing to ticket numbers. As a dev it's not the best experience to have to open a ticket to understand a line of code. Better to add concise, descriptive inline comments 😄 I saw several of those, didn't signal all of them in the PR. I'm pretty sure an agent can rapidly spot and improve them.

src/hooks/useAuthRecovery.js
StephenWithPH: Trimmed to your suggested wording — fixed in a994ffddf0744868e5d52caf5ec037cb0d21b324. Dropped the ticket refs and the "mount once, high in the tree" line. Thanks!

src/utils/auth-recovery.js
StephenWithPH: Trimmed — fixed in 16de022faf9ebe016c7ed92036eabc296af7a9bf. Cut the caller-by-caller narration and the ticket refs; kept a short header (single-flight + injected-not-imported) and moved the one useful bit — why `stepUp` is currently unused — to a one-line comment where it's actually read.

src/utils/auth-recovery.js
StephenWithPH: Dropped it (in 16de022faf9ebe016c7ed92036eabc296af7a9bf). The hazard it flagged — that the single-flight must not collapse a step-up into an in-flight _silent_ renewal once step-up is wired — is now recorded as a comment on ERA-13733, so it's tracked where the follow-up work will see it rather than lingering in the code.

src/RequestConfigManager/index.js
StephenWithPH: Good call, but I'm keeping it out of this PR — it rewrites the component's whole data-flow on an auth-sensitive change and would balloon the diff/review. Would you like us to file a follow-up ticket for the `connect`/`mapStateToProps` → `useSelector`/`useDispatch` migration? Thanks for the review! 🙏

luixlive (approved):

PR #1659, opened 2026-07-16
ERA-13405: RFC 9470 MFA step-up handler in the REST API client

StephenWithPH:
## What & why

[ERA-13405](https://allenai.atlassian.net/browse/ERA-13405)

Adds the RFC 9470 MFA step-up handler to ER Web's REST client, building on the shared `auth-recovery` unit from ERA-13685. A `401 WWW-Authenticate: Bearer error="insufficient_user_authentication"` now triggers a transparent Auth0 step-up (re-run PKCE with the challenge's `acr_values`/`max_age`) instead of an opaque auth failure.

## Commits

1. **parse RFC 9470 step-up challenges** — `isStepUpChallenge` / `parseAuthChallenge` in `utils/auth-recovery` (shared, so the socket handler in ERA-13733 reuses them).
2. **key auth-recovery single-flight by mode** — `inFlight` is `{ silent, stepUp }`, so a step-up can't collapse into an in-flight silent renewal (a same-ACR silent token never satisfies the MFA challenge).
3. **route step-up 401s in the interceptor** — `RequestConfigManager.handle401Errors` reads `WWW-Authenticate`; a step-up challenge routes to `recoverAuth({ stepUp: true, challenge })`, an ordinary 401 keeps the silent-renew path.
4. **register the interactive step-up primitive** — `useAuthRecovery` registers `stepUp` = `loginWithRedirect` with the challenge's `acr_values`/`max_age`, returnTo stashed; the redirect callback re-bootstraps with the MFA-bearing token.

## How it works (end to end)

Step-up 401 → interceptor parses `acr_values`/`max_age` → `recoverAuth({ stepUp })` → `loginWithRedirect` → user completes MFA → the redirect callback (`Auth0TokenManager`) re-bootstraps with the MFA token and returns the user to the original route.

https://github.com/user-attachments/assets/724ad217-fa0f-4065-85a1-5502d983739a

## Dependencies / status

- Builds on **ERA-13685** (shared `auth-recovery` unit) — merged.
- Server emission **ERA-13387** (`WWW-Authenticate` step-up on 401) is **In Test** — end-to-end exercise needs a `require_mfa` env, hence the `deploy` label.
- **Socket** step-up (**ERA-13733**) is deferred, blocked on server **ERA-13732**; this PR is REST-only.

## Behavior notes — worth discussing

- **The step-up redirect discards unsaved in-flight state.** `loginWithRedirect` is a full-page navigation, so a half-entered report / patrol / notes are lost. This is benign at the default `mfa_max_age_seconds` (365 days) — step-up is then effectively a login-time event, before any data entry. But a tenant that sets a **short** `mfa_max_age` re-introduces *mid-session, mid-form* step-ups and real data loss. It's not a regression (new with MFA) and is strictly better than the opaque sign-out it replaces (which would also lose the form, then drop you at login). Escape hatch if it bites: the **popup** variant (`getAccessTokenWithPopup`, no reload) — the design is seam'd so only the `stepUp` primitive changes.
- **Silent ↔ step-up interaction.**
  - **(A) handled here:** the per-request `retriedAfterRefresh` marker is bypassed for step-up (`stepUp || !retriedAfterRefresh`), so a token that is expired *and* MFA-stale (silent-renews, then the replay 401s with a step-up challenge) escalates to step-up instead of signing out. Step-up redirects rather than looping, so bypassing the marker is safe.
  - **(B) deferred to ERA-13732/13733:** until the socket can *see* the step-up challenge, on a `require_mfa` site it treats the MFA rejection as a generic 401 → silent-renew → sign-out, racing the REST redirect (likely benign — the hard redirect wins — but messy). The mode-keyed single-flight is already set up so that once the socket recognizes step-up, both transports coalesce into a single redirect.
  - Both only bite once `require_mfa` is enabled; **(A)** specifically needs token-expiry and MFA-staleness to coincide (short `mfa_max_age`; never at the 365-day default).

## Testing

- Unit: challenge parsing (incl. an `error_description` decoy), mode-keyed single-flight (both directions + concurrent step-up coalescing), interceptor silent/step-up routing + the escalation case, and the `stepUp` primitive (reject-surfaces + never-settle).
- Full suite green.

## Deploy

`deploy` label added to spin up a `require_mfa`-capable feature env for manual end-to-end verification (the unit tests mock at the seams).

[ERA-13405]: https://allenai.atlassian.net/browse/ERA-13405?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/hooks/useAuthRecovery.js
+++ b/src/hooks/useAuthRecovery.js
@@ -1,18 +1,40 @@
 import { useEffect } from 'react';
 import { useAuth0 } from '@auth0/auth0-react';
+import { useSelector } from 'react-redux';
+import { useLocation } from 'react-router';
 
+import appConfig from '../config';
 import { registerAuthRecovery } from '../utils/auth-recovery';
+import { buildAuth0AuthorizationParams } from '../utils/auth0';
+import { setIntendedPostAuth0SuccessRoute } from '../utils/auth';
 
 /**
  * Registers the live @auth0/auth0-react primitives into the shared auth-recovery unit
  * so the axios interceptor and the websocket handler can both drive token renewal.
  */
 const useAuthRecovery = () => {
-  const { getAccessTokenSilently } = useAuth0();
+  const { getAccessTokenSilently, loginWithRedirect } = useAuth0();
+  const { pathname, search } = useLocation();
+  const idpOrgId = useSelector((state) => state.view.systemConfig?.idp_org_id);
 
   useEffect(() => {
-    registerAuthRecovery({ silentRenew: () => getAccessTokenSilently() });
-  }, [getAccessTokenSilently]);
+    registerAuthRecovery({
+      silentRenew: () => getAccessTokenSilently(),
+      // Interactive MFA step-up: re-run PKCE with the challenge's acr_values/max_age. The
+      // redirect navigates away, so return a never-settling promise (no premature replay).
+      stepUp: async ({ acrValues, maxAge } = {}) => {
+        setIntendedPostAuth0SuccessRoute(`${pathname}${search}`);
+        await loginWithRedirect({
+          authorizationParams: {
+            ...buildAuth0AuthorizationParams(appConfig.auth0.audience, idpOrgId),
+            ...(acrValues ? { acr_values: acrValues } : {}),
+            ...(maxAge ? { max_age: maxAge } : {}),
+          },
+        });
+        return new Promise(() => {});
+      },
+    });
+  }, [getAccessTokenSilently, loginWithRedirect, idpOrgId, pathname, search]);
 };
 
 export default useAuthRecovery;
--- a/src/utils/auth-recovery.js
+++ b/src/utils/auth-recovery.js
@@ -2,10 +2,12 @@ import store from '../store';
 import { applyAccessToken } from '../ducks/auth';
 
 /**
- * Shared, single-flight auth recovery. Concurrent 401s collapse into one in-flight
- * token renewal that every caller awaits, and the renewed token is applied once.
- * Auth0 primitives are injected via `registerAuthRecovery` (they are React-hook-bound),
- * not imported.
+ * Shared, single-flight auth recovery, keyed by mode (silent renewal vs interactive
+ * step-up). Concurrent same-mode 401s collapse into one in-flight renewal that every
+ * caller awaits, and its token is applied once. The two modes run independently, so a
+ * step-up never collapses into an in-flight silent renewal (a same-ACR silent token can't
+ * satisfy an MFA challenge). Auth0 primitives are injected via `registerAuthRecovery`
+ * (they are React-hook-bound), not imported.
  */
 
 const STEP_UP_ERROR = 'insufficient_user_authentication';
@@ -44,17 +46,20 @@ const withTimeout = async (promise, ms) => {
 
 let primitives = { silentRenew: null, stepUp: null };
 
-// The one shared in-flight recovery promise (null when idle).
-let inFlight = null;
+// In-flight recovery, keyed by mode: a step-up must never collapse into an in-flight
+// silent renewal (a silent token is same-ACR and can't satisfy the MFA challenge).
+let inFlight = { silent: null, stepUp: null };
 
 export const registerAuthRecovery = (next) => {
   primitives = { ...primitives, ...next };
 };
 
 export const recoverAuth = ({ stepUp = false, challenge = null } = {}) => {
-  if (!inFlight) {
-    inFlight = (async () => {
-      // Only silent renewal is wired; the stepUp branch is a reserved seam for MFA step-up.
+  const mode = stepUp ? 'stepUp' : 'silent';
+  if (!inFlight[mode]) {
+    // Same-mode callers coalesce onto this flight, so the first caller's challenge is used
+    // (challenges are identical for a given requirement).
+    inFlight[mode] = (async () => {
       const recover = stepUp ? primitives.stepUp : primitives.silentRenew;
       if (typeof recover !== 'function') {
         throw new Error(`auth-recovery: no ${stepUp ? 'stepUp' : 'silentRenew'} primitive registered`);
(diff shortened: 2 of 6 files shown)


src/hooks/useAuthRecovery.js
```diff
@@ -1,18 +1,40 @@
...
+      // redirect navigates away, so return a never-settling promise (no premature replay).
+      stepUp: async ({ acrValues, maxAge } = {}) => {
+        setIntendedPostAuth0SuccessRoute(`${pathname}${search}`);
+        await loginWithRedirect({
+          authorizationParams: {
+            ...buildAuth0AuthorizationParams(appConfig.auth0.audience, idpOrgId),
+            acr_values: acrValues,
+            max_age: maxAge,
+          },
+        });
```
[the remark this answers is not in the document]
StephenWithPH: Fixed in 8488694cbee8d57ee54c02ad2b574bd7db849212 — `acr_values` and `max_age` are now spread into `authorizationParams` only when the challenge provides them (`...(acrValues ? { acr_values: acrValues } : {})`), so neither is ever passed through as `undefined`.

chrisj-er (approved):
Nothing blocking. Highlights:

- Step-up latch has no watchdog (minor). The stepUp primitive returns new Promise(() => {}) that never settles, so recoverAuth's .finally never releases inFlight.stepUp. Correct in the normal flow (page redirects and unloads), but if loginWithRedirect ever resolves without navigating, the latch stays stuck forever and every later step-up 401 hangs with no sign-out fallback. Low-probability, but it's a silent hang rather than a recoverable error.

StephenWithPH:
@chrisj-er good catch on the step-up latch watchdog. It's addressed in the stacked follow-up PR #1663 (ERA-13733, based on this branch), commit 1df48b31: the step-up flight is now time-boxed by a 60s redirect-handoff watchdog, so if `loginWithRedirect` ever resolves without navigating, `recoverAuth` rejects (→ sign-out) and releases `inFlight.stepUp` rather than hanging forever. It lives on ERA-13733 rather than here because that branch adds the websocket consumer that makes the hang exposure real, and the fix covers both transports.

src/hooks/useAuthRecovery.js
```diff
@@ -1,18 +1,40 @@
...
+      stepUp: async ({ acrValues, maxAge } = {}) => {
+        setIntendedPostAuth0SuccessRoute(`${pathname}${search}`);
+        await loginWithRedirect({
+          authorizationParams: {
+            ...buildAuth0AuthorizationParams(appConfig.auth0.audience, idpOrgId),
+            ...(acrValues ? { acr_values: acrValues } : {}),
+            ...(maxAge ? { max_age: maxAge } : {}),
+          },
+        });
+        return new Promise(() => {});
```
luixlive, line 34: From a Claude review:

**Step-up recovery can hang forever if loginWithRedirect resolves without navigating.**

`recoverAuth`'s cleanup only clears `inFlight.stepUp` via `.finally()`, which never fires if that promise never settles. In the normal case the page navigates away before this matters, but if `loginWithRedirect` ever resolves without an actual navigation, `inFlight.stepUp` stays stuck forever and every subsequent step-up 401 for the rest of the session silently hangs with no fallback to sign-out.

src/hooks/useAuthRecovery.js
```diff
@@ -1,18 +1,40 @@
...
+import { setIntendedPostAuth0SuccessRoute } from '../utils/auth';
 
 /**
  * Registers the live @auth0/auth0-react primitives into the shared auth-recovery unit
  * so the axios interceptor and the websocket handler can both drive token renewal.
  */
 const useAuthRecovery = () => {
-  const { getAccessTokenSilently } = useAuth0();
+  const { getAccessTokenSilently, loginWithRedirect } = useAuth0();
+  const idpOrgId = useSelector((state) => state.view.systemConfig?.idp_org_id);
```
luixlive: Extremely nit: we usually organize the component structure by firs having third party hooks, then state hooks (selectors, useState):
```Javascript
const useAuthRecovery = () => {
  const { getAccessTokenSilently, loginWithRedirect } = useAuth0();
  const { pathname, search } = useLocation();

  const idpOrgId = useSelector((state) => state.view.systemConfig?.idp_org_id);
```

src/utils/auth-recovery.js
```diff
@@ -8,6 +8,23 @@ import { applyAccessToken } from '../ducks/auth';
  * not imported.
  */
```
luixlive, line 11: **Stale top-of-file doc comment**

This PR's own change replaced the single global `inFlight` promise with a mode-keyed map (`{ silent, stepUp }`), so a silent renewal and a step-up can now run concurrently, each independently dispatching `applyAccessToken`. The header comment wasn't updated to reflect the new per-mode invariant and is now misleading.

luixlive (approved):
A couple items from a Claude review, and a nit from my side. Honestly, none of them are blocker so I'll approve. Thanks for the thorough testing!

src/hooks/useAuthRecovery.js
StephenWithPH: Yep... will get in fast-follow: https://github.com/PADAS/das-web-react/pull/1659#issuecomment-5007267601

src/hooks/useAuthRecovery.js
StephenWithPH: Reordered in 33b3685f — third-party hooks first (`useAuth0`, then `useLocation`), then the `useSelector` state hook. Thanks!

src/utils/auth-recovery.js
StephenWithPH: Good catch — updated in 5bd4cdbe. The header now describes the mode-keyed invariant: concurrent *same-mode* 401s collapse into one in-flight renewal (applied once), and the silent and step-up modes run independently — so it no longer implies a single global flight.

PR #1663, opened 2026-07-17
ERA-13733: RFC 9470 MFA step-up over the rt_api websocket

StephenWithPH:
## What & why

[ERA-13733](https://allenai.atlassian.net/browse/ERA-13733)

Handles the RFC 9470 MFA step-up challenge over the `rt_api` websocket, the socket sibling of the REST handler in #1659 (ERA-13405). On a `require_mfa` site the server now surfaces the challenge on a `resp_authorization` 401 as `status.www_authenticate` (a raw `WWW-Authenticate` string — das#4085 / ERA-13732). This routes it through the shared `auth-recovery` unit's interactive step-up instead of a silent-renew (which is same-ACR and can't satisfy MFA) followed by a blind sign-out.

Also hardens the shared unit with a step-up **redirect-handoff watchdog** (raised in review on #1659).

> **Stacked on #1659.** Base is `ERA-13405`, so this diff is just the two ERA-13733 commits. GitHub will auto-retarget the base to `develop` when #1659 merges; this PR must merge **after** #1659.

## Commits

1. **route step-up 401s over the websocket** — `useRealTimeImplementation`'s `resp_authorization` handler reads `status.www_authenticate`; a step-up challenge routes to `recoverAuth({ stepUp: true, challenge })` (redirect — never silent-renews), an ordinary 401 keeps the ERA-13685 silent-renew path. Step-up bypasses the per-bind `authRetried` guard (`stepUp || !authRetried`) so a renewed-but-MFA-stale token escalates to step-up rather than signing out. Mirrors the REST interceptor.
2. **bound the step-up redirect handoff (watchdog)** — see below.

## How it works (end to end)

Socket step-up 401 → handler parses `status.www_authenticate` → `recoverAuth({ stepUp })` → `loginWithRedirect` (the interactive primitive owned by ERA-13405) → user completes MFA → the redirect callback re-bootstraps with the MFA token. The rebuilt socket's `connect` handler reads the fresh token from the store and re-authorizes — so no bespoke socket recovery is needed. Because the flight is mode-keyed single-flight, a concurrent REST + socket step-up coalesces into **one** redirect.

## The watchdog (addresses @chrisj-er's note on #1659)

The `stepUp` primitive resolves `loginWithRedirect` and then stays pending (`new Promise(() => {})`) so recovery doesn't replay before the page unloads — which means `recoverAuth`'s `inFlight.stepUp` latch relied on unload to clear. If `loginWithRedirect` ever resolved *without* navigating, the latch stuck forever and every later step-up 401 hung with no sign-out fallback.

Fix: time-box the step-up flight with a generous `STEP_UP_REDIRECT_TIMEOUT_MS = 60_000` via the existing `withTimeout` (silent renewal stays 30s). A successful redirect unloads the page in well under a second, so the bound can **only** fire when the redirect never navigated — converting an indefinite hang into a recoverable sign-out and releasing the latch. It never truncates the user's MFA, which happens on the Auth0 page *after* unload (not on this pending promise).

## Dependencies / status

- **Stacked on #1659 (ERA-13405)** — the REST step-up + the shared parsers, single-flight, and interactive `stepUp` primitive this reuses.
- Builds on **ERA-13685** (shared `auth-recovery` unit + socket silent-renew) — merged.
- Server socket emission **ERA-13732** (das#4085) is **In Pull Request**. The payload is a raw `WWW-Authenticate` string in `status.www_authenticate`, byte-compatible with the HTTP header so the REST parser (`isStepUpChallenge`/`parseAuthChallenge`) is reused unchanged.
- `require_mfa` (**ERA-13387**) is **In Test** — end-to-end exercise needs a `require_mfa` env, hence the `deploy` label.

## Behavior notes — worth discussing

- **AC3 (concurrent REST + socket → a single step-up)** falls out of ERA-13405's mode-keyed single-flight; the socket handler joins the same `inFlight.stepUp` flight. Not re-asserted at the socket layer here (`recoverAuth` is mocked in the socket test); it's pinned in `auth-recovery.test.js`.
- **Deferred hardening (core-review finding).** Step-up loop-prevention is currently *implicit*: both transports bypass their anti-loop guards for step-up, safe only because the redirect primitive never resolves with a token (it navigates away, or the watchdog rejects → sign-out). This is fine for the redirect variant shipping here. **If the popup escape-hatch variant is ever adopted** (it resolves in-page with a token), an explicit "step-up attempted once per cycle" marker is required, or a repeated step-up 401 would loop. Tied to the popup variant, deliberately not in this PR.
- **The redirect discards unsaved in-flight state** — carries over verbatim from #1659 (same `stepUp` primitive). Benign at the default `mfa_max_age_seconds` (365 days); a short per-tenant window re-introduces mid-form step-ups. Escape hatch is the popup variant (see above).

## Testing

- **Socket** (`withSocketConnection/index.test.js`): step-up challenge routes to the interactive path (exact parsed `{ stepUp, challenge }` args, no silent-renew, no sign-out); an ordinary (non-step-up) `www_authenticate` still silent-renews; and the guard-bypass escalation (a silently-renewed socket then 401s with a step-up challenge → step-up, not sign-out).
- **Shared unit** (`utils/auth-recovery.test.js`): the watchdog rejects a step-up whose redirect never navigates, does not dispatch a token, and releases `inFlight.stepUp` so a later step-up retries.
- Full suite green (2891).

## Deploy

`deploy` label added to spin up a `require_mfa`-capable feature env for manual end-to-end verification (the unit tests mock at the seams).

[ERA-13733]: https://allenai.atlassian.net/browse/ERA-13733?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/utils/auth-recovery.js
+++ b/src/utils/auth-recovery.js
@@ -27,15 +27,15 @@ export const parseAuthChallenge = (challenge) => {
 export const isStepUpChallenge = (challenge) =>
   parseAuthChallenge(challenge)?.error === STEP_UP_ERROR;
 
-// A stalled silent renewal (e.g. a network black-hole) must not hang recovery for both
-// transports, so it is time-boxed. Interactive step-up is deliberately NOT bounded — it
-// waits on the user completing MFA, which can take far longer than any network timeout.
+// Time-boxed so a stalled recovery can't hang forever. Step-up's longer bound only trips
+// if the redirect never navigated (a real redirect unloads the page first).
 const SILENT_RENEW_TIMEOUT_MS = 30_000;
+const STEP_UP_REDIRECT_TIMEOUT_MS = 60_000;
 
-const withTimeout = async (promise, ms) => {
+const withTimeout = async (promise, ms, label = 'recovery') => {
   let timeoutId;
   const timeout = new Promise((_, reject) => {
-    timeoutId = setTimeout(() => reject(new Error('auth-recovery: renewal timed out')), ms);
+    timeoutId = setTimeout(() => reject(new Error(`auth-recovery: ${label} timed out`)), ms);
   });
   try {
     return await Promise.race([promise, timeout]);
@@ -65,9 +65,11 @@ export const recoverAuth = ({ stepUp = false, challenge = null } = {}) => {
         throw new Error(`auth-recovery: no ${stepUp ? 'stepUp' : 'silentRenew'} primitive registered`);
       }
 
-      const accessToken = stepUp
-        ? await recover(challenge)
-        : await withTimeout(recover(challenge), SILENT_RENEW_TIMEOUT_MS);
+      const accessToken = await withTimeout(
+        recover(challenge),
+        stepUp ? STEP_UP_REDIRECT_TIMEOUT_MS : SILENT_RENEW_TIMEOUT_MS,
+        stepUp ? 'step-up redirect' : 'silent renewal',
+      );
       if (!accessToken) {
         throw new Error('auth-recovery: renewal returned no token');
       }
--- a/src/withSocketConnection/useRealTimeImplementation/index.js
+++ b/src/withSocketConnection/useRealTimeImplementation/index.js
@@ -4,7 +4,7 @@ import { DAS_HOST } from '../../constants';
 import { resetSocketStateTracking } from './helpers';
 import { SOCKET_HEALTHY_STATUS } from '../../ducks/system-status';
 import { clearAuth } from '../../ducks/auth';
-import { recoverAuth } from '../../utils/auth-recovery';
+import { isStepUpChallenge, parseAuthChallenge, recoverAuth } from '../../utils/auth-recovery';
 import { events } from './config';
 import { calcEventFilterForRequest } from '../../utils/event-filter';
 import { calcPatrolFilterForRequest } from '../../utils/patrol-filter';
@@ -73,10 +73,13 @@ const useRealTimeImplementation = () => {
     socket.on('resp_authorization', async (msg) => {
       const { status } = msg;
       if (status.code === 401) {
-        if (!authRetried) {
+        // Step-up bypasses the once-per-cycle guard: it redirects rather than looping.
+        const challenge = status.www_authenticate;
+        const stepUp = isStepUpChallenge(challenge);
+        if (stepUp || !authRetried) {
           authRetried = true;
           try {
-            await recoverAuth();
+            await recoverAuth(stepUp ? { stepUp: true, challenge: parseAuthChallenge(challenge) } : undefined);
             console.log('realtime: token renewed; re-authorizing');
             return authorize();
           } catch (renewalError) {
--- a/src/utils/auth-recovery.test.js
+++ b/src/utils/auth-recovery.test.js
@@ -98,7 +98,7 @@ describe('auth-recovery', () => {
       const silentRenew = jest.fn(() => new Promise(() => {}));
       registerAuthRecovery({ silentRenew });
 
-      const rejection = expect(recoverAuth()).rejects.toThrow(/timed out/);
+      const rejection = expect(recoverAuth()).rejects.toThrow(/silent renewal timed out/);
       await jest.advanceTimersByTimeAsync(30_000);
       await rejection;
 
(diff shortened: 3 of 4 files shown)


src/utils/auth-recovery.js
```diff
@@ -63,9 +68,10 @@ export const recoverAuth = ({ stepUp = false, challenge = null } = {}) => {
         throw new Error(`auth-recovery: no ${stepUp ? 'stepUp' : 'silentRenew'} primitive registered`);
       }
 
-      const accessToken = stepUp
-        ? await recover(challenge)
-        : await withTimeout(recover(challenge), SILENT_RENEW_TIMEOUT_MS);
+      const accessToken = await withTimeout(
+        recover(challenge),
+        stepUp ? STEP_UP_REDIRECT_TIMEOUT_MS : SILENT_RENEW_TIMEOUT_MS,
+      );
```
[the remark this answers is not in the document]
StephenWithPH: Good call. `withTimeout` now takes a `label`, and the two flights pass distinct ones, so the rejection reads `auth-recovery: step-up redirect timed out` vs `auth-recovery: silent renewal timed out` — the two failure modes (a redirect that never navigated vs a silent-renew network stall) are now distinguishable in the sign-out logs. Also tightened the two timeout tests to assert the specific labels. Folded into the redirect-watchdog commit.

src/utils/auth-recovery.js
```diff
@@ -27,15 +27,20 @@ export const parseAuthChallenge = (challenge) => {
...
-// A stalled silent renewal (e.g. a network black-hole) must not hang recovery for both
-// transports, so it is time-boxed. Interactive step-up is deliberately NOT bounded — it
-// waits on the user completing MFA, which can take far longer than any network timeout.
+// A stalled recovery must not hang recovery for both transports, so each flight is
+// time-boxed. Silent renewal guards a network black-hole. Step-up guards the redirect
+// handoff: the interactive primitive resolves loginWithRedirect and then stays pending
+// until the page unloads — a successful redirect unloads in well under a second, so this
+// bound only fires when the redirect never navigated, converting an indefinite hang into a
+// recoverable sign-out. It never truncates the user's MFA, which happens on the Auth0 page
+// after unload (not on this pending promise), so the step-up bound is generous.
```
luixlive: Ok this comment is definitely too big haha
```Javascript
// Each flight is time-boxed so a stalled recovery can't hang forever. Silent renewal
// guards a network black-hole. Step-up guards the redirect handoff.
```

I feel each Claude model comes trained to add more and more verbose comments 🤣 We'll end up having more commentary than code.

src/withSocketConnection/useRealTimeImplementation/index.js
```diff
@@ -73,10 +73,16 @@ const useRealTimeImplementation = () => {
...
+        // Step-up bypasses the once-per-cycle guard: a renewed-but-MFA-stale token surfaces
+        // the challenge only on the re-authorize, and step-up redirects rather than looping.
+        const challenge = status.www_authenticate;
+        const stepUp = isStepUpChallenge(challenge);
+        if (stepUp || !authRetried) {
           authRetried = true;
           try {
-            await recoverAuth();
+            // Silent-renew is same-ACR and can't satisfy MFA, so a step-up 401 must route to
+            // the interactive path; it redirects and never resolves (the page navigates away).
```
luixlive: Two separate comments explain "step-up bypasses the guard because it redirects instead of looping." We could merge them into:
```Javascript
// Step-up bypasses the once-per-cycle guard: it redirects rather than looping, and a
// renewed-but-MFA-stale token surfaces the challenge only on the re-authorize.
```

luixlive (approved):
As I have mentioned in other PRs, we are getting big verbose blocks of comments. This PR again has a few of them. The code itself looks right, I'd just clean some of the comments 😅

src/utils/auth-recovery.js
StephenWithPH: Trimmed to 2 lines in 889d4970 — kept only the redirect-unload rationale. Also did a broader terseness pass over the test comments while here. 😄

src/withSocketConnection/useRealTimeImplementation/index.js
StephenWithPH: Merged to one line in b0c96ece; dropped the same-ACR note since it is already stated in auth-recovery.js.

PR #1666, opened 2026-07-20
ERA-13723: Blue start/end patrol pins with patrol-name labels

tiffanynwong:
## Summary
Updates the patrol **start** and **end** map markers:

- **Shared blue pins** replace the per-patrol-type icon at the track endpoints — a play glyph (▶) for start and a square (■) for end, in the design blue (`#0056C7`) with a white border.
- **Labels show the patrol name** instead of "Patrol Start" / "Patrol End": patrol title → patrol type display name → the generic "Patrol Start" / "Patrol End" fallback (the `(Est)` suffix for estimated points is preserved).
- The name renders **to the right of the pin**, vertically centered, as black text with a white halo (no background chip).
- Pins render **32px tall at full zoom** and scale down when zoomed out, matching the other map icons.

## New assets
- `src/common/images/icons/patrol-start-pin.svg`
- `src/common/images/icons/patrol-end-pin.svg`

## Notes for reviewers
- **`ClustersLayer/utils.js` (out of scope):** removes a pre-existing duplicate `import` from `../constants` that was preventing the app from compiling on this branch. Included because it blocks the build; happy to split it out if preferred.
- **Reliability fix:** the start/stop symbol layer is added by a memoized child whose add-layer effect runs before the parent creates the map source. When patrol data is already loaded at mount, the source didn't exist yet and the pins silently failed to attach. The layer now waits for the source (via a `sourcedata` event) before mounting the symbol child.

## Testing
- `yarn test` — `src/utils/patrols.test.js` passes.
- `yarn lint` — clean on changed files.
- Verified locally against a dev tenant: start/end pins render with patrol-name labels; confirmed the layer attaches reliably and the icon-size expression is accepted by Mapbox.

## Follow-ups / open questions
- Long patrol names wrap to two lines (existing Mapbox label behavior); not truncated with an ellipsis like the mock. Can cap to one line if desired.
- Dropped the pin drop-shadow so the pin fills its icon box and matches other icon heights — can re-add a subtle shadow if wanted.

start and stop pins


src/PatrolStartStopLayer/layer.js
```diff
@@ -31,16 +31,48 @@ const lineLayout = {
   'line-cap': 'round',
 };
 
+// The patrol name renders on the icon layer (to the right of the pin) as black
```
luixlive, line 34: I see too much commentary. We've asked the agents to avoid unnecessary blocks of comments when doing changes, but I think this PR was sent before I updated the AGENTS.md 😄

src/utils/patrols.js
```diff
@@ -571,22 +584,26 @@ export const extractPatrolPointsFromTrackData = ({ leader, patrol, trackData },
...
-      patrol_points.end_location = makePatrolPointFromFeature(`Patrol End${lastTrackPointMatchesEndTime ? '' : ' (Est)'}`, [longitude, latitude], icon_id, stroke, lastTrackPoint.properties.time);
+      patrol_points.end_location = makePatrolPointFromFeature(`${endLabel}${lastTrackPointMatchesEndTime ? '' : ' (Est)'}`, [longitude, latitude], patrolEndPin, stroke, lastTrackPoint.properties.time);
     }
   }
 
   if (!!patrol_points.start_location && !patrol_points.end_location &&
   isPatrolDone) {
     patrol_points.end_location = cloneDeep(patrol_points.start_location);
-    patrol_points.end_location.properties.title = 'Patrol End (Est)';
+    patrol_points.end_location.properties.image = patrolEndPin;
```
luixlive, line 594: Claude:
The new end-pin assignment in the done-patrol clone branch is unconditionally dead code — the cloned `end_location` is always deleted three lines later, so this case never gets the square end glyph the PR is about.

src/utils/patrols.js
```diff
@@ -571,22 +584,26 @@ export const extractPatrolPointsFromTrackData = ({ leader, patrol, trackData },
...
   if (!!patrol_points.end_location && !!patrol_points.start_location
     && booleanEqual(
       point(patrol_points.end_location.geometry.coordinates),
       point(patrol_points.start_location.geometry.coordinates),
     )) {
-    patrol_points.start_location.properties.title += ` & ${patrol_points.end_location.properties.title}`;
+    if (patrol_points.start_location.properties.title !== patrol_points.end_location.properties.title) {
+      patrol_points.start_location.properties.title += ` & ${patrol_points.end_location.properties.title}`;
+    }
     delete patrol_points.end_location;
```
luixlive, line 607: Claude:
When start and end coincide, `delete patrol_points.end_location` keeps `patrolStartPin`, so the new end glyph is silently dropped and — with the new equality guard — the label no longer says the patrol ended there either.

src/utils/patrols.js
```diff
@@ -521,6 +523,17 @@ export const extractPatrolPointsFromTrackData = ({ leader, patrol, trackData },
     || (!!leader && !!leader.additional && !!leader.additional.rgb && `rgb(${leader.additional.rgb})`)
     || DEFAULT_STROKE;
 
+  // Label the start/end markers with the patrol's name (title, else patrol type
+  // display name), falling back to the generic "Patrol Start"/"Patrol End".
+  const patrolType = firstLeg?.patrol_type;
+  const { data: { patrolTypes } } = store.getState();
```
luixlive, line 529: Claude:
`patrolTypes` is read via `store.getState()` from inside a reselect compute function, so it is not a memoization input and stale generic labels are never recomputed.

src/utils/patrols.js
```diff
@@ -571,22 +584,26 @@ export const extractPatrolPointsFromTrackData = ({ leader, patrol, trackData },
...
+    patrol_points.end_location.properties.title = `${endLabel} (Est)`;
   }
 
   if (!!patrol_points.end_location && !!patrol_points.start_location
     && booleanEqual(
       point(patrol_points.end_location.geometry.coordinates),
       point(patrol_points.start_location.geometry.coordinates),
     )) {
-    patrol_points.start_location.properties.title += ` & ${patrol_points.end_location.properties.title}`;
+    if (patrol_points.start_location.properties.title !== patrol_points.end_location.properties.title) {
```
luixlive, line 604: Claude:
The new `title !== title` guard compares titles after the ' (Est)' suffix is appended, so a coincident start/end renders the patrol name twice in one label.

src/utils/patrols.js
```diff
@@ -521,6 +523,17 @@ export const extractPatrolPointsFromTrackData = ({ leader, patrol, trackData },
...
 
+  // Label the start/end markers with the patrol's name (title, else patrol type
+  // display name), falling back to the generic "Patrol Start"/"Patrol End".
+  const patrolType = firstLeg?.patrol_type;
+  const { data: { patrolTypes } } = store.getState();
+  const matchingPatrolType = patrolType
+    ? (patrolTypes || []).find(({ value, id }) => value === patrolType || id === patrolType)
+    : null;
+  const patrolName = patrol.title || matchingPatrolType?.display || null;
+  const startLabel = patrolName || 'Patrol Start';
```
luixlive, line 534: Claude:
'Patrol Start', 'Patrol End' and ' (Est)' are left hardcoded in English across six locales, violating the repo's i18n rule, in a file that already uses i18next for exactly this fallback.

src/utils/patrols.js
```diff
@@ -521,6 +523,17 @@ export const extractPatrolPointsFromTrackData = ({ leader, patrol, trackData },
     || (!!leader && !!leader.additional && !!leader.additional.rgb && `rgb(${leader.additional.rgb})`)
     || DEFAULT_STROKE;
 
+  // Label the start/end markers with the patrol's name (title, else patrol type
+  // display name), falling back to the generic "Patrol Start"/"Patrol End".
+  const patrolType = firstLeg?.patrol_type;
```
luixlive, line 528: Claude:
The 8-line patrol-name derivation re-implements `displayTitleForPatrol` from the same file but drops its leader-name fallback, so the map pin and the track legend show different names for the same patrol.

src/utils/patrols.js
```diff
@@ -530,20 +543,20 @@ export const extractPatrolPointsFromTrackData = ({ leader, patrol, trackData },
   const startTime = new Date(start_time);
 
   if (start_location) {
-    patrol_points.start_location = makePatrolPointFromFeature('Patrol Start', [start_location.longitude, start_location.latitude], icon_id, stroke, start_time);
+    patrol_points.start_location = makePatrolPointFromFeature(startLabel, [start_location.longitude, start_location.latitude], patrolStartPin, stroke, start_time);
```
luixlive, line 546: Claude:
Replacing the per-patrol-type sprite with one shared blue pin, plus dropping the per-patrol label color, leaves every patrol's endpoints visually identical and desynchronizes the map markers from the track legend.

src/PatrolStartStopLayer/layer.js
```diff
@@ -31,16 +31,48 @@ const lineLayout = {
...
+  0, (0.2 / MAP_ICON_SCALE) * PIN_SIZE_SCALE,
+  14, (1 / MAP_ICON_SCALE) * PIN_SIZE_SCALE,
+];
+
+// Pin tip sits on the patrol location; the patrol name sits to the right of the
+// pin, vertically centered on the pin head.
+const iconLayout = {
+  'icon-anchor': 'bottom',
+  'icon-size': PIN_ICON_SIZE,
   'text-field': '{title}',
```
luixlive, line 72: Claude:
 Moving `'text-field': '{title}'` onto the icon layer bypasses `mapUserLayoutConfigByLayerId`, so Settings → Map → Map Markers → Patrols no longer hides patrol start/end names.

src/PatrolStartStopLayer/layer.js
```diff
@@ -99,11 +127,33 @@ const StartStopLayer = ({ patrolData, ...rest }) => {
     layout: lineLayout
   }]);
 
+  // The symbol layer is added by a memoized child whose add-layer effect runs
+  // before this component's source-creation effect. When the patrol data is
+  // already loaded at mount, the source doesn't exist yet on that first pass and
+  // the memoized child never re-renders to retry. Wait for the source to exist
+  // (signalled by a sourcedata event) before mounting the child so its layer
+  // reliably attaches.
+  const [isSourceReady, setIsSourceReady] = useState(false);
```
luixlive, line 136: Claude:
`isSourceReady` is a one-way latch fed by a global `sourcedata` event with no synchronous check and no re-arm, so it both delays and can permanently prevent the symbol layer from attaching.

src/PatrolStartStopLayer/layer.js
```diff
@@ -56,20 +88,16 @@ const StartStopLayer = ({ patrolData, ...rest }) => {
...
-      }
-
+    const pointFeatures = patrolData?.startStopGeometries?.points;
+
+    if (pointFeatures) {
+      Object.values(pointFeatures)
+        .filter(Boolean)
+        .forEach(({ properties: { image } = {} }) => {
+          if (image && !map.hasImage(calcImgIdFromUrlForMapImages(image))) {
+            addMapImage({ src: image });
```
luixlive, line 98: Claude:
Two static app-bundled pins are registered from a per-patrol effect keyed on a churning `patrolData`, with an un-awaited `addMapImage` and no in-flight guard, instead of once at map init.

luixlive (changes requested):
Thanks Tiffany 😄  Nice PR, sorry it took me that much to review it. I have some feedback for it, but most of it can be delegated to Claude. Just ask it to read my comments from the PR and fix them.

Three high level comments:
1- Tests are broken after these changes 🥲  You can ask Claude to see what's going on and fix them. Then you can make sure they pass with `yarn test`.
2- This branch is a bit old. Let's make sure to update develop and merge it here (you can ask Claude to do all this). And there are a couple conflicts we need to fix.
3- We are missing unit tests for the changes.

The rest are trivial comments, a couple from me, a bunch from a local Claude review. Most seem trivial to fix!

Edit: btw, I also see a couple items from Copilot that we should work on.

tiffanynwong:
Thanks @luixlive for the review 🙏 Pushed updates addressing the feedback — here's the rundown.

### Branch + tests
- Merged latest `develop` and reconciled with the patrol-selector refactor (`selectPatrolTrackData`, the `extractLegPatrolPoints` / `finalizeCombinedPatrolPoints` split, multi-segment). Conflicts resolved — the PR is mergeable now.
- `yarn test` is green (4315 passing). Updated the existing patrol tests for the new behavior and added unit tests for `getPatrolPointLabels`, `extractLegPatrolPoints`, and `finalizeCombinedPatrolPoints`.

### Copilot
- **Memoization:** `patrolTypes` is now a reselect input to `selectPatrolTrackData` / `selectPatrolsWithTracksData` and threaded through `buildPatrolData`, so labels recompute when patrol types finish loading (no more `store.getState()` in the compute path).
- **`isSourceReady`:** removed the gate.
- **Label unit tests:** added.

### Review comments
- **Patrol name fallback:** now resolves the name the same way as `displayTitleForPatrol` (title → leader name → patrol type display) before the generic "Patrol Start/End", so the pins match the track legend.
- **Dead clone code + coincident start/end double-name:** fixed.
- **i18n:** "Patrol Start" / "Patrol End" / "(Est)" moved into i18next across all locales. Translations for es/fr/pt/sw/ne-NP are best-effort — a native-speaker check would be appreciated.
- **Map Markers hide-names toggle:** the name stays on the label layer (so the toggle still applies) but is positioned to the right of the pin.
- **Image registration:** the two static pins are registered once instead of from a per-patrol effect.
- **Comment blocks:** trimmed.

### Kept intentionally
Per the ticket, all patrols share the same blue start/end pin, so the endpoints no longer mirror the per-type legend icon — flagging in case that's a concern.

Ready for another look!

PR #1671, opened 2026-07-21
ERA-13761: Reduce the size of the community download

jeslefcourt:
## ERA-13761 — Reduce the size of the community download

https://allenai.atlassian.net/browse/ERA-13761

The community input route (`/community/:value`) renders the full `ReportManager`, so its static import graph pulled in the app's heaviest map/geo vendors even though a community submitter never sees a map. This PR removes the two biggest offenders from that download.

### Changes

1. **Code-split event location markers** — `SchemaForm` no longer statically imports `useMapLocationMarkers`. The marker machinery moved into a lazily-loaded `EventLocationMarkers` component that mounts only when a `MapContext` map is present; a ref bridge + readiness flag preserves the initial marker push without changing the hook's public API or its tests.
2. **Drop `mapbox-gl` from `useJumpToLocation`** — it only used `LngLatBounds` to build a bounds envelope, but `map.fitBounds` accepts a plain `[[w,s],[e,n]]` array. It was the *sole* mapbox-gl importer in the community graph, so this evicts the entire library.
3. **Lazy-load `@turf/buffer`** in `utils/query` (`getBboxParamsFromMap`) and `ducks/analyzers` — both already-async functions. `buffer` pulls in `@turf/jsts` (~250 KB min), which was landing in the combined-reducer store chunk. Importing it dynamically moves jsts to an on-demand chunk. `@turf/buffer` is now declared as an explicit dependency (it was a hoisted transitive of `@turf/turf`).

### Measured impact (before/after production builds)

| | Community route raw JS |
|---|---|
| Before | ~4,165 kB |
| After | ~2,118 kB (**~49% smaller**) |

- mapbox-gl (~1,756 kB) removed from the community download.
- `@turf/jsts` gone from the store chunk (−291 kB, ~37%) and no longer in the community route's static closure.

### Behavior

No user-facing behavior change. `useJumpToLocation` computes the same fitBounds envelope (tests assert exact corners for single-point, multi-point, and nested-ring inputs); analyzer proximity polygons and event-fetch bbox queries are unchanged (the turf functions just resolve via a dynamic import). Full suites for the touched areas pass and lint is clean.

### Note on scope

The remaining `@turf/turf` in the community bundle is the cheap, tree-shaken functions (`bearing`, `point`, `bbox`, …) woven through core plumbing (`ducks/events`, `selectors/index`, etc.) that the report form genuinely reaches. Removing those would mean Epic-scale decoupling of `ReportManager` from the Redux/selector layer, with sharply diminishing returns — intentionally out of scope here.

### DIFF
--- a/src/LocationPicker/MenuPopover/index.js
+++ b/src/LocationPicker/MenuPopover/index.js
@@ -101,14 +101,15 @@ const MenuPopover = ({
     if (!isPickingLocation) {
       const onKeyDown = (event) => {
         if (event.key === 'Tab') {
+          const lastFocusableElement = lastFocusableElementRef.current || gpsInputRef.current;
           if (event.shiftKey && document.activeElement === gpsFormatToggleRef.current) {
             event.preventDefault();
 
-            lastFocusableElementRef.current?.focus();
-          } else if (!event.shiftKey && document.activeElement === lastFocusableElementRef.current) {
+            lastFocusableElement.focus();
+          } else if (!event.shiftKey && document.activeElement === lastFocusableElement) {
             event.preventDefault();
 
-            gpsFormatToggleRef.current?.focus();
+            gpsFormatToggleRef.current.focus();
           }
         }
       };
@@ -196,12 +197,6 @@ const MenuPopover = ({
             {t('getUserLocationButton')}
           </>}
         />}
-
-        {!map && !showUserLocation && (
-          // No action buttons render in this context (e.g. the map-less community form),
-          // but the focus trap still needs a defined "last focusable" element to wrap to.
-          <span ref={lastFocusableElementRef} tabIndex={-1} aria-hidden="true" className={styles.focusSentinel} />
-        )}
       </div>
     </div>
   </Popover>;
--- a/src/LocationPicker/MenuPopover/index.js
+++ b/src/LocationPicker/MenuPopover/index.js
@@ -1,4 +1,4 @@
-import { lazy, Suspense, useEffect, useRef } from 'react';
+import { lazy, Suspense, useContext, useEffect, useRef } from 'react';
 import Popover from 'react-bootstrap/Popover';
 import { useSelector } from 'react-redux';
 import { useTranslation } from 'react-i18next';
@@ -7,6 +7,8 @@ import { ReactComponent as GpsLocationIcon } from '../../common/images/icons/gps
 import { ReactComponent as MarkerFeedIcon } from '../../common/images/icons/marker-feed.svg';
 
 import { EVENT_REPORT_CATEGORY, trackEventFactory } from '../../utils/analytics';
+
+import { MapContext } from '../../MapContext';
 
 import GetUserLocationButton from '../../GetUserLocationButton';
 import GpsInput from '../../GpsInput';
@@ -34,6 +36,8 @@ const MenuPopover = ({
   ...otherProps
 }) => {
   const { t } = useTranslation('components', { keyPrefix: 'locationPicker.menuPopover' });
+
+  const map = useContext(MapContext);
 
   const isPickingLocation = useSelector((state) => state.view.mapLocationSelection.isPickingLocation);
   const showUserLocation = useSelector((state) => state.view.showUserLocation);
@@ -167,18 +171,20 @@ const MenuPopover = ({
       />
 
       <div className={styles.buttons}>
-        <Suspense fallback={null}>
-          <PickMapLocationButton
-            onClick={() => eventReportTracker.track('Click \'Set on map\'')}
-            onPick={onMapLocationPick}
-            ref={!showUserLocation ? lastFocusableElementRef : undefined}
-            renderContent={() => <>
-              <MarkerFeedIcon />
+        {map && (
+          <Suspense fallback={null}>
+            <PickMapLocationButton
+              onClick={() => eventReportTracker.track('Click \'Set on map\'')}
+              onPick={onMapLocationPick}
+              ref={!showUserLocation ? lastFocusableElementRef : undefined}
+              renderContent={() => <>
+                <MarkerFeedIcon />
 
-              {t('pickMapLocationButton')}
-            </>}
-          />
-        </Suspense>
+                {t('pickMapLocationButton')}
+              </>}
+            />
+          </Suspense>
+        )}
 
         {showUserLocation && <GetUserLocationButton
           onClick={() => eventReportTracker.track('Click \'Use my location\'')}
--- a/src/LocationPicker/MenuPopover/index.js
+++ b/src/LocationPicker/MenuPopover/index.js
@@ -196,6 +196,12 @@ const MenuPopover = ({
             {t('getUserLocationButton')}
           </>}
         />}
+
+        {!map && !showUserLocation && (
+          // No action buttons render in this context (e.g. the map-less community form),
+          // but the focus trap still needs a defined "last focusable" element to wrap to.
+          <span ref={lastFocusableElementRef} tabIndex={-1} aria-hidden="true" className={styles.focusSentinel} />
+        )}
       </div>
     </div>
   </Popover>;
--- a/src/LocationPicker/MenuPopover/index.js
+++ b/src/LocationPicker/MenuPopover/index.js
@@ -104,11 +104,11 @@ const MenuPopover = ({
           if (event.shiftKey && document.activeElement === gpsFormatToggleRef.current) {
             event.preventDefault();
 
-            lastFocusableElementRef.current.focus();
+            lastFocusableElementRef.current?.focus();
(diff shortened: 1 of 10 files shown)


src/hooks/useJumpToLocation/index.js
```diff
@@ -45,7 +36,13 @@ const useJumpToLocation = () => {
...
 
     if (isArrayCoords && coords.length > 1) {
-      const mapBoundaries = coords.reduce(buildLocationJumpBounds, new LngLatBounds());
+      const points = coords.flatMap(toLeafCoords);
+      const lngs = points.map((point) => point[0]);
+      const lats = points.map((point) => point[1]);
+      const mapBoundaries = [
+        [Math.min(...lngs), Math.min(...lats)],
+        [Math.max(...lngs), Math.max(...lats)],
+      ];
```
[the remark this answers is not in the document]
jeslefcourt: Good catch — fixed in 8cd088da. Replaced the `Math.min(...)`/`Math.max(...)` spreads with a single-pass min/max over `points`, which avoids the argument-limit RangeError risk on large polygons and drops the two intermediate arrays. Envelope output is identical (exact-corner tests still pass).

src/ducks/analyzers.js
```diff
@@ -17,6 +17,7 @@ let featureLayerIdentifier = 1000;
 
 export const fetchAnalyzers = () => async (dispatch) => {
   // fetch the active analyzers, only processing the non-null spatial group urls
+  const { default: buffer } = await import('@turf/buffer');
   const { data: { data } } = await axios.get(ANALYZERS_API_URL, { params: { active: true } });
 
   const analyzers = await Promise.all(data.map(async (analyzer) => {
```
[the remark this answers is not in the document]
jeslefcourt: Agreed — fixed in 8cd088da. `fetchAnalyzers` now fetches the analyzer list first and only dynamically imports `@turf/buffer` when at least one proximity analyzer is present, so geofence-only deployments never load the buffer/jsts chunk.

src/utils/query.js
```diff
@@ -19,6 +19,7 @@ export const getBboxParamsFromMap = async (map, asString = true) => {
       (distance(asPointArray[0], asPointArray[1]) / 10), 10
     ), 0.333,
   );
+  const { default: buffer } = await import('@turf/buffer');
```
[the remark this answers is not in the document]
jeslefcourt: Leaving this as-is. Dynamic `import()` is memoized by the ES module registry: the `@turf/buffer` chunk is fetched and evaluated only once, and subsequent `getBboxParamsFromMap` calls resolve from cache in a microtask — there's no repeated chunk load. Since the function is already async and does network I/O, the resolved-promise await is negligible. (The `epsg-index/all.json` case is different — that's a ~7 MB JSON where an explicit one-time load reads more clearly.)

src/LocationPicker/MenuPopover/index.js
```diff
@@ -167,18 +171,20 @@ const MenuPopover = ({
...
+              onPick={onMapLocationPick}
+              ref={!showUserLocation ? lastFocusableElementRef : undefined}
+              renderContent={() => <>
+                <MarkerFeedIcon />
+
+                {t('pickMapLocationButton')}
+              </>}
+            />
+          </Suspense>
+        )}
```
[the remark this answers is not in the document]
jeslefcourt: Good catch — fixed in the latest commit. This actually pre-dated the change (PickMapLocationButton returned null without a map, so the ref was never attached even before), but it's real: with no map and `showUserLocation` false, the focus trap's Shift+Tab handler called `.focus()` on undefined. Added a non-interactive focus sentinel (`tabIndex={-1}`, `aria-hidden`) as the last focusable element in that case, so the trap wraps correctly instead of throwing. Added a test simulating Shift+Tab from the GPS format toggle with no map.

src/hooks/useJumpToLocation/index.js
```diff
@@ -14,15 +13,7 @@ const DEFAULT_LOCATION_JUMP_PADDING = {
...
-const extendBoundsForMultiDimensionalCoords = (coords, mapBounds) => {
-  coords.forEach(coord => mapBounds.extend(coord));
-  return mapBounds;
-};
-
-const buildLocationJumpBounds = (bounds, coords) => {
-  const isMultiDimensionalCoords = Array.isArray(coords[0]);
-  return isMultiDimensionalCoords ? extendBoundsForMultiDimensionalCoords(coords, bounds) : bounds.extend(coords);
-};
+const toLeafCoords = (coords) => (Array.isArray(coords[0]) ? coords.flatMap(toLeafCoords) : [coords]);
```
luixlive: This function name is a bit obscure. Can we change it for something a bit more descriptive? One option that describes what the method does: `flattenToCoordinatePairs`. But if you can think of a better one please use that.

src/LocationPicker/MenuPopover/index.js
```diff
@@ -190,6 +196,12 @@ const MenuPopover = ({
             {t('getUserLocationButton')}
           </>}
         />}
+
+        {!map && !showUserLocation && (
+          // No action buttons render in this context (e.g. the map-less community form),
+          // but the focus trap still needs a defined "last focusable" element to wrap to.
+          <span ref={lastFocusableElementRef} tabIndex={-1} aria-hidden="true" className={styles.focusSentinel} />
```
luixlive: Hmm this feels like an ugly patch, not a real solution for the focus trap. The aria-hidden sentinel `span` fixes the crash but moves real keyboard focus onto an element that's invisible to screen readers and has no visible focus outline. This is a WCAG violation.

The popover always has a real fallback element already: when no action button renders, the GPS input is the true last focusable element in the DOM.

Instead of a dummy span, the `useEffect` should fall back to focusing the GPS input ref when `lastFocusableElementRef.current` is empty, in both the `Shift+Tab` and `Tab` wrap conditions. That also removes the need for the sentinel's CSS and the ?. guards, and fixes the same gap during the brief window before the map-pick button's lazy chunk loads.

src/sw-build.js
```diff
@@ -11,8 +11,16 @@ const buildSW = () => {
       swDest: path.join(process.cwd(), 'build/sw.js'), // sw output file (auto-generated)
       globDirectory: path.join(process.cwd(), 'build'),
       globPatterns: ['**/*.{js,html,css,png,svg}'],
-      globIgnores: ['**/*service-worker*.js', '**/*precache-manifest*.js'],
-      maximumFileSizeToCacheInBytes: 10 * 1024 * 1024, // 10 MB
+      globIgnores: [
+        '**/*service-worker*.js',
+        '**/*precache-manifest*.js',
+        // Heavy, dynamically-imported chunks: exclude from eager precache so a first
```
luixlive: We've been noticing Claude models rely more and more on big block comments. It's like if the model uses comments as its scratchpad for thinking. We've tried to control this by explicitly asking not to in the AGENTS.md (symlinked to CLAUDE.md), but 🤷‍♂️  It's great to add brief comments to explain complex or obscure code blocks, but no need to add a full paragraph:
```Javascript
globPatterns: ['**/*.{js,html,css,png,svg}'],
globIgnores: [
  '**/*service-worker*.js',
  '**/*precache-manifest*.js',
  // Heavy on-demand chunks; runtime-cached instead (see sw-custom.js).
  '**/all-*.js', // epsg-index/all.json, ~7.4 MB
  '**/PickMapLocationButton-*.js', // mapbox-gl, ~1.76 MB
],
maximumFileSizeToCacheInBytes: 3 * 1024 * 1024, // 3 MB
```

luixlive (changes requested):
Great work. Almost ready. The lose of accessibility after the change to the focus mechanism in the `LocationPicker` is a regression. Let's just fix that and this will be ✅

src/hooks/useJumpToLocation/index.js
jeslefcourt: Renamed to `flattenToCoordinatePairs` — clearer, agreed.

src/LocationPicker/MenuPopover/index.js
jeslefcourt: You're right — the sentinel traded a crash for a WCAG regression (focus on an invisible, non-interactive element). Reworked it to fall back to the always-present GPS input as the last focusable element, in both the Tab and Shift+Tab wrap branches. That keeps focus on a real, visible, screen-reader-friendly element and also covers the lazy-load window, so the sentinel, its CSS, and the `?.` guards are all gone. Updated the test to assert focus lands on the GPS input.

src/sw-build.js
jeslefcourt: Fair callout — trimmed to your condensed version, and tightened the other verbose comment I'd added (ducks/analyzers). Thanks for the nudge; keeping them terse per CLAUDE.md.

luixlive (approved):
Nice 💯

PR #1677, opened 2026-07-28
feat:inline media players for audio/video

JoshuaVulcan:
[draft]


public/locales/en-US/details-view.json
```diff
@@ -21,6 +21,9 @@
     "collapseOpenButtonLabel": "Collapse",
     "collapseOpenButtonTitle": "Expanded",
     "imagePreviewAlt": "{{fileName}} preview",
+    "videoPreviewAlt": "{{fileName}} video preview",
```
luixlive, line 24: nit: Sort the keys alphabetically so it's easier to search for stuff 👍

src/DetailViewComponents/ActivitySection/AttachmentListItem/index.js
```diff
@@ -37,26 +39,61 @@ const AttachmentListItem = ({
...
+        })
+        .catch(() => {
+          mediaFetchPromiseRef.current = null;
+          setMediaError(true);
+
+          return null;
+        });
+    }
+
+    return mediaFetchPromiseRef.current;
```
luixlive, line 74: Why is this returning the ref? The places from where it is called are not storing it, and the refs are anyway accessible from anywhere. It's a bit misleading.

src/ImageModal/index.js
```diff
@@ -72,9 +74,19 @@ const ImageModal = ({ id, src, title, url, tracker }) => {
     </Header>
 
     <Body className={styles.body}>
-      {!loaded && <LoadingOverlay />}
+      {!loaded && !fetchError && <LoadingOverlay />}
+
+      {!showError && mediaType === 'video' && <video
```
luixlive, line 79: This component is called `ImageModal`. The clean solution here would be to either have a different specific modal for video (if worth it), or renaming this one to a more generic name.

src/DetailViewComponents/ActivitySection/AttachmentListItem/index.js
```diff
@@ -37,26 +39,61 @@ const AttachmentListItem = ({
...
   const [imageIconSource, setImageIconSource] = useState(null);
   const [imageOriginalSource, setImageOriginalSource] = useState(null);
+  const [mediaObjectUrl, setMediaObjectUrl] = useState(null);
+  const [mediaError, setMediaError] = useState(false);
 
   const currentImageSource = useMemo(() => imageOriginalSource || imageThumbnailSource, [imageOriginalSource, imageThumbnailSource]);
 
-  const onShowImageFullScreen = useCallback((event) => {
+  const mediaFetchPromiseRef = useRef(null);
+  const pendingModalIdRef = useRef(null);
```
luixlive, line 55: From Claude:
A single `pendingModalIdRef` is overwritten on every fullscreen click, so opening the same video's fullscreen modal twice in quick succession leaves the first modal permanently stuck showing the loading spinner.

src/DetailViewComponents/ActivitySection/AttachmentListItem/index.js
```diff
@@ -65,50 +102,84 @@ const AttachmentListItem = ({
...
+  useEffect(() => {
+    if (mediaError && pendingModalIdRef.current) {
+      dispatch(updateModal({ id: pendingModalIdRef.current, fetchError: true }));
+      pendingModalIdRef.current = null;
+    }
+  }, [dispatch, mediaError]);
+
+  useEffect(() => () => {
+    if (mediaObjectUrl) {
+      URL.revokeObjectURL(mediaObjectUrl);
```
luixlive, line 159: From Claude:
The object URL created for video/audio playback is revoked on component unmount, but a fullscreen `ImageModal` referencing that same URL lives in global Redux modal state and isn't closed when the detail view unmounts.

src/DetailViewComponents/ActivitySection/styles.module.scss
```diff
@@ -4,6 +4,10 @@
 
 $cardExpansionTransitionTime: 300ms;
 
+@keyframes mediaLoadingSpin {
```
luixlive, line 7: From Claude:
A new hand-rolled CSS spinner (`mediaLoadingSpin` keyframes + `.mediaLoadingSpinner`) duplicates the existing `LoadingOverlay` component that this same PR already uses one file away.

src/ReportManager/DetailsSection/SchemaForm/formElements/Attachment/index.js
```diff
@@ -176,15 +219,35 @@ const AttachmentListItem = ({ actionButtonRefs, attachment, onRemove, readOnly }
...
+
+      {actionButton}
+    </div>
+
+    {isMedia && isPlayerOpen && attachment.status === 'complete' && <div className={styles.player}>
+      {mediaError
+        ? <span className={styles.error}>{t('playerErrorLabel')}</span>
+        : mediaObjectUrl
+          ? (attachment.fileType === 'audio'
+            ? <audio
```
luixlive, line 239: Here we have like 3 nested ternary operators, which makes it hard to reason about. We should abstract some of this code into separate components or methods.

luixlive (changes requested):
Some feedback from me and some from Claude. Candidly, some of these changes feel like a patch over our implementation for images, with hardcoded conditions and complex flows with several `useEffect`s and different fetching and error branches. I feel like it would be cleaner to have specific methods and components per file type so the code is easier to understand, which also makes it less error-prone and easier to test. Of course we can have shared files to avoid repetition.

luixlive (approved):
Awesome, thanks for the fixes Joshua! Nothing blocking stands out for me, some of my own findings were also listed by the agentic review I'm sharing. I also see you got some comments from Copilot. I'll approve it since the files and changes are coherent and make sense, but I'd recommend to work on some of the findings shared before merging since they seem like possible annoying bugs. Putting an agent to read this and Copilot's comment and tackle them one by one (validating the issue plus an elegant fix) should be trivial 👍 

1. **`useMediaObjectUrl` returns a revoked object URL and crashes on an `undefined` url** - `src/hooks/useMediaObjectUrl.js:52`

`return fetchedMedia?.url === url ? fetchedMedia : NO_MEDIA;` never clears state when the effect re-runs, so it misbehaves two ways. **(a)** Collapse a player and reopen it: the guard matches the previous entry and hands back the `objectUrl` the cleanup already revoked, so the element mounts against a dead blob until the refetch lands (same for a stale `error: true`, which shows the error message instead of the spinner during a retry). **(b)** With `url === undefined`, `null?.url === undefined` is `true`, so the hook returns `null` and every caller throws `Cannot destructure property 'error' of … as it is null`. Both reproduced locally. Resetting the state at the top of the effect fixes both. Note the existing tests only exercise close/reopen *after a failure*, where `objectUrl` was never set, so the gap is invisible.

2. **Audio and video attachments lose their download action** -`src/DetailViewComponents/ActivitySection/AttachmentListItem/index.js:165,196` · `src/SchemaForm/formElements/Attachment/index.js:195`

Widening the branch to `isImageAttachment || isPlayableAttachment` moves audio/video out of the generic row that rendered the download button, and the new branch suppresses the fullscreen button for audio — so an audio attachment in the Activity section now has no action but expand/collapse. In the report form, `else if (isMedia)` pre-empts the download branch, so saved audio and video have no download path either, including in read-only mode. Only documents keep a download button; images reach one through the modal.

3. **A media file that downloads but cannot be decoded never reaches the error state** - `src/utils/file/index.js:25` · `AttachmentListItem/index.js:80,89` · `SchemaForm/formElements/Attachment/index.js:116,119`

`fetchFileAsObjectUrlFromUrl` drops both safeguards its sibling `fetchImageAsBase64FromUrl` carries — the content-type check (whose comment says the SPA's `index.html` answers 200 to any unmatched path) and `calcUrlForImage`. An HTML body becomes a valid blob URL and the hook reports `error: false`. Combined with the fact that neither inline player wires `onError` (only `MediaModal` does), an HEVC `.mov`, a truncated upload or a mis-routed URL renders a dead player and the new `mediaLoadErrorMessage` / `playerErrorLabel` strings are unreachable.

4. **Clicking anywhere while a video downloads closes the fullscreen modal** - `src/MediaModal/index.js:53`

`mediaRef` is only attached once `mediaSource` exists, so during the blob fetch (and in the error state) `!mediaRef.current?.contains(...)` is unconditionally true: any `pointerdown` — including on the spinner — dispatches `removeModal` and fires a bogus "Click modal background to close media modal" event. Verified with a probe. The old modal always rendered the ``, so this window did not exist; it is now as long as the download. Put the ref on a wrapper that always renders, or skip the handler while `!mediaSource`.

5. **Downloads are never aborted and never cached** - `src/hooks/useMediaObjectUrl.js:43` · `src/utils/file/index.js:25`

The cleanup only flips `isCancelled`; the axios GET runs to completion after a collapse, a modal close or an unmount, and the result is revoked on arrival. There is no cache either, so every expand → collapse → expand re-downloads the whole file. `src/ducks/user-content` already uses `AbortController` + `{ signal }`, and `downloadFileFromUrl` already threads a cancel token — pass one through here.

6. **"Expand All" starts a full-file download of every audio and video attachment at once** - `src/DetailViewComponents/ActivitySection/useActivityFeed.js:224`

`collapsibleAttachments` feeds `collapsibleCardIds`, which `onToggleExpandAll` writes wholesale into `expandedCardIds`. One click fires N concurrent, uncancellable whole-file GETs and pins every blob in memory until the rows are collapsed. Images were safe here because they only pull thumbnails/originals. Consider leaving playable attachments out of the bulk set, or gating the fetch behind an explicit play the way the report form does.

7. **Collapsed media rows render the loading overlay — during the collapse animation, and in print** - `AttachmentListItem/index.js:73,123` · `ActivitySection/styles.module.scss:105,239`

`Collapse` keeps its children mounted and a closed row passes `null` to the hook, so `MediaAttachmentPreview` permanently falls into the `!objectUrl` branch. Three consequences: collapsing swaps a 26rem video for the 8rem dark spinner box *before* the 300ms exit transition runs (visible jump); every collapsed media row carries a dead MoonLoader; and because audio/video no longer get the `.nonImageAttachment` class — whose `@media print { display: none }` used to hide them — while `.collapse` forces `display: block !important` for print, a printed event report now shows a dark 8rem spinner block per media attachment where it previously showed nothing.

8. **`I18N_FILES_VERSION` was not bumped** - `src/i18n.js:10`

12 locale files changed, including the `imageModal` → `mediaModal` rename, but the constant is still `'1.66'`, the value `origin/develop` carries. `yarn check-i18n-files-version` fails. Returning users keep the cached bundle for seven days and see raw keys — `mediaModal.errorTitle`, `playButtonLabel`, `playerErrorLabel`, `audioPreviewAlt` — several of which are a control's only accessible name.

9. **Opening a video fullscreen downloads it a second time** - `AttachmentListItem/index.js:139`

`src: isVideoAttachment ? null : defaultImageSource` deliberately withholds the source so the modal fetches its own copy. The comment explains why (the row revokes its blob on collapse), but the effect is that an already-expanded, fully-buffered video is pulled over the network again with both copies resident, and the modal's own download button pulls it a third time. Ref-counted shared ownership of the object URL would remove both the workaround and the duplicate transfer.

10. **The modal's `<video>` stays hidden until `loadeddata`, with no `preload`** - `src/MediaModal/index.js:92`

`style={{ display: loaded ? 'block' : 'none' }}` is an ``-era trick: `onLoad` always fires, but `loadeddata` needs `readyState >= HAVE_CURRENT_DATA`. On a UA that honours `preload="metadata"` for an offscreen element the spinner never clears and the user cannot tap play to break it, because the element is `display: none`. Reveal on `onLoadedMetadata`/`onCanPlay`, or keep the element visible and stack the overlay over it. The inline player renders the same element with no gate at all, so the two disagree about what "loaded" means.

11. **The inline player is taller than the list it lives in** - `src/SchemaForm/formElements/Attachment/styles.module.scss:58,81`

`.attachmentsList` keeps its `max-height: 11.25rem; overflow-y: auto`, sized for 2.5rem rows, while `.player video` is up to 12rem plus padding. An expanded video can never be fully visible, and nothing scrolls the newly opened player into view — open the player on the last of several attachments and it renders entirely below the fold of the scroll box.

12. **Accessibility gaps on the new control and its states** - `SchemaForm/formElements/Attachment/index.js:196,104` · `AttachmentListItem/index.js:63`

The play/close button is a disclosure but has no `aria-expanded` and no `aria-controls`, and the player div has no `useId` — the sibling collapse toggle at `AttachmentListItem:209` does set `aria-expanded`, so this is an inconsistency inside one feature. Its label "Play {{fileName}}" also does not describe what it does: it reveals a *paused* player behind a chevron icon, and `closePlayerButtonLabel` is already phrased as a visibility action. And all three async states are silent — the error is a bare `<span>`/`<p>` and the loading state an unlabelled `LoadingOverlay` — though this very file already owns an `aria-live="polite" role="status"` region at line 484, and the repo's precedent is `role="alert"` / `role="status"` plus an sr-only label.

13. **The two players are duplicated instead of shared, and the new branches carry dead code** - `AttachmentListItem/index.js:63,138` · `SchemaForm/formElements/Attachment/index.js:104,235` · `MediaModal/index.js:40,85`

This is the standing objection from the previous round, still open. `MediaAttachmentPreview` and `MediaPlayer` are the same component written twice (error → message, `!objectUrl` → overlay, audio vs video), with the message duplicated across two namespaces × six locales, the CSS duplicated across two modules, and `useBase64ImageSource` still inline right next to the new hook doing the same job. On top of that: `mediaType: isVideoAttachment ? 'video' : 'image'` and `src: isVideoAttachment ? null : defaultImageSource` are both no-ops (the button only renders for image/video, and `defaultImageSource` is already gated on `isImageAttachment`); `setMediaLoaded()` inside `setMediaError` is dead now that the overlay checks `showError`; `isMedia && isPlayerOpen && status === 'complete'` reduces to `isPlayerOpen`; the three mutually-exclusive icon conditionals are the `X_BY_Y` lookup CLAUDE.md names explicitly; and `MediaModal` renders nothing at all (permanent spinner) for any `mediaType` outside the two literals.

14. **Repo conventions in the new code** - `AttachmentListItem/index.js:63` · `SchemaForm/formElements/Attachment/index.js:104,132,236` · `MediaModal/index.js:18,29` · `setupTests.js:58`

- Subcomponents belong in a nested folder under their parent, not in the parent's `index.js`.
- `t={t}` is threaded into both subcomponents — the only two occurrences in `src/`. Each should call `useTranslation` with its own key prefix.
- Comments over the 80-column ceiling: `setupTests.js:58` (99) and `:59` (101), `useMediaObjectUrl.js:22` (95), `AttachmentListItem:131` (95), `MediaModal:29` (86).
- Body-order drift: `useMediaObjectUrl` lands after `useContext`/`useState` in both components; `MediaModal`'s signature ends `url, tracker` instead of `tracker, url`.
- `setupTests.js:60` stubs `URL.createObjectURL` to a single constant, so no test can distinguish two sources or catch a double revoke — which is exactly why the hook tests miss finding 1. A counter-based stub would make the lifecycle assertable. (The comment's premise is also wrong: Node's `URL` does implement it; it throws only because axios yields a jsdom Blob.)

15. **Test and translation gaps** - `src/MediaModal/index.test.js` · `ActivitySection/index.test.js:91` · `sw/schema-form.json:28` · `ne-NP/details-view.json:20,28` · `fr/details-view.json:62`

**Tests:** nothing asserts the tracker (all three action strings were renamed with zero coverage, and the mock is inlined so no test can hold it), the click-outside dispatch, or the `display` toggle; the new suites reach for `document.querySelector('video'/'audio'/'svg')` where `getByLabelText` works — line 57 of the same file proves it; the hook has no success → close → reopen test, the gap that hid finding 1. The fixture id swap in `ActivitySection/index.test.js` is backwards: the note is now named `media-attachment-video-1` while the video attachment carries the note's freed UUID (a rename was needed, but the new attachment should have taken the descriptive id).

**Translations** (key parity and ordering are otherwise clean across all six locales): sw `"Funga kicheza cha {{fileName}}"` uses a bare verb form where its own siblings use a compound (`Kicheza sauti cha`, `Kicheza video cha`) — "kichezaji" is the idiomatic agent noun, and all three keys are new here so there is no precedent to match. ne-NP spells preview `पुर्वावलोकन` in both new keys, propagating a local misspelling — `ne-NP/schema-form.json:65` has the correct `पूर्वावलोकन`. fr `errorTitle` is the only locale missing the trailing period, and the line the diff rewrote sits directly above an `errorMessage` reading "avez télécharger … veuillez patientez", which is ungrammatical and uses *télécharger* (download) for upload. Finally, `errorTitle` still precedes `errorMessage` in all six locales — pre-existing, but the diff rewrites both lines.

PR #1683, opened 2026-08-03
Patrol Types preparation work

luixlive:
## What does this PR do?
- Upgrade dependencies.
- Migrate our drag and drop implementations with the legacy `dnd-kit/core` to `dnd-kit/react`.
- Modernize the React Router implementation of this app to follow the conventions of v8 for apps using the declarative mode.
- Abstract `SchemaForm` to be a standalone component, agnostic of the parent implementation, so it can be reused for patrol schemas.
- Define the architectural approach to follow for the new patrol views (https://allenai.atlassian.net/wiki/spaces/ER/pages/31569149954/ER+Web+Architecture); and defined placeholder routes and components in the app.
- Update AGENTS.md.

## Notes
This PR has lots of changed files, but most of the changes are trivial to review, specifically for an agent. Moving the `SchemaForm` component to be standalone resulted in more than 100 file changes that are only updates in the imports.


chrisj-er (approved):
Very clean for 162 files. Claude mentions:

 onItemMove(-1, -1) is reachable (SortableList/index.js:60-66). @dnd-kit/helpers' mutate can return a new array when the source id isn't in items, so find returns undefined and both indexOf calls give -1. Guard on movedItem, or drive off event.operation.source.initialIndex/.index.

luixlive:
> Very clean for 162 files. Claude mentions:
> 
> onItemMove(-1, -1) is reachable (SortableList/index.js:60-66). @dnd-kit/helpers' mutate can return a new array when the source id isn't in items, so find returns undefined and both indexOf calls give -1. Guard on movedItem, or drive off event.operation.source.initialIndex/.index.

That was a pretty nice catch. Claude really went deep into the flow and the library internals!

Fixed. I just added a guard to make sure we don't try to mutate the list if the source id of the moved items isn't found. It seems to be unlikely to happen, but glad to have those edge cases covered 👍

